PR #13, opened 2020-01-03
Project structure re-organization

### LINKED ISSUE #11: Create a setup.py
...as noted in #5

wasade:
This pull request fixes #11. Completion of this issue required a reorganization of the project structure, placing modules within a common package. 

It is now possible use `setup.py` to install this package. An example of this is in the readme. Manual modification of `PYTHONPATH` is no longer necessary. 

A few notes:

- all imports were updated that I could find, including the `operationId` entries in the openapi
- the Flask server appears to start, and swagger UI seems to be there
- LEGACY was moved as we're still pretty coupled to the locale structures at the moment
- configuration was moved out of legacy as it's pretty light, and we need something like that anyway
- `versioneer` was used to keep in line with project management flow of QIIME 2
- the package name is long, which is not ideal, but we can refactor in the future if that becomes annoying. Alternatively, we could look to standardize on a short name, for example `import microsetta_private_api as mpa`

Heads up:

- the `versioneer.py` script included here *DOES NOT* need to be reviewed. That package requires the `.py` at the top level with `setup.py`
- the `_version.py` script within `microsetta_private_api` *DOES NOT* need to be reviewed. This is autogenerated by `versioneer`. 

On an aside, I set the `author` and `author_email` entries to me. @AmandaBirmingham and @dhakim87, please let me know if this is a concern or if a change is desired. I don't believe multiple entries can be specified there, and the use of these fields (as I understand it) is I believe for the first point-of-contact.


microsetta_private_api/server.py
```diff
@@ -16,13 +16,11 @@
...
-    app.add_api('microsetta_private_api.yaml')
+    app.add_api('api/microsetta_private_api.yaml')
 
     # Set default json encoder
     # Note: app.app is the actual Flask application instance, so any Flask
     # settings have to be set there.
     app.app.json_encoder = JsonifyDefaultEncoder
 
-    app.app.template_folder = "../templates"
-    app.app.static_folder = "../static"
```
AmandaBirmingham: Just checking my understanding: these lines were removed because they are no longer needed because the templates and static directories are now in whatever default location flask expects for them?

setup.cfg
```diff
@@ -0,0 +1,7 @@
+[versioneer]
+VCS = git
+style = pep440
+versionfile_source = microsetta_private_api/_version.py
+versionfile_build = microsetta_private_api/_version.py
+tag_prefix =
+parentdir_prefix = microsetta-priavte-api-
```
AmandaBirmingham: Whoops, typo! "microsetta-priavte-api-" should be "microsetta-private-api-"

microsetta_private_api/server.py
wasade: That is correct as I understand it. It looks like the flask app constructor has `templates` and `static` as the default paths

setup.cfg
wasade: Fixed, thanks!!

microsetta_private_api/server.py
dhakim87: As far as I can tell, those paths are relative to the server.py file itself, which is why it needed the ../ before.  Just checked it locally and it's able to find these paths correctly.  Looks good.

PR #14, opened 2020-01-03
BLD: initial travis configuration

### LINKED ISSUE #5: Setup Travis CI
It would be pleasant to have continuous integration setup to automatically trigger unit tests and other code base checks on pull requests

### LINKED ISSUE #12: Establish pip / conda requirements files
As noted in #5...

wasade:
Let's see how this works. Once this configuration passes, we can hook up a test runner which will make the use of travis a bit more interesting :) 

Fixes #5 
Fixes #12


wasade:
Any issues with merging this?

.travis.yml
```diff
@@ -0,0 +1,35 @@
...
+  - linux
+  - os
+
+env:
+  - CONDA_PY=3.8
+  - CONDA_PY=3.7
+  - CONDA_PY=3.6
+
+addons:
+  postgresql: "9.5"
```
AmandaBirmingham: Remedial question: Does this limit usage to postgresql 9.5? I'm using postgresql 12 ...

microsetta_private_api/api/implementation.py
```diff
@@ -21,19 +24,16 @@
 from microsetta_private_api.repo.sample_repo import SampleRepo
 
 from microsetta_private_api.model.source import Source
-from microsetta_private_api.model.survey_template import SurveyTemplate
-from microsetta_private_api.model.mock_jinja import MockJinja
```
AmandaBirmingham: AFAIcan tell, this (unused) import was the only usage of the MockJinja class (and thus the mock_jinja.py file in model).  Should we remove that file?

Also, I understand that SurveyTemplate isn't being used yet, but it definitely will be ... I suppose Dan can just add it back then.

.travis.yml
wasade: It's what we have been using for AG. If the database works as is with postgres 12, then I'd say we use this as an opportunity to update

microsetta_private_api/api/implementation.py
wasade: If it'll be used then we should keep it :)

dhakim87: I used the MockJinja objects to test display of forms in the Jinja framework.  It's a mock object that should never be used on a live server, but it's helpful for testing changes in the consent doc.  It should exist at the same level as any unit test packages we add in the future.

wasade: okay cool. Not deleting it, just removing the import because it was not actively being used in that module.

.travis.yml
```diff
@@ -0,0 +1,35 @@
+sudo: false
+os:
+  - linux
+  - os
```
dhakim87: os or osx?

wasade:
@dhakim87 good to merge?

AmandaBirmingham:
@dhakim87 says ok.  Merging.

PR #17, opened 2020-01-04
Consent schema fix, authorization token mocking example

AmandaBirmingham:
This PR 
1) Fixes (I think!) the schema validation failures for the example sources provided (this required rejiggering both the source schemas and the source examples) ... jsonschema additionalProperties is a harsh mistress.
2) makes the token validation function always use a hardcoded dummy token, no matter what is actually passed in via the authentication: bearer header.  This is obviously not going to be this way forever, but for now it allows us to use curl to make requests that looks structurally like those that would result from oauth2 authentication even though we have no source for valid oauth2-generated tokens.


wasade:
Thanks, @AmandaBirmingham! I think this is good, didn't see anything on review

PR #22, opened 2020-01-07
Updated style for classmethod vs staticmethod

### LINKED ISSUE #4: classmethod or staticmethod?
[`AccountRepo._account_to_row`](https://github.com/biocore/microsetta-private-api/blob/52798b9823332114aa98b20a939cabdbbbdc1527/repo/account_repo.py#L21) and its inverse are currently decorated with `classmethod`, however these appear to instead be `staticmethod`s. Should these be changed?

dhakim87:
Resolves #4   

This is almost entirely stylistic, but it can't hurt to define a particular style.  

Factory methods will use @classmethod, and will use the cls parameter for object construction, allowing for subclasses to use these methods (so long as their __init__ parameter lists match).  Honestly if we are subclassing classes with factory methods, we are probably going to have a bad time anyway, so this shouldn't come up often.  

Utility methods (that don't require a class instance, but should be associated with a class for namespacing purposes) will use @staticmethod

### DIFF
--- a/microsetta_private_api/model/sample.py
+++ b/microsetta_private_api/model/sample.py
@@ -25,4 +25,4 @@ class Sample:
         # NB a sample may NOT have date and time collected if it has been sent out but not yet used
         if date_collected is not None and time_collected is not None:
             datetime_collected = datetime.combine(date_collected, time_collected)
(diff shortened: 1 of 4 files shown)


wasade:
:+1:

PR #38, opened 2020-01-09
Add ids to sample and source get endpoints, add age_range to human source consent

### LINKED ISSUE #19: Add id to sample and source schemas
Working my way through the yaml, I'm unclear on how the user can navigate the api in several places:

POST /accounts generates a new account, but does not return the account ID.  How do we get the id for the GET,PUT /accounts/{account_id} ?

Same issue in GET,POST /accounts/{account_id}/sources: No source ID is returned, so no id is available for GET,PUT,DELETE /accounts/{account_id}/sources/{source_id}

Same issue in samples, though this may be blocked until schema is modified.  We sometimes pass sample_barcode around, but the url appears to be expecting a uuid.  

Same issue in POST /accounts/{account_id}/sources/{source_id}/surveys, the GET seems fine.  

GET /kits/ should probably also return id fields in the objects for the same reasons as above

Should we add id fields to these model objects?

### LINKED ISSUE #26: Add age_range to human source definition in api
right now the api isn't passing age_range back from the front end bc I thought it was only used to decide which consent to show, not saved--but turns out it is.

AmandaBirmingham:
This PR fixes #19,  of ids not being returned with GETs on collections of samples and of sources. Note the heavy new use of the `readOnly` property, which is intended to address this very issue: "You can use the readOnly and writeOnly keywords to mark specific properties as read-only or write-only. This is useful, for example, when GET returns more properties than used in POST – you can use the same schema in both GET and POST and mark the extra properties as readOnly. readOnly properties are included in responses but not in requests" (https://swagger.io/docs/specification/data-models/data-types/#readonly-writeonly). Cool!  
However, because of complex and unpleasant interactions between the `oneOf` property, the `examples` property, and the `readOnly` property, it seems I still have to define separate *examples* for source objects that have a source_id in them (i.e., GET responses) vs those that don't (i.e., POST or PUT requests).  Oh well.

This PR also fixes #26, adding age_range as part of the consent info for human sources.

### DIFF
--- a/microsetta_private_api/api/microsetta_private_api.yaml
+++ b/microsetta_private_api/api/microsetta_private_api.yaml
@@ -131,12 +131,14 @@ paths:
                   consent:
                     participant_name: "Ophelia Q. Doe"
                     participant_email: "nunnery@example.com"
+                    age_range: "18+"
                 - source_id: "df077cd7-f2c7-42e4-b8ed-9c7e9dd47ce5"
                   source_name: "P. Doe"
                   source_type: human
                   consent:
                     participant_name: "Persephone Doe"
                     participant_email: "pomegranate@example.com"
+                    age_range: "0-6"
                     child_info:
                       parent_1_name: "Demeter Doe"
                       parent_2_name: "Zeus Doe"
@@ -1025,6 +1027,9 @@ components:
               type: string
             participant_email:
               type: string
+            age_range:
+              type: string
+              enum: ["0-6", "7-12", "13-17", "18+"]
             child_info:
               type: object
               properties:
@@ -1165,6 +1170,7 @@ components:
         consent:
           participant_name: "Nellie Doe"
           participant_email: "nell@example.com"
+          age_range: "18+"
     human_child_source:
       value:
         source_name: "K. Doe"
@@ -1172,6 +1178,7 @@ components:
         consent:
           participant_name: "Kelly Doe"
           participant_email: "kd@kidsmail.com"
+          age_range: "7-12"
           child_info:
             parent_1_name: "Maman Doe"
             parent_2_name: "Pere Doe"
--- a/microsetta_private_api/api/microsetta_private_api.yaml
+++ b/microsetta_private_api/api/microsetta_private_api.yaml
@@ -950,9 +950,8 @@ components:
       type: string
       example: "Oops, I dropped it"
     sample_site:
-      type: string
       enum: ["Ear wax", "Forehead", "Fur", "Hair", "Left hand", "Left leg", "Mouth", "Nares", "Nasal mucus",
-             "Right hand", "Right leg", "Stool", "Tears", "Torso", "Vaginal mucus", "Null"]
+             "Right hand", "Right leg", "Stool", "Tears", "Torso", "Vaginal mucus", null]
       example: "Stool"
     sample_locked:
       type: boolean
--- a/microsetta_private_api/api/microsetta_private_api.yaml
+++ b/microsetta_private_api/api/microsetta_private_api.yaml
@@ -20,21 +20,23 @@ paths:
       requestBody:
         content:
           application/json:
-              schema:
-                type: object
-                properties:
-                  first_name:
-                    $ref: '#/components/schemas/first_name'
-                  last_name:
-                    $ref: '#/components/schemas/last_name'
-                  email:
-                    $ref: '#/components/schemas/email'
-                  address:
-                    $ref: '#/components/schemas/address'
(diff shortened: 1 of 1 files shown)


AmandaBirmingham:
Asking for a review from @dhakim87  as he id'd both the issues I'm (hopefully) addressing here.  Review from @wasade always welcome :) 

Quick review would be appreciated as I have another PR lined up but want to keep them separate so that reviewing process is not confusing due to too many unrelated changes.

microsetta_private_api/api/microsetta_private_api.yaml
```diff
@@ -934,14 +944,15 @@ components:
...
       example: "dae21127-27bb-4f52-9fd3-a2aa5eb5b86f"
     sample_notes:
       type: string
       example: "Oops, I dropped it"
     sample_site:
       type: string
-      enum: ["", "Ear wax", "Forehead", "Fur", "Hair", "Left hand", "Left leg", "Mouth", "Nares", "Nasal mucus",
-             "Please select...", "Right hand", "Right leg", "Stool", "Tears", "Torso", "Vaginal mucus", "Null"]
+      enum: ["Ear wax", "Forehead", "Fur", "Hair", "Left hand", "Left leg", "Mouth", "Nares", "Nasal mucus",
+             "Right hand", "Right leg", "Stool", "Tears", "Torso", "Vaginal mucus", "Null"]
```
dhakim87: Is that the string "Null"?  Is this intended to say sample_site is optional, in which case we set the column null in the database?

microsetta_private_api/api/microsetta_private_api.yaml
```diff
@@ -934,14 +944,15 @@ components:
...
     sample_id:
       type: string
+      readOnly: true # sent in GET, not in POST/PUT/PATCH
       example: "dae21127-27bb-4f52-9fd3-a2aa5eb5b86f"
     sample_notes:
       type: string
       example: "Oops, I dropped it"
     sample_site:
       type: string
-      enum: ["", "Ear wax", "Forehead", "Fur", "Hair", "Left hand", "Left leg", "Mouth", "Nares", "Nasal mucus",
```
dhakim87: The empty string "" appeared as the sample site for several records in the database.  I'm hoping that was in error, but we should double check with Daniel.

microsetta_private_api/api/microsetta_private_api.yaml
```diff
@@ -1161,6 +1188,34 @@ components:
       value:
         source_name: "Fluffy"
         source_type: animal
+    human_adult_source_w_id:
```
dhakim87: I assume this duplication is a necessary evil :(.  Is there no way to have a schema reference another schema but add a single field?

microsetta_private_api/api/microsetta_private_api.yaml
AmandaBirmingham: Good catch!  Taking out the "Null" string; column is already nullable in db.

microsetta_private_api/api/microsetta_private_api.yaml
AmandaBirmingham: My notes from our 20191213 meeting say:
Q: ag.ag_kit_barcodes.site_sampled includes some bs entries like "", "Please select...", and "Null" ... include in api enum? 
A (Daniel): Null is ok, others are not

Based on this, I deleted the others (even though, as you say, they DO exist in records in the db--which I know because a distinct on the db is where I got the list in the first place :D ).  This DOES mean we would not be able to return some records currently in the db through the API ... I assumed from @wasade 's answer that we were accepting this as a limitation in order to prevent creating new crappy records, but perhaps I misunderstood. @wasade , can you please comment?

microsetta_private_api/api/microsetta_private_api.yaml
AmandaBirmingham: Oh, the tales I could tell you!  So: there is absolutely a way to have a schema reference another schema but add a single field (teaser: see my next PR :D :D ) HOWEVER, what you are looking at here are NOT schemas--they are "examples" (which don't nest under the "schema" key but in fact have their own top-level key ... and apparently their own rules :(  ).  I tried to set up the examples as referencing other, more granular example pieces (similar to how we set up the actual schemas) using `$ref '#/components/examples/<whatever_little_piece>'`, and connexion/swagger threw up at me and I had to back out all my cunning changes :-/ 

I agree this is objectively awful.  I am opening a new issue (for someday) to return to this and see if fresh eyes/updated connexion releases/new stackoverflow examples can find a way to remove this code smell.

PR #51, opened 2020-01-10
Fill in Sample.sample_projects from the database.  

### LINKED ISSUE #23: Fill in Sample.sample_projects
Daniel has in the past indicated that the project name-to-sample relationships are many-to-many and are retrieved by joining as shown below:

select barcodes.project.project FROM
barcodes.barcode
LEFT JOIN barcodes.project_barcode ON
barcodes.barcode.barcode = barcodes.project_barcode.barcode
LEFT JOIN barcodes.project ON
barcodes.project_barcode.project_id = barcodes.project.project_id
where barcodes.barcode.barcode = XXXXX

(provide your own where clause, obviously :) . The results will need to be put into a list.

dhakim87:
Resolves #23 

Adds a secondary query to fill in sample_projects whenever we pull in a Sample object using the SampleRepo.  Fixed KitRepo to use SampleRepo internally so we don't have to keep both in sync.


microsetta_private_api/model/sample.py
```diff
@@ -23,23 +24,23 @@ def is_locked(self):
...
         return {
             "sample_barcode": self.barcode,
             "sample_site": self.site,
-            "sample_locked": self.is_locked(),
+            "sample_locked": self.is_locked,
             "sample_datetime": self.datetime_collected,
             "sample_notes": self.notes,
-            "sample_projects": None  # TODO: Where is this info?
+            "sample_projects": list(self.sample_projects)
         }
```
AmandaBirmingham: Q: Just curious--what is the usage of list(...) here?  I would have thought from the line 

`sample_projects = [project[0] for project in project_rows]`

in `sample_repo.get_sample()` that the `sample_projects` argument was already a list ... is this belt-and-suspenders approach, or a copy thing, or ...?

dhakim87: Just a copy thing.  Didn't like the idea that modifying any of the to_api fields could mutate the input objects.  Maybe I'll look into some way to freeze the output dictionaries so I can stop copying things for no reason

Issue #39, opened 2020-01-09
Return *array* of survey templates from GET /accounts/{account_id}/sources/{source_id}/survey_templates


AmandaBirmingham:
Errr ... somehow I apparently incorrectly set this endpoint to return a *single* survey_template info rather than an *array* of survey_template info:

https://github.com/biocore/microsetta-private-api/blob/cd599ea218bbd00f3e6c4a7f98a11d4fce7588b3/microsetta_private_api/api/microsetta_private_api.yaml#L299-L325

AmandaBirmingham:
Have a fix ready to put into PR ... :)

AmandaBirmingham:
Fixed by #54

Issue #28, opened 2020-01-08
Make source_type optional for GET /accounts/{account_id}/sources


AmandaBirmingham:
To allow users to get all source types at once rather than having to loop over them

AmandaBirmingham:
Fixed by #54

PR #54, opened 2020-01-10
Fixes issue #39 and #28

AmandaBirmingham:
Fixes issue #39 by returning an *array* of survey templates from GET /accounts/{account_id}/sources/{source_id}/survey_template . Also refactors survey- and survey_template-related request and response schemas to reduce what used to be very heavy duplication (@dhakim87 check it out: adding a single field to an existing schema! :D ) NOTE that the only change caused by this refactor is addition of `survey_template_id` to the (answered) *`survey`*-related endpoints (which isn't a big change as these endpoints already included various info about the template the answered survey was based on--such as template name, etc.) .  This will need to be included when the survey and survey_template related implementation methods are brought into alignment with the api (currently they are pretty divergent).

Fixes issue #28 , making source_type optional for GET /accounts/{account_id}/sources .

### DIFF
--- a/microsetta_private_api/api/microsetta_private_api.yaml
+++ b/microsetta_private_api/api/microsetta_private_api.yaml
@@ -108,8 +108,8 @@ paths:
       operationId: microsetta_private_api.api.implementation.read_sources
       tags:
         - Sources
-      summary: Get sources associated with account, filtered by source type
-      description: Get sources associated with account, filtered by source type
+      summary: Get sources associated with account, filtered by source type if a source type is provided
+      description: Get sources associated with account, filtered by source type if a source type is provided
       parameters:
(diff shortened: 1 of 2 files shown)


microsetta_private_api/api/microsetta_private_api.yaml
```diff
@@ -989,7 +973,7 @@ components:
     source_id:
       type: string
       readOnly: true # sent in GET, not in POST/PUT/PATCH
-      example: "59688a9b-c664-4a0d-9125-ccdc04909c8e"
+      example: "b0b0b0b0-b0b0-b0b0-b0b0-b0b0b0b0b0b0"
```
dhakim87: 🐵 
Should we pick a better magic number for test uuids if we're going to expose them in our public api until the end of time?

PR #55, opened 2020-01-18
Support source creation from consent, remove kit password

### LINKED ISSUE #27: Determine and implement correct submission path for consent forms
~~age_range IS used to decide which consent text to return--and we didn't implement that logic in the api yaml OR the python implementation~~

Consent endpoint returns a form, and embedded in that form is the age range question, when then determines what div fields of the consent text are shown--so we are good on employing consent to figure out what to show.  

What is unclear now is where the consent form results should be sent and how ... do we set the form's submit url to send results to the create source api endpoint but change that endpoint to expect non-json input (e.g., openapi3 supports the `multipart/form-data` media type)?  Or do we do ... something else?  Right now consent form submissions don't go anywhere meaningful, AFAIK, so *some* change is necessary.

AmandaBirmingham:
Fixes #27, moving the consent endpoint under account and adding a post to enable creation of a human source from a submitted consent form.  

Unfortunately, those who live by the shiny new supporting library also die by the shiny new supporting library: although the openapi3 specs support allowing different media types to be passed into a single endpoint (which would have allowed us to add creating a human source from consent form data as another aspect of the /accounts/{account_id}/sources POST that already can take in json data), *`connexion`* doesn't currently implement the "multiple media types per endpoint" aspect of openapi3 (https://github.com/zalando/connexion/issues/655, https://github.com/zalando/connexion/issues/805 ).  Therefore it was necessary to kludge a little.  The new /accounts/{account_id}/consent endpoint has a GET that gets consent form html as well as, now, a **POST** that creates a human source from a submitted consent form (using ONLY the form data media type).  I am open to suggestions for better ways to do this. 

Because @dhakim87 has ongoing work on `source.py`, I can't carry through the full implementation of creating a source from the consent form, but I have verified I can parse the incoming data and toss it over to the existing `create_source` call.  ~~Note that, for some reason I can't fathom, connexion sends *all* form field values to the back end as lists, regardless of the input element type ... e.g.,~~

~~`{'age_range': ['7-12'], 'consent_child': ['Yes'], 'participant_name': ['a'], 'participant_email': ['b@c.com'], 'consent_witness': ['Yes'], 'obtainer_name': ['Me'], 'consent_parent': ['Yes'], 'parent_1_name': ['b'], 'parent_2_name': ['c'], 'deceased_parent': ['true']}`~~

~~... Given this, the main thing I do with the form data is ensure each of these lists is really just a single value and then dig it out to make, e.g.,~~

~~`{'age_range': '7-12', 'consent_child': 'Yes', 'participant_name': 'a', 'participant_email': 'b@c.com', 'consent_witness': 'Yes', 'obtainer_name': 'Me', 'consent_parent': 'Yes', 'parent_1_name': 'b', 'parent_2_name': 'c', 'deceased_parent': 'true'}`~~

 ~~... which is what I thought we should get anyway ...~~

This PR also fixes #27, removing use of kit_password from the yaml api, `implementation.py`, and `kit_repo.py`.

### DIFF
--- a/microsetta_private_api/api/implementation.py
+++ b/microsetta_private_api/api/implementation.py
@@ -288,13 +288,31 @@ def read_kit(kit_name):
         return jsonify(kit.to_api()), 200
 
 
-def consent_doc():
+def render_consent_doc(account_id):
     # return render_template("new_participant.jinja2",
     #                        message=MockJinja("message"),
     #                        media_locale=MockJinja("media_locale"),
     #                        tl=MockJinja("tl"))
 
+    # NB: Do NOT need to explicitly pass account_id into template for integration into form submission URL
+    # because form submit URL builds on the base of the URL that called it (which includes account_id)
     return render_template("new_participant.jinja2",
                            message=None,
                            media_locale=american_gut.media_locale,
                            tl=american_gut._NEW_PARTICIPANT)
+
+
+def create_human_source_from_consent(account_id, body):
+    parsed_form = dict()
+    for key, value in body.items():
+        if len(value) != 1:
+            raise ValueError("For form field '{0}', received unexpected values '{1}'".format(key, value))
+        parsed_form[key] = value[0]
+
+    # TODO: boolean deceased_parent will come through as
+    #  'deceased_parent': 'true'
+    # and will need to be converted to a boolean True ...
+    # think this should probably happen in source.HumanInfo.__init__
+
+    # NB: Don't expect to handle errors 404, 422 in this function; expect to farm out to `create_source`
+    return create_source(account_id, parsed_form)
--- a/microsetta_private_api/api/implementation.py
+++ b/microsetta_private_api/api/implementation.py
@@ -278,11 +278,11 @@ def dissociate_answered_survey(account_id, source_id, sample_id, survey_id):
     return not_yet_implemented()
 
 
-def read_kit(kit_name, kit_password):
+def read_kit(kit_name):
     with Transaction() as t:
         kit_repo = KitRepo(t)
-        # TODO: Ensure this name and password are what the repo layer expects
-        kit = kit_repo.get_kit(kit_name, kit_password)
+        # TODO: Ensure this name is what the repo layer expects
+        kit = kit_repo.get_kit(kit_name)
         if kit is None:
             return jsonify(error=404, text="No such kit"), 404
         return jsonify(kit.to_api()), 200
--- a/microsetta_private_api/api/implementation.py
+++ b/microsetta_private_api/api/implementation.py
@@ -294,8 +294,9 @@ def render_consent_doc(account_id):
     #                        media_locale=MockJinja("media_locale"),
     #                        tl=MockJinja("tl"))
 
-    # NB: Do NOT need to explicitly pass account_id into template for integration into form submission URL
-    # because form submit URL builds on the base of the URL that called it (which includes account_id)
+    # NB: Do NOT need to explicitly pass account_id into template for
+    # integration into form submission URL because form submit URL builds on the
+    # base of the URL that called it (which includes account_id)
     return render_template("new_participant.jinja2",
                            message=None,
                            media_locale=american_gut.media_locale,
@@ -306,7 +307,8 @@ def create_human_source_from_consent(account_id, body):
     parsed_form = dict()
     for key, value in body.items():
         if len(value) != 1:
-            raise ValueError("For form field '{0}', received unexpected values '{1}'".format(key, value))
+            raise ValueError("For form field '{0}', received unexpected values "
+                             "'{1}'".format(key, value))
         parsed_form[key] = value[0]
 
     # TODO: boolean deceased_parent will come through as
@@ -314,5 +316,6 @@ def create_human_source_from_consent(account_id, body):
     # and will need to be converted to a boolean True ...
     # think this should probably happen in source.HumanInfo.__init__
 
-    # NB: Don't expect to handle errors 404, 422 in this function; expect to farm out to `create_source`
+    # NB: Don't expect to handle errors 404, 422 in this function; expect to
+    # farm out to `create_source`
     return create_source(account_id, parsed_form)
--- a/microsetta_private_api/api/implementation.py
+++ b/microsetta_private_api/api/implementation.py
@@ -303,19 +303,12 @@ def render_consent_doc(account_id):
                            tl=american_gut._NEW_PARTICIPANT)
 
 
+# This function simply passes its arguments on to the create_source function;
+# however, it is necessary because connexion will not allow two endpoints to
+# use the same implementation function (it throws an error
+# "AssertionError: View function mapping is overwriting an existing endpoint
+# function")
 def create_human_source_from_consent(account_id, body):
-    parsed_form = dict()
-    for key, value in body.items():
-        if len(value) != 1:
-            raise ValueError("For form field '{0}', received unexpected values"
-                             " '{1}'".format(key, value))
-        parsed_form[key] = value[0]
-
-    # TODO: boolean deceased_parent will come through as
-    #  'deceased_parent': 'true'
-    # and will need to be converted to a boolean True ...
-    # think this should probably happen in source.HumanInfo.__init__
-
     # NB: Don't expect to handle errors 404, 422 in this function; expect to
     # farm out to `create_source`
-    return create_source(account_id, parsed_form)
+    return create_source(account_id, body)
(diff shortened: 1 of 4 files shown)


microsetta_private_api/api/implementation.py
```diff
@@ -278,23 +278,44 @@ def dissociate_answered_survey(account_id, source_id, sample_id, survey_id):
     return not_yet_implemented()
 
 
-def read_kit(kit_name, kit_password):
+def read_kit(kit_name):
     with Transaction() as t:
```
dhakim87: Hmm...  I know we aren't going to have kit_password in the future, but does this break migration for existing users?  Are our existing kit_name's secure enough to allow everyone to login by kit_name alone until they first register an account with us?  I wonder if we need a flag in the db for old_kit or new_kit to enable this...

microsetta_private_api/api/microsetta_private_api.yaml
```diff
@@ -103,6 +97,97 @@ paths:
...
+        content:
+          application/x-www-form-urlencoded:
+            schema:
+              properties:  # each is a form field name
+                age_range:
+                  type: array
+                  items:
+                    $ref: '#/components/schemas/age_range'
+                participant_name:
+                  type: array
```
dhakim87: Well I think I know why all your form encoded variables are coming in as lists 😀

microsetta_private_api/api/implementation.py
```diff
@@ -278,23 +278,44 @@ def dissociate_answered_survey(account_id, source_id, sample_id, survey_id):
...
     return render_template("new_participant.jinja2",
                            message=None,
                            media_locale=american_gut.media_locale,
                            tl=american_gut._NEW_PARTICIPANT)
+
+
+def create_human_source_from_consent(account_id, body):
+    parsed_form = dict()
+    for key, value in body.items():
+        if len(value) != 1:
```
dhakim87: This is silly 😛

microsetta_private_api/api/microsetta_private_api.yaml
```diff
@@ -103,6 +97,97 @@ paths:
...
+                  items:
+                    $ref: '#/components/schemas/participant_name'
+                participant_email:
+                  type: array
```
dhakim87: Array?

microsetta_private_api/api/microsetta_private_api.yaml
```diff
@@ -131,7 +216,7 @@ paths:
                   consent:
                     participant_name: "Ophelia Q. Doe"
                     participant_email: "nunnery@example.com"
-                    age_range: "18+"
+                    age_range: "18-plus"
```
dhakim87: Wait what the... 
Okay so apparently age_range can be any of [None, "", "0-6", "13-17", "18+", "7-12", "18-plus", "ANIMAL_SURVEY"].  I strongly vote we take this back to 18+ and restrict this enum in my migration scripts

microsetta_private_api/api/microsetta_private_api.yaml
```diff
@@ -978,6 +1030,29 @@ components:
...
+    participant_name:
+      type: string
+      example: "Hija Doe"
+    participant_email:
+      type: string
+      format: email
+      example: "mdoe@example.com"
+    age_range:
+      type: string
+      enum: ["0-6", "7-12", "13-17", "18-plus"]
```
dhakim87: "18-plus" or "18+" ???  I need this for my migration scripts

microsetta_private_api/templates/new_participant.jinja2
```diff
@@ -185,7 +185,7 @@
     <table>
         <tr><td>{{ tl['PARTICIPANT_PARENT_1'] |e}}</td><td><input tabindex="7" type="text" name="parent_1_name" id="7-12-form_parent_1_name" required data-rule-required="true"></td></tr>
         <tr><td>{{ tl['PARTICIPANT_PARENT_2'] |e}}</td><td><input tabindex="8" type="text" name="parent_2_name" id="7-12-form_parent_2_name" required data-rule-required="true"></td></tr>
-        <tr><td>{{ tl['PARTICIPANT_DECEASED_PARENTS'] |e}}</td><td><input  tabindex="9" type="checkbox" name="deceased_parent" id="deceased_parent" value="yes" onclick="relax_parents(this)"></td></tr>
+        <tr><td>{{ tl['PARTICIPANT_DECEASED_PARENTS'] |e}}</td><td><input  tabindex="9" type="checkbox" name="deceased_parent" id="deceased_parent" value="true" onclick="relax_parents(this)"></td></tr>
```
dhakim87: Honestly this is probably more a problem in the existing database than on this change... the deceased_parent column is a varchar(10) rather than a bool.  Do I need to change that in the migration?

microsetta_private_api/api/microsetta_private_api.yaml
AmandaBirmingham: No, I'm afraid you have the cause and effect reversed! :) That is not *why* they are coming in as lists--because they DO come in as lists from the form submit, they must be defined as lists in the api or else you get API validation errors (e.g., if you try to define them as single values in the API).  For example, when you submit the form, you get an error like `validation error: ['0-6'] is not of type 'string' - 'age_range'` unless you define age_range as an array in the API.

microsetta_private_api/api/microsetta_private_api.yaml
AmandaBirmingham: "18-plus" is the form value that is currently produced by the `new_participant.jinja2` template, and "18-plus" is also the value stored in the ag.ag_consent.age_range field (at least, in the test database), so I changed the yaml to match the rest of the ecosystem :) What necessitates changing over to "18+"?

microsetta_private_api/api/implementation.py
AmandaBirmingham: My understanding was that the decision had been made to take this out, so if that is being rethought, I gotta kick this one to @wasade  ....

microsetta_private_api/api/microsetta_private_api.yaml
AmandaBirmingham: @dhakim87 see above comment :)

microsetta_private_api/api/microsetta_private_api.yaml
AmandaBirmingham: See above :)

microsetta_private_api/api/implementation.py
AmandaBirmingham: More clarity here, please?

microsetta_private_api/templates/new_participant.jinja2
AmandaBirmingham: Yeah ... in the existing test agp db, the deceased_parent column is holding a terrifying mixture of "No", "false", "true", "yes" (note lack of capitalization), and null.  I can change this back to "yes" and we can treat it as an enum if this would make the db migration easier?

microsetta_private_api/api/microsetta_private_api.yaml
AmandaBirmingham: PS--I should perhaps clarify that the "coming back as lists" problem is handling somewhere in the under-the-hood handling on the flask/connexion side (but I don't know where).  What comes through in the POST request is something that looks like this:

`curl -X POST "http://localhost:8082/api/accounts/aaaaaaaa-bbbb-cccc-dddd-eeeeffffffff/consent" -H "accept: application/json" -H "Content-Type: application/x-www-form-urlencoded" -d "age_range=0-6&deceased_parent=true&obtainer_name=Professor%20X&parent_1_name=Madre%20Doe&parent_2_name=Padre%20Doe&participant_email=mdoe%40example.com&participant_name=Hija%20Doe"`

... which includes no list indicators.  However, sometime after the server receives this but *before* code drops into the actual handler that we write (in `implementation.py`), flask or connexion or some other library involved in the process interprets these as lists ... which then causes the connexion api validation (again, happening even before reaching our handler code) to fail.

microsetta_private_api/api/microsetta_private_api.yaml
dhakim87: You're right, there is exactly one instance of 18+ and 7470 instances of 18-plus, and that single 18+ is mr bobobobobobo.  So 18-plus it is.

AmandaBirmingham:
@dhakim87 , think we need an update on what is required before this PR can be approved.  AFAIK, the form field lists issue is fixed (thanks again :) and the 18+ vs 18-plus question has been resolved (in a way that doesn't require changes to this PR).  The two remaining issues I see are:

1) Are we really taking out kit_password or no? I think we are still waiting for feedback from @wasade  on this?
2) How should deceased_parent be represented? My current bid is to change it from a bool to an enum (so we can still load the existing freaky values but users can't make up new ones). Let me know if this works for you or if you want a different solution ...

microsetta_private_api/templates/new_participant.jinja2
dhakim87: Leave it as true, I'll fix this in the migration scripts and change the column to a boolean.

PR #83, opened 2020-03-03
Issue 81

### LINKED ISSUE #81: PUT /accounts/{account_id}/sources/{source_id} Broken?
I'm changing the fields specified in the example in /api/ui, but the object returned is always the starting object.  Maybe missing a commit() call?

dhakim87:
Fixes #81 

Looks like I missed implementing the PUT Source functionality.  Had to specify which fields they could edit, so I restricted them to just source_name and source_description for now.  Not sure how much of a pain it would be to specify just these fields in the yaml and everything else readonly.  

Also noticed the GET on source was never updated after we added the source_description fields, so we that's in now too.  

Unit tests for both.


wasade:
Per the slack conversation, it would be pleasant to be able to update a participants email but that is a future development item.

PR #88, opened 2020-03-04
Python 3.7 and ISO 8601 support

### LINKED ISSUE #65: Drop Python 3.6?
Do we have a reason to support python 3.6?  Some of the datetime formatting is cleaned up in 3.7 and may make our conversion code cleaner to make use of it.

dhakim87:
Fix #65


microsetta_private_api/api/tests/test_integration.py
```diff
@@ -835,6 +835,16 @@ def test_edit_sample_info(self):
         )
         check_response(response, 200)
 
+        invalid_date_fuzzy_info = copy.copy(fuzzy_info)
+        invalid_date_fuzzy_info['sample_datetime'] = "1-800-BAD-DATE"
```
wasade: 🎉

PR #98, opened 2020-04-11
Minimal interface: installation items identified during RC testing

wasade:
Adds non-python files that were not being included on install

Adds a test to verify the app can be imported and built from the installed entity 

Adds a missing `__init__.py` file to ensure items from `microsetta_private_api.model.vue` can be imported

PR #99, opened 2020-04-11
Migrate all handouts

wasade:
Migrate all handout kits so they work with the new kit registration system


wasade:
Build failures are due to flake8, and resolved in #98

wasade:
They were not, so should be good. Thanks!!

PR #103, opened 2020-04-15
(WIP) Minimal interface admin capability

dhakim87:
Laying out my proposal for administrator capabilities.  I'm using a separate implementation file with it's own version of access validation.  I would have preferred to use a separate yaml as well, but getting swagger and/or connexion to accept this yet let me share schema objects seems to be nigh impossible at the moment.  

Remains a Work In Progress until I have a baseline list of all admin routes implemented

### DIFF
--- a/microsetta_private_api/api/microsetta_private_api.yaml
+++ b/microsetta_private_api/api/microsetta_private_api.yaml
@@ -814,9 +814,9 @@ paths:
         '422':
           $ref: '#/components/responses/422UnprocessableEntity'
   # START ADMINISTRATOR PATHS
-  '/admin/samples/scan':
+  '/admin/search/sample/{sample_barcode}':
     get:
-      operationId: microsetta_private_api.admin.admin_impl.scan_barcode
+      operationId: microsetta_private_api.admin.admin_impl.search_barcode
       tags:
         - Admin
       summary: Retrieve diagnostic information about a sample by barcode
@@ -825,7 +825,7 @@ paths:
         - $ref: '#/components/parameters/sample_barcode'
       responses:
         '200':
-          description: Object containing linked account (if any), source (if any), and sample
+          description: Object containing linked account (if any), source (if any), sample (if any) and information about the barcode
           content:
             application/json:
               schema:
@@ -882,6 +882,14 @@ components:
       schema:
         $ref: '#/components/schemas/source_id'
       required: true
+    sample_barcode:
+      name: sample_barcode
+      in: path
+      description: barcode of sample
+      schema:
+        $ref: '#/components/schemas/sample_barcode'
+      required: true
+
 
     # query parameters
     kit_name:
@@ -912,13 +920,6 @@ components:
       description: Type of sample
       schema:
         $ref: '#/components/schemas/source_type'
-    sample_barcode:
-      name: sample_barcode
-      in: query
-      description: Scanned barcode of sample
-      schema:
-        $ref: '#/components/schemas/sample_barcode'
-      required: true
   responses:
     401Unauthorized:   # Can be referenced as '#/components/responses/401Unauthorized'
       description: Invalid or missing token.
--- a/microsetta_private_api/api/microsetta_private_api.yaml
+++ b/microsetta_private_api/api/microsetta_private_api.yaml
@@ -813,6 +813,27 @@ paths:
           $ref: '#/components/responses/404NotFound'
         '422':
           $ref: '#/components/responses/422UnprocessableEntity'
+  # START ADMINISTRATOR PATHS
+  '/admin/samples/scan':
+    get:
+      operationId: microsetta_private_api.admin.admin_impl.scan_barcode
+      tags:
+        - Admin
+      summary: Retrieve diagnostic information about a sample by barcode
+      description: Retrieve diagnostic information about a sample by barcode
+      parameters:
+        - $ref: '#/components/parameters/sample_barcode'
+      responses:
+        '200':
+          description: Object containing linked account (if any), source (if any), and sample
+          content:
+            application/json:
+              schema:
+                type: object
+        '401':
+          $ref: '#/components/responses/401Unauthorized'
+        '404':
+          $ref: '#/components/responses/404NotFound'
 
 
 components:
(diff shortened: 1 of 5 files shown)


microsetta_private_api/repo/admin_repo.py
```diff
@@ -0,0 +1,57 @@
+from microsetta_private_api.repo.account_repo import AccountRepo
+from microsetta_private_api.repo.base_repo import BaseRepo
+from microsetta_private_api.repo.sample_repo import SampleRepo
+from microsetta_private_api.repo.source_repo import SourceRepo
+
+
+class AdminRepo(BaseRepo):
```
wasade, line 7: Any chance we could start to pair these items going in with unit tests?

microsetta_private_api/repo/admin_repo.py
```diff
@@ -0,0 +1,57 @@
...
+                "ON "
+                "ag_kit_barcodes.source_id = source.id "
+                "LEFT OUTER JOIN "
+                "account "
+                "ON "
+                "account.id = source.account_id "
+                "WHERE "
+                "ag_kit_barcodes.barcode = %s",
+                (sample_barcode,))
+
```
wasade, line 31: In preparation particularly for scanning, it would be helpful to include the project information. It may be cleaner to do it as a second query but I'll defer to you. Any chance the project information could be obtained? An example showing a sample that is part of multiple projects:

```sql
ag=> select barcode, array_agg(project) 
from barcodes.barcode 
    join barcodes.project_barcode using(barcode) 
    join barcodes.project using (project_id) 
where barcode = '000074725' 
group by barcode;
-[ RECORD 1 ]--------------------------------------
barcode   | 000074725
array_agg | {"American Gut Project",Ornish_Program}
```

wasade: Note: if this introduces other complexity like needing to modify the OpenAPI schema and other consequences, then it may be better to defer this addition until we implement sample scanning.

microsetta_private_api/api/microsetta_private_api.yaml
```diff
@@ -813,6 +813,27 @@ paths:
           $ref: '#/components/responses/404NotFound'
         '422':
           $ref: '#/components/responses/422UnprocessableEntity'
+  # START ADMINISTRATOR PATHS
+  '/admin/samples/scan':
```
wasade, line 817: Would it make sense for this to instead be `/admin/search` where it can accept ONE OF `kit_id=foo` or `sample_barcode=foo` or `name=foo` or `email=foo`?

Our immediate target was search by barcode, but just keeping an eye toward the next steps.

wasade: Scratch that. I think having endpoints specific for the search is appropriate as the type of information returned reasonably varies.

dhakim87: Either way is fine with me, one route or many doesn't really change the underlying code.  I'd probably avoid anything that requires a query language though, converting anything complex into sql would increase scope quite a bit.

microsetta_private_api/repo/admin_repo.py
dhakim87: Easily done, I'll add it to the returned model

wasade: Thanks!!

microsetta_private_api/api/microsetta_private_api.yaml
wasade: Agree avoiding a query language

microsetta_private_api/repo/admin_repo.py
dhakim87: One thing I've been meaning to bring up for a very long while...  I've found that a lot of the unit tests we write (not necessarily within microsetta-private-api, but in general as I've played around with our code bases) are calling functions to check that behavior hasn't changed, rather than to check that behavior is correct.  It's a subtle distinction, but when we're adding tests to increase code coverage rather than adding tests to check a specific api is maintained, I think we're doing unit testing incorrectly.  It increases maintenance time when the functions are modified (as all the tests must be modified as well) without actually providing validation of functionality.  

The fix for that is laying out a specification of what the admin routes (or the adminrepo class, or any other module under test) need to accomplish.  I don't consider the administrative functionality to be part of the microsetta-private-api.  It can be versioned separately.  The modules it interacts with are completely separate.  The schemas of the objects it returns aren't fixed, and I'm not sure they need to be (pending discussion on how your module will interact with it)  I can't really test it very well until I know what its supposed to do more specifically.  

tl;dr: I'll call the function and make sure it returns something, that'll at least catch when the database changes drastically.  But I don't like doing any introspection of what it returns because what it returns hasn't been locked down yet.

microsetta_private_api/repo/admin_repo.py
dhakim87: Actually, I'm confused, the Sample model object already has sample_projects in it.  That whole object should be loaded in by the sample_repo._get_sample_by_id call.  Is it sufficient to just return the whole sample object as I'm currently doing?

wasade: I think so

microsetta_private_api/repo/admin_repo.py
wasade: Thanks. We should have a good idea of, for example, what `retrieve_diagnostics_by_barcode` does when presented `000004216`, and a good idea of what it will do when presented `a-bad-barcode`. It should not be difficult to determine the result seems correct without doing deep introspection. The intent with those is not coverage, but basic positive/negative controls. This does not enumerate edge cases, but it can catch a lot of unexpected behavior over the life of the codebase

microsetta_private_api/api/microsetta_private_api.yaml
wasade: I think this endpoint describes a searching not scanning btw, so `/admin/search/sample`. Would it make sense for this to be `/admin/search/sample/{sample_barcode}`?

wasade:
Thanks!!!

PR #105, opened 2020-04-16
Endpoint for creating kits

wasade:
Kit creation. This ended up more complex than initially intended. 

Previously, the notion of a "kit" was limited to the `ag` schema as only AGP projects had collections of samples. The TMI "box" is really a kit, and we intend to ship to other projects, which *may not* use our participant portal. We will also be receiving FedEx tracking information on kits shipped. SO, there is now a `kit`s table in `barcodes`. References updated, and existing kit names are backported. 

To facilitate tracking of shipments, our logistics partner is putting a QR code on the outside of the kits, which they will scan and send to us (exact mechanism of "sending to us" is not final). That code is `kit_uuid` in `barcodes.kit` right now. 

I do not think kits *need* to live in both `barcodes.kit` and `ag.ag_kit`, however I think adapting the database to address is a major refactor.

### DIFF
--- a/microsetta_private_api/admin/admin_impl.py
+++ b/microsetta_private_api/admin/admin_impl.py
@@ -36,10 +36,13 @@ def create_kits(body, token_info):
 
     with Transaction() as t:
         admin_repo = AdminRepo(t)
-        kits = admin_repo.create_kits(number_of_kits, number_of_samples,
-                                      kit_prefix, projects)
 
-        if kits:
+        try:
+            kits = admin_repo.create_kits(number_of_kits, number_of_samples,
+                                          kit_prefix, projects)
+        except KeyError:
+            return jsonify(code=422, message="Unable to create kits"), 422
+        else:
             t.commit()
 
     return jsonify(kits), 201
--- a/microsetta_private_api/admin/admin_impl.py
+++ b/microsetta_private_api/admin/admin_impl.py
@@ -24,3 +24,22 @@ def validate_admin_access(token_info):
                                                    token_info['sub'])
         if account is None or account.account_type != 'admin':
             raise Unauthorized()
+
+
+def create_kits(body, token_info):
+    validate_admin_access(token_info)
+
+    number_of_kits = body['number_of_kits']
+    number_of_samples = body['number_of_samples']
+    kit_prefix = body.get('kit_prefix', None)
+    projects = body['projects']
+
+    with Transaction() as t:
+        admin_repo = AdminRepo(t)
+        kits = admin_repo.create_kits(number_of_kits, number_of_samples,
+                                      kit_prefix, projects)
+
+        if kits:
+            t.commit()
+
+    return jsonify(kits), 201
--- a/microsetta_private_api/api/microsetta_private_api.yaml
+++ b/microsetta_private_api/api/microsetta_private_api.yaml
@@ -875,6 +875,10 @@ paths:
 
         '401':
           $ref: '#/components/responses/401Unauthorized'
+        
+        '422':
+          $ref: '#/components/responses/422UnprocessableEntity'
+
 
 components:
   parameters:
--- a/microsetta_private_api/api/microsetta_private_api.yaml
+++ b/microsetta_private_api/api/microsetta_private_api.yaml
@@ -813,6 +813,7 @@ paths:
           $ref: '#/components/responses/404NotFound'
         '422':
           $ref: '#/components/responses/422UnprocessableEntity'
+
   # START ADMINISTRATOR PATHS
   '/admin/search/sample/{sample_barcode}':
     get:
@@ -835,6 +836,45 @@ paths:
         '404':
           $ref: '#/components/responses/404NotFound'
 
+  '/admin/create/kits':
+    post:
+      operationId: microsetta_private_api.admin.admin_impl.create_kits
+      tags:
+        - Admin
+      summary: Create kit identifiers and associated sample barcodes
+      description: Create kit identifiers and associated sample barcodes
+      requestBody:
+        content:
+          application/json:
+            schema:
+                type: "object"
+                properties:
+                  number_of_kits:
+                    type: integer
+                  number_of_samples:
+                    type: integer
+                  kit_id_prefix:
+                    type: string
+                  projects:
+                    type: array
+                    items:
+                      type: string
+
+                required:
+                  - number_of_kits
+                  - number_of_samples
+                  - projects
+
+      responses:
+        '201':
+          description: Kit identifiers and associated samples were successfully created
+          content:
+            application/json:
+              schema:
+                type: object
+
+        '401':
+          $ref: '#/components/responses/401Unauthorized'
 
 components:
   parameters:
@@ -989,6 +1029,7 @@ components:
         - last_name
         - email
         - address
+
     address:   # taken from https://opensource.zalando.com/restful-api-guidelines/#address-fields
       description:
         an address of a location/destination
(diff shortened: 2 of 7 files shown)


microsetta_private_api/admin/admin_impl.py
```diff
@@ -24,3 +24,22 @@ def validate_admin_access(token_info):
...
+
+    with Transaction() as t:
+        admin_repo = AdminRepo(t)
+        kits = admin_repo.create_kits(number_of_kits, number_of_samples,
+                                      kit_prefix, projects)
+
+        if kits:
+            t.commit()
+
+    return jsonify(kits), 201
```
dhakim87, line 45: Appears kits may be None/Falsy according to the logic above.  If this is possible, that should be checked to prevent returning a 201.

microsetta_private_api/api/microsetta_private_api.yaml
```diff
@@ -835,6 +836,45 @@ paths:
...
+                    type: array
+                    items:
+                      type: string
+
+                required:
+                  - number_of_kits
+                  - number_of_samples
+                  - projects
+
+      responses:
```
dhakim87, line 868: Possibly need a 422 if kits fail to be created?  Appears there are code paths that result in KeyError (which probably becomes a 500 right now?  But I'm not sure if these paths are plausible or require absurdly bad luck and key collisions.

microsetta_private_api/db/patches/0055.sql
```diff
@@ -0,0 +1,40 @@
...
+-- A note on projects: previously, barcodes were associated with projects not
+-- kits. This was necessary as non-AGP projects did not have a notion of a kit.
+
+-- With TMI kits, we are also now positioned to (for many kits) get fedex tracking 
+-- information, so lets us up for that...
+CREATE TABLE barcodes.kit (
+    kit_uuid uuid DEFAULT uuid_generate_v4() NOT NULL,
+    kit_id VARCHAR NOT NULL CONSTRAINT unq_kit_id UNIQUE,
+    fedex_tracking VARCHAR NULL,
+    address VARCHAR NULL,
```
dhakim87, line 16: Using address as a string here, when it appears as street/city/state/post_code/country_code in ag.account will have us implement addresses twice.  Unifying may be a concern for the future, assuming these addresses can be unified at all, depending on their sources and how they are processed.

microsetta_private_api/repo/admin_repo.py
```diff
@@ -75,3 +79,140 @@ def retrieve_diagnostics_by_barcode(self, sample_barcode):
...
+    def _generate_random_kit_name(self, name_length, prefix):
+        if prefix is None:
+            prefix = 'tmi_'
+
+        # O, o, S and l removed to improve readability
+        chars = 'abcdefghjkmnpqrstuvwxyz' + 'ABCDEFGHJKMNPQRTUVWXYZ'
+        chars += string.digits
+        rand_name = ''.join(random.choice(chars)
+                            for i in range(name_length))
+        return prefix + '_' + rand_name
```
dhakim87, line 92: This, with the pref "tmi_" will result in a double underscore.  Is this intentional?

microsetta_private_api/repo/admin_repo.py
```diff
@@ -75,3 +79,140 @@ def retrieve_diagnostics_by_barcode(self, sample_barcode):
             }
 
             return diagnostic
+
+    def _generate_random_kit_name(self, name_length, prefix):
+        if prefix is None:
+            prefix = 'tmi_'
+
+        # O, o, S and l removed to improve readability
```
dhakim87, line 87: Are capital i and the number 1 distinguishable on the printed kits?

microsetta_private_api/repo/admin_repo.py
```diff
@@ -75,3 +79,140 @@ def retrieve_diagnostics_by_barcode(self, sample_barcode):
...
+                        "FROM barcodes.project")
+            known_projects = {prj.lower(): id_ for prj, id_ in cur.fetchall()}
+            for name in projects:
+                if name.lower() not in known_projects:
+                    raise KeyError("%s does not exist" % name)
+
+            # get existing kits to test for conflicts
+            cur.execute("""SELECT kit_id FROM barcodes.kit""")
+            existing = set(cur.fetchall())
+            names = [self._generate_random_kit_name(8, kit_prefix)
```
dhakim87, line 128: Looks like entropy of about 2^48 in each kit name.  (<2^6 bits per letter, 8 letters)  This is brute forceable in an offline setting, so we will need to rate limit.  I'm guessing we can't up it to 15 or 20 letters because it won't fit on the piece of paper?

microsetta_private_api/repo/admin_repo.py
```diff
@@ -1,3 +1,7 @@
+import uuid
+import string
+import random
```
dhakim87, line 3: Consider import secrets over import random, the kit names must be unguessable, and I'm not sure what guarantees the random library makes about seeding.

microsetta_private_api/repo/admin_repo.py
```diff
@@ -75,3 +79,140 @@ def retrieve_diagnostics_by_barcode(self, sample_barcode):
...
+            existing = set(cur.fetchall())
+            names = [self._generate_random_kit_name(8, kit_prefix)
+                     for i in range(number_of_kits)]
+
+            # if we observe ANY conflict, lets bail. This should be extremely
+            # rare, from googling seemed a easier than having postgres
+            # generate a unique identifier that was reasonably short, hard to
+            # guess
+            if len(set(names) - existing) != number_of_kits:
+                raise KeyError("Conflict in created names, kits not created")
```
dhakim87, line 136: Birthday paradox says ~50% chance of conflict when we hit 2^24 kits.  So maybe not super high priority to clean up these error conditions, but one day we'll get there :D

microsetta_private_api/db/patches/0055.sql
wasade: This is an unknown right now. Not sure what the address detail will look like. We can drop this and the fedex field until we know more, and defer to a later patch?

microsetta_private_api/repo/admin_repo.py
wasade: No actual difference for `random.choice`... see [here](https://github.com/python/cpython/blob/3.8/Lib/secrets.py#L20-L25)

microsetta_private_api/repo/admin_repo.py
wasade: lets omit, good call

wasade: ...actually it was omitted, but lowercase "i" was not. Updated doc string to reflect that, and added "i" in

microsetta_private_api/repo/admin_repo.py
wasade: nope good catch

microsetta_private_api/repo/admin_repo.py
wasade: This doesn't need to be cryptographic, just obfuscated. And we should rate limit regardless

microsetta_private_api/repo/admin_repo.py
wasade: I stared at SQL code to do something like this an auto-increment, and it looked like a pit of death. UUID would be great but it would piss off participants due to length. We could (now or in the future) turn this into a stupid while loop?

microsetta_private_api/api/microsetta_private_api.yaml
wasade: added

microsetta_private_api/admin/admin_impl.py
wasade: Done

Issue #104, opened 2020-04-16
Admin endpoints


wasade:
The administrative interface needs additional ways to get information in and out of the database. The  The following are the endpoints known about at this time. A numbering of 1-3 is used to indicate priority where 1 is most urgent, 3 is least urgent.

Sample search:
- [ ] (1) `GET /admin/search/barcode/{barcode}` returning detailed information about the sample including sample type, kit ID, account ID, name, email, etc
- [ ] (3) `GET /admin/search/kit_id/{kit_id}` returning detailed information about the kit including the samples contained, account ID, any source IDs, etc
- [ ] (2) `GET /admin/search/email/{email}` returning detailed information about the associated account including kits and sample information

Sample scanning will also need the `GET .../{barcode}` endpoint implemented for search. It will also need:
- [ ] (1) `POST /admin/scan/{barcode}` with a body containing `{'sample_status': <sample_status enum>, 'technician_notes': str}`. These data should be recorded in the `barcodes` schema, and a `scan_date` entry in the database should get a timestamp of `NOW`. This recorded information should be obtainable from the search `GET .../{barcode}` described above. Sample status values are: `{not-received, sample-is-valid, no-associated-consent, no-registered-account, sample-has-inconsistencies}`

Kit creation:
- [ ] (2) `POST /admin/create/kits` with a body containing `{'number_of_kits': <int>, 'samples_per_kit': <int>, 'kit_id_prefix': <str> | None, 'projects': <[str]>}`. This should trigger the addition of entries into the `barcodes` schema. If "American Gut Project" is one of the project names, appropriate entries should be created in `ag.ag_kit` and `ag.ag_kit_barcodes`. If `kit_id_prefix` is specified, that string should be prefixed to all created kit IDs. Barcode IDs created *SHOULD NOT* consist only of digits, but should always include a non-numeric character to avoid programs like Excel accidentally assuming these are numbers. This endpoint should return `{'created': [{'kit_id': <str>, 'sample_barcodes': <[str]>, 'kit_uuid': <uuid>]}`. The `kit_uuid` needs to be safe to share; this ID will be visible on the outside of the kit. All created samples get an initial sample status of `not-received`. 

Kit tracking. We will have the ability to obtain the FedEx tracking information, as well as the exact address a kit was shipped too. Recording this will be very useful.
[...]

PR #106, opened 2020-04-16
Admin search functionality

dhakim87:
Added admin ability to search accounts by email and search kits by supplied kit id.  Added kit information when searching for barcodes for #104

### DIFF
--- a/microsetta_private_api/api/microsetta_private_api.yaml
+++ b/microsetta_private_api/api/microsetta_private_api.yaml
@@ -834,6 +834,57 @@ paths:
           $ref: '#/components/responses/401Unauthorized'
         '404':
           $ref: '#/components/responses/404NotFound'
+  '/admin/search/kit/{kit_id}':
+    get:
+      operationId: microsetta_private_api.admin.admin_impl.search_kit_id
+      tags:
+        - Admin
+      summary: Retrieve diagnostic information about a sample by barcode
+      description: Retrieve diagnostic information about a sample by barcode
+      parameters:
+        - in: path
+          name: kit_id
+          description: User-facing id of kit
+          schema:
+            $ref: '#/components/schemas/kit_name'
+          required: true
+      responses:
+        '200':
+          description: Object containing linked account (if any), source (if any), sample (if any) and information about the barcode
+          content:
+            application/json:
+              schema:
+                type: object
+        '401':
+          $ref: '#/components/responses/401Unauthorized'
+        '404':
+          $ref: '#/components/responses/404NotFound'
+  '/admin/search/account/{email}':
+    get:
+      operationId: microsetta_private_api.admin.admin_impl.search_email
+      tags:
+        - Admin
+      summary: Retrieve diagnostic information about an account by email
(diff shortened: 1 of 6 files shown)


microsetta_private_api/repo/admin_repo.py
```diff
@@ -74,4 +83,63 @@ def retrieve_diagnostics_by_barcode(self, sample_barcode):
...
+            'sample_diagnostic_info': sample_assoc
+        }
+
+        return diagnostic
+
+    def retrieve_diagnostics_by_email(self, email):
+
+        # TODO: I leaned toward making this a way to discover what account to
+        #  look into rather than giving information about all accounts with
+        #  similar emails.  Does this approach work?  -DH
```
wasade, line 136: :+1:

microsetta_private_api/api/microsetta_private_api.yaml
```diff
@@ -834,6 +834,57 @@ paths:
           $ref: '#/components/responses/401Unauthorized'
         '404':
           $ref: '#/components/responses/404NotFound'
+  '/admin/search/kit/{kit_id}':
+    get:
+      operationId: microsetta_private_api.admin.admin_impl.search_kit_id
+      tags:
+        - Admin
+      summary: Retrieve diagnostic information about a sample by barcode
```
wasade, line 842: Copypasta, this is by kit id :)

microsetta_private_api/api/microsetta_private_api.yaml
```diff
@@ -834,6 +834,57 @@ paths:
...
+          schema:
+            $ref: '#/components/schemas/kit_name'
+          required: true
+      responses:
+        '200':
+          description: Object containing linked account (if any), source (if any), sample (if any) and information about the barcode
+          content:
+            application/json:
+              schema:
+                type: object
```
wasade, line 857: Goal is to define these later right? If so, possible to create an issue?

dhakim87: I'm torn on this.  If the admin user interface is parsing a fixed json schema, then yes we need to define what it is.  If that interface is just displaying general json, then I can just return whatever debugging information I have.  

Defining these schemas in yaml is semi painful up front, and will entail a maintenance cost in keeping the models and any to_api adaptation up to date when our objects are changed.  

For the various GET operations in the admin interface, I don't see a particular benefit to fully specifying the schema.  There is the option of a partial schema object in the yaml, guaranteeing the existence of several fixed fields and leaving any additional information open.  For the POSTs which must write exact information to the table, I think we need full specification.

wasade: The admin interface will need to make at least some expectations on this structure. If they chance, then we may silently break the admin interface. So I think we incur pain here or there. Ultimately, the royal we are maintaining this, so if you think there is less pain in manipulating jinja, then lets proceed as is

dhakim87: Created #111

PR #107, opened 2020-04-17
St dh admin scan sample

dhakim87:
Adds sample scanning with updates for sample_status, technician_notes and scan_date.  Includes a schema change on the barcodes.barcode table.  

Applies for one checkbox in #104

### DIFF
--- a/microsetta_private_api/admin/tests/test_admin.py
+++ b/microsetta_private_api/admin/tests/test_admin.py
@@ -1,6 +1,6 @@
 from unittest import TestCase
 
-from werkzeug.exceptions import Unauthorized
+from werkzeug.exceptions import Unauthorized, NotFound
 
 from microsetta_private_api.model.account import Account
 from microsetta_private_api.model.address import Address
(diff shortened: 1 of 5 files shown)


microsetta_private_api/admin/tests/test_admin.py
```diff
@@ -121,3 +121,44 @@ def test_search_barcode(self):
...
+
+            diag = admin_repo.retrieve_diagnostics_by_barcode(TEST_BARCODE)
+            self.assertEqual(diag['barcode_info'][0]['technician_notes'],
+                             TEST_NOTES)
+            self.assertEqual(diag['barcode_info'][0]['sample_status'],
+                             TEST_STATUS)
+            self.assertEqual(diag['barcode_info'][0]['status'],
+                             prestatus)
+
+            try:
```
wasade, line 154: equiv to:

```python
with self.assertRaises(NotFound):
    admin_repo.scan_barcode(...)
```

microsetta_private_api/repo/admin_repo.py
```diff
@@ -75,3 +79,31 @@ def retrieve_diagnostics_by_barcode(self, sample_barcode):
             }
 
             return diagnostic
+
+    def scan_barcode(self, sample_barcode, scan_info):
+        update_args = (
+            scan_info['sample_status'],
+            scan_info['technician_notes'],
+            date.today(),  # TODO: Do we need date or datetime here?
```
wasade, line 87: datetime would be good

dhakim87: I checked the database and scan_date was expecting a date rather than a datetime.  Did we want to change it to a datetime as part of the schema change, or should I leave it as a date?

wasade: you're correct. path of least resistance: let's stick with date

PR #109, opened 2020-04-17
Legacy acct support

AmandaBirmingham:
Add support to allow users who created accounts before authrocket to "claim" their existing accounts with their new authrocket account

### DIFF
--- a/microsetta_private_api/api/implementation.py
+++ b/microsetta_private_api/api/implementation.py
@@ -22,7 +22,7 @@ from microsetta_private_api import localization
 from microsetta_private_api.model.address import Address
 from microsetta_private_api.model.sample import SampleInfo
 from microsetta_private_api.repo.transaction import Transaction
-from microsetta_private_api.repo.account_repo import AccountRepo, RepoException
+from microsetta_private_api.repo.account_repo import AccountRepo
 from microsetta_private_api.repo.source_repo import SourceRepo
 from microsetta_private_api.repo.kit_repo import KitRepo
 from microsetta_private_api.repo.survey_template_repo import SurveyTemplateRepo
@@ -632,7 +632,7 @@ def validate_access(token_info, account_id):
             # is_auth can return True, None, or False, and *both*
             # True and None are authorized values (legacy records,
             # long story). *ONLY* explicitly false values should error here.
-            if is_auth == False:
+            if is_auth is False:
                 return Unauthorized()
 
         return account
--- a/microsetta_private_api/api/implementation.py
+++ b/microsetta_private_api/api/implementation.py
@@ -69,6 +69,29 @@ def find_accounts_for_login(token_info):
         return jsonify([acct.to_api()]), 200
 
 
+def claim_legacy_acct(token_info):
+    # If there exists a legacy account for the email in the token, which the
+    # user represented by the token does not already own but can claim, this
+    # claims the legacy account for the user and returns a 200 code with json
+    # containing an object for the claimed account.  Otherwise, this returns a
+    # 404 code. This function can also trigger a 422 from the repo layer in the
+    # case of inconsistent account data.
+
+    email = token_info['email']
+    auth_iss = token_info[JWT_ISS_CLAIM_KEY]
+    auth_sub = token_info[JWT_SUB_CLAIM_KEY]
+
+    with Transaction() as t:
+        acct_repo = AccountRepo(t)
+        acct = acct_repo.claim_legacy_account(email, auth_iss, auth_sub)
+        t.commit()
+
+        if acct is None:
+            return jsonify(code=404, message=ACCT_NOT_FOUND_MSG), 404
+
+        return jsonify(acct.to_api()), 200
+
+
 def register_account(body, token_info):
     # First register with AuthRocket, then come here to make the account
     new_acct_id = str(uuid.uuid4())
--- a/microsetta_private_api/api/implementation.py
+++ b/microsetta_private_api/api/implementation.py
@@ -22,7 +22,7 @@ from microsetta_private_api import localization
 from microsetta_private_api.model.address import Address
 from microsetta_private_api.model.sample import SampleInfo
 from microsetta_private_api.repo.transaction import Transaction
-from microsetta_private_api.repo.account_repo import AccountRepo
+from microsetta_private_api.repo.account_repo import AccountRepo, RepoException
 from microsetta_private_api.repo.source_repo import SourceRepo
 from microsetta_private_api.repo.kit_repo import KitRepo
 from microsetta_private_api.repo.survey_template_repo import SurveyTemplateRepo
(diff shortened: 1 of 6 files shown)


microsetta_private_api/api/implementation.py
```diff
@@ -595,12 +612,27 @@ def verify_authrocket(token):
...
         account_repo = AccountRepo(t)
         account = account_repo.get_account(account_id)
-        if account is None or \
-           'iss' not in token_info or \
-           'sub' not in token_info or \
-                account.auth_issuer != token_info['iss'] or \
-                account.auth_sub != token_info['sub']:
-            raise Unauthorized()
+
+        if account is not None:
```
dhakim87, line 627: Trying to parse the logic in my head, I think there's a slight change in behavior here.  It looks like if someone passes in a gibberish account_id such that account_repo.get_account returns None, it will no longer throw Unauthorized.  Granted it should throw NotFound doing any of the repo layer operations that require an account id, but in the case of some dumb repo code that didn't technically need the account_id to be correct (because we have uuids at all layers), that could open us up ever so slightly.  

Maybe add an 
else:
    raise Unauthorized()

microsetta_private_api/api/implementation.py
```diff
@@ -595,12 +612,27 @@ def verify_authrocket(token):
...
+            is_auth = account.account_matches_auth(
+                token_info['email'], token_info[JWT_ISS_CLAIM_KEY],
```
dhakim87, line 629: Hmm...  Technically the only thing we are allowed to key on in tokens is the 'sub' claim.  This will open us up to the issue that a user changes their email in AuthRocket and then is no longer allowed to access their account in our system, even though AuthRocket reports they are the same user.  

Ah.  I already see a TODO for this in account_matches_auth.  I guess fine as long as it is known.  Potentially we should make an issue, or just wait for a user to actually do this, which may take years to occur.

microsetta_private_api/api/implementation.py
```diff
@@ -595,12 +612,27 @@ def verify_authrocket(token):
...
+                token_info[JWT_SUB_CLAIM_KEY])
+            # Note: DO NOT pythonify this to `if not is_auth`!
+            # is_auth can return True, None, or False, and *both*
```
dhakim87, line 632: This is confusing, can we return an enum { FULL_MATCH, LEGACY_MATCH, FAIL } or something so that this is more clear?

microsetta_private_api/repo/account_repo.py
```diff
@@ -50,6 +50,57 @@ def _account_to_row(a):
...
+            # any other situation is an error and shouldn't happen,
+            # e.g. one of auth_iss or auth_sub is null in db but the other
+            # isn't, or one or more of non-null auth_iss and auth_sub values
+            # in db do not match the analogous input auth_iss and auth_sub
+            # values for the provided email ... may be more edge cases as well
+            raise RepoException("Inconsistent data found for provided email.")
+
+    def _find_account_by_email(self, email):
+        # select from account table anything that has this email.
+        # TODO: Do we need to check for multiple results?
```
dhakim87, line 89: I don't think, unless email is somehow None.  idx_account_email on the account table guarantees uniqueness.

microsetta_private_api/api/implementation.py
```diff
@@ -69,6 +69,29 @@ def find_accounts_for_login(token_info):
...
+    email = token_info['email']
+    auth_iss = token_info[JWT_ISS_CLAIM_KEY]
+    auth_sub = token_info[JWT_SUB_CLAIM_KEY]
+
+    with Transaction() as t:
+        acct_repo = AccountRepo(t)
+        acct = acct_repo.claim_legacy_account(email, auth_iss, auth_sub)
+        t.commit()
+
+        if acct is None:
```
wasade, line 89: Do lines 89-92 need to happen within the `Transaction` context?

microsetta_private_api/api/implementation.py
wasade: Agree an `enum` may be nice here as the response is not assured to be boolean

microsetta_private_api/repo/account_repo.py
```diff
@@ -50,6 +50,57 @@ def _account_to_row(a):
...
 
+    def claim_legacy_account(self, email, auth_iss, auth_sub):
+        # Returns now-claimed legacy account if an unclaimed legacy account
+        # that matched the input email was found; otherwise returns None.
+        # (Note that None is returned in the case where there is a NON-legacy
+        # account with the input email--find such accounts with
+        # find_linked_account instead.) Throws a RepoException
+        # if logic indicates inconsistent auth info.
+
+        # TODO: do I need to validate any of these inputs?
```
wasade, line 61: It looks like there are validations on line 68 (`account_matches_auth`), or are there others that may make sense to consider?

microsetta_private_api/api/implementation.py
AmandaBirmingham: @dhakim87 per our discussion on Slack, I've retained the 404 (rather than 401) for not-found accounts, as specified in the API, but have moved the handling for that into `validate_access` so as to prevent the sort of exploit you described where a user could pass in a non-valid account and manage to get real info if the handler function didn't itself specifically test if the found account was none.  Please let me know if you still have concerns about this!

microsetta_private_api/api/implementation.py
AmandaBirmingham: Have created an issue to deal with users changing their email in authrocket but not in microsetta at https://github.com/biocore/microsetta-private-api/issues/113

microsetta_private_api/repo/account_repo.py
AmandaBirmingham: Yay!  Removed TODO comment.

microsetta_private_api/api/implementation.py
AmandaBirmingham: No, it doesn't, assuming I understand the difference between context and scoping correctly :) .  However, right now the idiom in `implementation.py` is generally that the response is issued from within the transaction context (e.g., see `find_accounts_for_login`, `read_account`, `update_account`, `read_sources`, etc) so I am loath to change this one and not others.  @wasade would you like me to change them all?

microsetta_private_api/repo/account_repo.py
AmandaBirmingham: I'm going to take that as a no :)

microsetta_private_api/api/implementation.py
AmandaBirmingham: Changed to enum

microsetta_private_api/api/implementation.py
wasade: Okay thanks, consistency is preferable. I doubt this would have any practical consequence

dhakim87: Ugh.  Yeah, about half the routes I just looked at do that...  That's silly, no reason to keep a lock on the database while building response objects.  No practical concern until we hit a bottleneck on database access, but I'll make a separate issue to deal with this.

PR #110, opened 2020-04-17
St dh minimal interface admin sample pulldown

dhakim87:
Sample Pulldown

### DIFF
--- a/microsetta_private_api/admin/admin_impl.py
+++ b/microsetta_private_api/admin/admin_impl.py
@@ -17,6 +17,29 @@ def search_barcode(token_info, sample_barcode):
         return jsonify(diag), 200
 
 
+def sample_pulldown_single_survey(token_info,
+                                  sample_barcode,
+                                  survey_template_id):
+    validate_admin_access(token_info)
+
+    with Transaction() as t:
+        admin_repo = AdminRepo(t)
+        sample_pulldown = admin_repo.get_survey_metadata(
+            sample_barcode,
+            survey_template_id=survey_template_id)
+    return jsonify(sample_pulldown), 200
+
+
+def sample_pulldown_multiple_survey(token_info,
+                                    sample_barcode):
+    validate_admin_access(token_info)
+
+    with Transaction() as t:
+        admin_repo = AdminRepo(t)
+        sample_pulldown = admin_repo.get_survey_metadata(sample_barcode)
+    return jsonify(sample_pulldown), 200
+
+
 def validate_admin_access(token_info):
     with Transaction() as t:
         account_repo = AccountRepo(t)
--- a/microsetta_private_api/admin/tests/test_admin.py
+++ b/microsetta_private_api/admin/tests/test_admin.py
@@ -1,6 +1,6 @@
 from unittest import TestCase
 
-from werkzeug.exceptions import Unauthorized
+from werkzeug.exceptions import Unauthorized, NotFound
 
 from microsetta_private_api.model.account import Account
 from microsetta_private_api.model.address import Address
@@ -121,3 +121,39 @@ class AdminTests(TestCase):
             self.assertIsNone(diag['source'])
             self.assertIsNone(diag['sample'])
             self.assertEqual(len(diag['barcode_info']), 0)
+
+    def test_get_survey(self):
+        with Transaction() as t:
+            admin_repo = AdminRepo(t)
+
+            BARCODE = '000004216'
+
+            try:
+                nometa = admin_repo.get_survey_metadata("NOTABARCODE")
+                self.fail("Should not find NOTABARCODE")
+            except NotFound:
+                pass
+
+            meta = admin_repo.get_survey_metadata(BARCODE,
+                                                  survey_template_id=1)
+
+            self.assertEqual(meta['sample_barcode'], BARCODE)
+            self.assertIn('host_subject_id', meta)
+            # And there should be one survey answered
+            self.assertEqual(len(meta['survey_answers']), 1)
+
+            all_meta = admin_repo.get_survey_metadata(BARCODE)
+
+            self.assertEqual(all_meta['sample_barcode'], BARCODE)
+            self.assertEqual(all_meta['host_subject_id'],
+                             all_meta['host_subject_id'])
+            # And there should be more than one survey answered
+            self.assertGreater(len(all_meta['survey_answers']), 1)
+
+            # TODO FIXME HACK:  Unclear that the order surveys are returned is
+            #  guaranteed, so this particular part of the test is fragile.
+            #  Feel free to blow it away if it starts failing due to reordering
+            #  of returned surveys
+            # And the first survey should match
+            self.assertDictEqual(meta['survey_answers'][0],
+                                 all_meta['survey_answers'][0])
--- a/microsetta_private_api/admin/tests/test_admin.py
+++ b/microsetta_private_api/admin/tests/test_admin.py
@@ -128,11 +128,8 @@ class AdminTests(TestCase):
 
             BARCODE = '000004216'
 
-            try:
+            with self.assertRaises(NotFound):
                 admin_repo.get_survey_metadata("NOTABARCODE")
-                self.fail("Should not find NOTABARCODE")
-            except NotFound:
-                pass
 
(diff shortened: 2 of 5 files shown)


wasade:
@dhakim87 note, conflicts

microsetta_private_api/admin/admin_impl.py
```diff
@@ -17,6 +17,29 @@ def search_barcode(token_info, sample_barcode):
...
+
+    with Transaction() as t:
+        admin_repo = AdminRepo(t)
+        sample_pulldown = admin_repo.get_survey_metadata(
+            sample_barcode,
+            survey_template_id=survey_template_id)
+    return jsonify(sample_pulldown), 200
+
+
+def sample_pulldown_multiple_survey(token_info,
```
wasade, line 33: What's the difference between these functions?

microsetta_private_api/admin/tests/test_admin.py
```diff
@@ -121,3 +121,40 @@ def test_search_barcode(self):
...
             self.assertIsNone(diag['sample'])
             self.assertEqual(len(diag['barcode_info']), 0)
+
+    def test_get_survey(self):
+        with Transaction() as t:
+            admin_repo = AdminRepo(t)
+
+            BARCODE = '000004216'
+
+            try:
```
wasade, line 131:
```python
with self.assertRaises(NotFound):
    admin_repo.get_survey_metadata("NOTABARCODE")
```

microsetta_private_api/admin/tests/test_admin.py
```diff
@@ -121,3 +121,40 @@ def test_search_barcode(self):
...
+            self.assertEqual(len(meta['survey_answers']), 1)
+
+            all_meta = admin_repo.get_survey_metadata(BARCODE)
+
+            self.assertEqual(all_meta['sample_barcode'], BARCODE)
+            self.assertEqual(all_meta['host_subject_id'],
+                             all_meta['host_subject_id'])
+            # And there should be more than one survey answered
+            self.assertGreater(len(all_meta['survey_answers']), 1)
+
```
wasade, line 152: Is `sample_type` and the date/time of sample collection included?

microsetta_private_api/api/microsetta_private_api.yaml
```diff
@@ -835,6 +835,47 @@ paths:
...
+      tags:
+        - Admin
+      summary: Retrieve survey responses for a particular sample and survey
+      description: Retrieve survey responses for a particular sample and survey
+      parameters:
+        - $ref: '#/components/parameters/sample_barcode'
+        - $ref: '#/components/parameters/survey_template_id'
+      responses:
+        '200':
+          description: Qiime compatible sample metadata
```
wasade, line 850: QIIME

microsetta_private_api/api/microsetta_private_api.yaml
```diff
@@ -835,6 +835,47 @@ paths:
...
+      operationId: microsetta_private_api.admin.admin_impl.sample_pulldown_multiple_survey
+      tags:
+        - Admin
+      summary: Retrieve all survey responses for a particular sample
+      description: Retrieve all survey responses for a particular sample
+      parameters:
+        - $ref: '#/components/parameters/sample_barcode'
+      responses:
+        '200':
+          description: Qiime compatible sample metadata
```
wasade, line 870: QIIME

microsetta_private_api/repo/admin_repo.py
```diff
@@ -75,3 +87,76 @@ def retrieve_diagnostics_by_barcode(self, sample_barcode):
...
+            source_repo = SourceRepo(self._transaction)
+            source = source_repo.get_source(account_id, source_id)
+
+        if source is None:
+            raise RepoException("Barcode is not associated with a source")
+
+        # TODO: This is my best understanding of how the data must be
+        #  transformed to get the host_subject_id, needs verification that it
+        #  generates the expected values for preexisting samples.
+        prehash = account_id + source.source_data.name.lower()
```
wasade, line 111: To check, when we checked hashes with "Daniel McDonald" on Friday, was `.lower()` used?

microsetta_private_api/repo/admin_repo.py
```diff
@@ -75,3 +87,76 @@ def retrieve_diagnostics_by_barcode(self, sample_barcode):
...
+                template_id = survey_answers_repo.find_survey_template_id(
+                    answer_id)
+                if template_id == survey_template_id:
+                    answers.append(answer_id)
+
+            if len(answers) == 0:
+                raise NotFound("This barcode is not associated with any "
+                               "surveys matching this template id")
+            if len(answers) > 1:
+                #  I really hope this can't happen.  (x . x)
```
wasade, line 132: Same...

wasade:
Sorry, just remembered an additional item: can location information associated with the account be provided?

microsetta_private_api/admin/admin_impl.py
dhakim87: sample_pulldown_single_survey takes a survey_template_identifier and returns only survey responses matching that identifier.  (I believe that up to one survey response for a given questionnaire is expected per sample, but I don't know that that's enforced at the database layer.)  

sample_pulldown_multiple_survey returns all survey responses associated with the sample, so a single request might give back answers for the primary survey and a covid survey for example.  Again, I expect only a single response on each survey will be sample associated, but I'm not sure it's actually enforced by the schema.

microsetta_private_api/repo/admin_repo.py
dhakim87: Yes.  The .lower() was present in the snippet you sent me, so I believe it will be necessary to match up existing participant hashes.  There is a potential issue, especially for environmental samples, that participants will share the same name, either before or after the .lower(), causing the hash to combine sources.  I don't have a good way to fix this without changing the hashing policy and diverging from existing hashes.

microsetta_private_api/admin/tests/test_admin.py
dhakim87: I've gone ahead and added a sample object at the root.  This is another spot where knowing if this api needs to be machine parseable is important.  We locked down the field names in the to_api version of this object, but the internal field names and organization haven't been relied upon and I feel like they could still change.  It would be clear to a user that sample_type, sample_site, and site might mean more or less the same thing, but knowing which one will appear is tricky at the moment.  

I can't tell if I should be passed the to_api information that is visible to all users, or if the admin needs more of the internal state than is otherwise visible.  But the internal state is hard to parse - for example, it will be called _scan_date to grab date time information right now.

microsetta_private_api/repo/admin_repo.py
wasade: Okay great, thanks. Environmental samples do not have a host, so no subject ID.

microsetta_private_api/admin/tests/test_admin.py
wasade: I think I misspoke on the original comment, and the field name is `site_sampled`. But, I'm not sure I follow all of this. Does the output include "sample_date", "sample_time", "site_sampled" and "notes" from `ag.ag_kit_barcodes`? And what's the distinction here between machine parsable and not? 

It's less clear if information like `scan_date` are needed. And we do not currently expose project level information, although we may want to in the future.

dhakim87: What I'm sending down contains the same information but under different names.  The object I pass down currently matches the object defined in microsetta_private_api/model/sample.py  That's the one I consider to define the internal representation.  I could grab instead the exact schema definition from the sql, or call to_api on the model object to match the yaml.  But whichever of these we pick, the microsetta-admin project becomes coupled to.  So we need to decide whether microsetta-admin either understands the database schema, the yaml api, or the /model objects.  

Alternately, microsetta-admin understands none of these and all it can show is a generic tree view of a bunch of collapsible json that a user can click on to find things.

wasade: Thanks! let's do what's in `microsetta_private_api/model/sample.py`

dhakim87:
I've thrown in source and account information into the returned object, so you can pull the address field out of the account.

wasade:
Awesome, so this is all good?

PR #116, opened 2020-04-22
Email verification

AmandaBirmingham:
Per @wasade, any user of the private API must have validated their email (with the authentication provider, AuthRocket) before they are allowed to perform any actions on the API.

Note that since it has been requested that the minimalInterface inform users who do not have a verified email that they need to go do that, it was also necessary to differentiate between a user who just isn't authenticated and one who is but who still needs to verify their email.  I have set this up via a 403 forbidden http response code (Forbidden) for the authenticated-but-email-not-verified case, but if this sparks disagreement, let me know.

### DIFF
--- a/microsetta_private_api/api/implementation.py
+++ b/microsetta_private_api/api/implementation.py
@@ -618,8 +618,23 @@ def verify_authrocket(token):
     except InvalidTokenError as e:
         raise(Unauthorized(INVALID_TOKEN_MSG, e))
 
+    if JWT_ISS_CLAIM_KEY not in token_info or \
+            JWT_SUB_CLAIM_KEY not in token_info or \
+            JWT_EMAIL_CLAIM_KEY not in token_info:
+        # token is malformed--no soup for you
+        raise Unauthorized(INVALID_TOKEN_MSG)
 
-def validate_access(token_info, account_id):
+    # if the user's email is not yet verified, they are forbidden to
+    # access their account even regardless of whether they have
+    # authenticated with authrocket
+    if email_verification_key not in token_info or \
+            token_info[email_verification_key] is not True:
+        raise Forbidden("Email is not verified")
+
+    return token_info
+
+
+def _validate_account_access(token_info, account_id):
     with Transaction() as t:
         account_repo = AccountRepo(t)
         account = account_repo.get_account(account_id)
--- a/microsetta_private_api/api/implementation.py
+++ b/microsetta_private_api/api/implementation.py
@@ -51,6 +51,7 @@ AUTHROCKET_PUB_KEY = pkg_resources.read_text(
     "authrocket.pubkey")
 JWT_ISS_CLAIM_KEY = 'iss'
 JWT_SUB_CLAIM_KEY = 'sub'
+JWT_EMAIL_CLAIM_KEY = 'email'
 
 ACCT_NOT_FOUND_MSG = "Account not found"
 INVALID_TOKEN_MSG = "Invalid token"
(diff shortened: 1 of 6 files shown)


microsetta_private_api/api/implementation.py
```diff
@@ -591,29 +592,34 @@ def create_human_source_from_consent(account_id, body, token_info):
...
+                'email' not in token_info:
+            # token is malformed--no soup for you
+            raise Unauthorized(INVALID_TOKEN_MSG)
+
+        # if the user's email is not yet verified, they are forbidden to
+        # access their account even regardless of whether they have
+        # authenticated with authrocket
+        if email_verification_key not in token_info or \
+                token_info[email_verification_key] is not True:
+            raise Forbidden("Email is not verified")
```
wasade, line 615: Is it necessary for these lines to be within the try/except? It looks like only `InvalidTokenError` is caught, and I apologize if this is off but my assumption is that can only be thrown by `jwt.decode`

microsetta_private_api/api/tests/test_auth.py
```diff
@@ -130,10 +169,31 @@ def test_cant_do_anything(self):
...
-        resp = self.client.get('/api/accounts/%s/sources?language_tag=en-US' %
-                               ACCT_ID, headers={
-                                   "Authorization": MODIFIED_TOKEN
-                               })
-        check_response(resp, 401)
-        self.assertEqual(json.loads(resp.data)['detail'], "Invalid Token")
+    def _help_test_mock_decode_token(self, fake_token, expected_code,
+                                     expected_error_detail):
+        with patch("jwt.decode") as mock_d:
+            mock_d.side_effect = decode_fake_token
```
wasade, line 175: I'm rapidly growing very fond of `patch`

microsetta_private_api/api/implementation.py
```diff
@@ -591,29 +592,34 @@ def create_human_source_from_consent(account_id, body, token_info):
...
+        if JWT_ISS_CLAIM_KEY not in token_info or \
+                JWT_SUB_CLAIM_KEY not in token_info or \
+                'email' not in token_info:
+            # token is malformed--no soup for you
+            raise Unauthorized(INVALID_TOKEN_MSG)
+
+        # if the user's email is not yet verified, they are forbidden to
+        # access their account even regardless of whether they have
+        # authenticated with authrocket
+        if email_verification_key not in token_info or \
```
wasade, line 613: An alternative, but I'll defer to you as to if this is more or less readable:

```python
if token_info.get(email_verification_key, False) is not True:
    raise ...
```

At first I thought `is not True` should be `is False`, however the present approach testing explicitly against `True` is more defensive given Python's ducktyping.

dhakim87:
I didn't see anything to tell clients of the minimal interface to check their emails or resend the email verification with a redirect to authrocket.  Fine if that is going into a separate PR, but I don't want it to get forgotten.  This could lock everyone out of the api if there are any issues in the email verification workflow.

microsetta_private_api/api/implementation.py
AmandaBirmingham: This is less readable to me, but maybe that is just because I haven't yet made fast friends with `get` ... I will change it if either of you feel strongly about it but otherwise will leave.

wasade: No, let's leave as is. Present is more explicit. Thanks!

microsetta_private_api/api/implementation.py
AmandaBirmingham: The only constraint is that they need to come before the return, which was housed within the try.  I have moved it (and those extra checks) outside.

wasade: Thanks!

AmandaBirmingham:
@dhakim87 , I know you approved the previous version of this PR, but it has some non-trivial new additions to handle issue 113, so I would be very grateful if you could re-review ...

dhakim87:
Looks good, I like that we'll be able to tell the user if their email doesn't match now.

PR #117, opened 2020-04-22
Admin Barcode Search Cleanup

dhakim87:
Minor changes to the behavior in Admin Barcode Search.  Now pushes all the projects inside one barcode_info rather than duplicating whole records.  Now returns 404 when no information about a barcode can be found.


wasade:
Thanks!

PR #118, opened 2020-04-23
Prep for ui email verification and legacy acct support

AmandaBirmingham:
While implementing the client changes for requiring email verification and letting users claim legacy accounts, I ran across a couple of things that I wanted/needed to change.  First of all, although AuthRocket's documentation said they will send back a claim named 'email_verification', actually they send back one called 'email_verified' ... ok.  Second, in the interface code it is convenient to have the interface for finding a legacy account (by email) be of the same structure as the interface for finding an account by login info; therefore, I modified POST /accounts/legacies to return an array--empty if it found (and claimed) no legacy account, filled with a single account if it did find and claim one.  Just as for `find_accounts_for_login`, the *api* structure supports returning more than one account, but the underlying code in `implementation.py` is not actually capable of returning more than one account at a time.

### DIFF
--- a/microsetta_private_api/api/implementation.py
+++ b/microsetta_private_api/api/implementation.py
@@ -89,9 +89,9 @@ def claim_legacy_acct(token_info):
         t.commit()
 
         if acct is None:
-            return jsonify(code=404, message=ACCT_NOT_FOUND_MSG), 404
+            return jsonify([]), 200
 
-        return jsonify(acct.to_api()), 200
+        return jsonify([acct.to_api()]), 200
 
 
 def register_account(body, token_info):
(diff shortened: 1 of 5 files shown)


microsetta_private_api/api/implementation.py
```diff
@@ -89,9 +89,9 @@ def claim_legacy_acct(token_info):
         t.commit()
 
         if acct is None:
-            return jsonify(code=404, message=ACCT_NOT_FOUND_MSG), 404
+            return jsonify([]), 200
```
wasade, line 92: Just to make sure, this is supposed to switch from 404 -> 200?

AmandaBirmingham: @wasade Thanks for the check, and yes, it is intentional.  I agree it feels a bit odd, but basically the change in messaging is from "I looked for a particular account and failed to find it (404)" to "I successfully searched for accounts and here is a list of all the ones I found (200)", even when when the list is empty because I found no accounts.  This matches the idiom of `find_accounts_for_login`,  which also considers success to mean "I searched successfully", not necessarily "I found something" :)

wasade:
Thanks!!

PR #119, opened 2020-04-24
UI support for email verification, email updating, and claiming a legacy account

AmandaBirmingham:


microsetta_private_api/example/client_impl.py
```diff
@@ -64,17 +84,23 @@ def home():
...
-        acct_id = workflow_state.get("account_id", None)
-        show_wizard = False  # workflow_needs != ALL_DONE
+
+        if email_verified:
+            workflow_needs, workflow_state = determine_workflow_state()
+            if workflow_needs == NEEDS_REROUTE:
+                return workflow_state["reroute"]
+
+            acct_id = workflow_state.get("account_id", None)
+            show_wizard = False  # workflow_needs != ALL_DONE
```
wasade, line 97: I'm not sure I understand the comment?

microsetta_private_api/example/client_impl.py
```diff
@@ -88,36 +114,52 @@ def authrocket_callback(token):
...
+            return NEEDS_REROUTE, current_state
+        # if no legacy account found, need new account
+        if len(accts_output) == 0:
+            return NEEDS_ACCOUNT, current_state
 
     acct_id = accts_output[0]["account_id"]
     current_state['account_id'] = acct_id
 
+    # If we haven't yet checked for email mismatches and gotten user decision:
+    if not session.get(EMAIL_CHECK_KEY):
```
wasade, line 150: I _think_ `session.get` will act like a `dict.get` and return `None` by default, however I can't find documentation on that object from the Flask website at the moment... 

Assuming this is the case, would it be possible to either alter this condition to be `if session.get(EMAIL_CHECK_KEY) is not None` or `if not session.get(EMAIL_CHECK_KEY, False)`?

microsetta_private_api/example/client_impl.py
```diff
@@ -224,6 +268,45 @@ def post_workflow_create_account(body):
...
+    return render_template("update_email.jinja2")
+
+
+def post_workflow_update_email(body):
+    next_state, current_state = determine_workflow_state()
+    if next_state != NEEDS_EMAIL_CHECK:
+        return redirect(WORKFLOW_URL)
+
+    # if the customer wants to update their email:
+    update_email = int(body["do_update"])  # 0 or 1
```
wasade, line 285: To avoid a potential (granted unlikely) type issue, what about making this `update_email = body.get('do_update') == "1"`?

microsetta_private_api/templates/home.jinja2
```diff
@@ -41,8 +36,23 @@
                 </a>
                 {% endif %}
             </div>
+            {% else %}
+            <p>
+                It appears that your email has not yet been verified.  Please check your email account (and spam folder)
```
wasade, line 41: It appears that your email has not yet been verified.  Please check your email account (and spam folder) for a verification email from "The Microsetta Initiative" and follow its instructions.  This email will come from our authentication service AuthRocket, and use the email address "noreply@loginrocket.com".

microsetta_private_api/templates/update_email.jinja2
```diff
@@ -0,0 +1,60 @@
...
+    <script src="/static/vendor/js/jquery.validate.min.js"></script>
+    <link rel="stylesheet" href="/static/css/minimal_interface.css" />
+    <script>
+        function cancelEmailUpdate(){
+            $('#do_update').val("0");
+        }
+    </script>
+</head>
+<body>
+    <a href="http://microsetta.ucsd.edu" title="microsetta.ucsd.edu">
```
wasade, line 17: Can this be https?

microsetta_private_api/example/client_impl.py
```diff
@@ -41,13 +41,32 @@
...
+NEEDS_PRIMARY_SURVEY = "NeedsPrimarySurvey"
+ALL_DONE = "AllDone"
 
 
 # Client might not technically care who the user is, but if they do, they
 # get the token, validate it, and pull email out of it.
 def parse_jwt(token):
     decoded = jwt.decode(token, PUB_KEY, algorithms=['RS256'], verify=True)
-    return decoded["name"]
+    email_verified = decoded.get('email_verified') is True
```
wasade, line 68: `.get` returns `None` by default, so if this key is missing the comparison will be `None is True` which will work as expected. However, for type consistency, it may be nice to change this to `decoded.get('email_verified', False)` changing the default value used.

microsetta_private_api/example/client_impl.py
AmandaBirmingham: Have removed; was actually a bit of left-over code that was commented out and is no longer relevant.

microsetta_private_api/example/client_impl.py
AmandaBirmingham: Sensible ... as long as I'm doing it that way, I might as well just have the field be set to Yes or No as that is more human-readable than 1 and 0, so I have made that change.

microsetta_private_api/templates/home.jinja2
AmandaBirmingham: Text changed.

microsetta_private_api/example/client_impl.py
AmandaBirmingham: Done :)

microsetta_private_api/templates/update_email.jinja2
AmandaBirmingham: Done.  Note that because we built this interface the fast and stupid way, this had to be changed in every different template (because since we expected to throw this UI away very soon, we didn't invest time in setting up a base template and having all the UI pages feed into it).  

This is was not a problem for this change, which was a simple search-and-replace.  However, the more we expand the minimalInterface functionality, the more this will become a booby trap ...

microsetta_private_api/example/client_impl.py
AmandaBirmingham: Changed to `if not session.get(EMAIL_CHECK_KEY, False)` .

microsetta_private_api/example/client_impl.py
```diff
@@ -88,36 +114,52 @@ def authrocket_callback(token):
...
+        needs_reroute, email_match = ApiRequest.get(
+            "/accounts/%s/email_match" % acct_id)
+        if needs_reroute:
+            current_state["reroute"] = email_match
+            return NEEDS_REROUTE, current_state
+        # if they don't match AND the user hasn't already refused update
+        if not email_match["email_match"]:
+            return NEEDS_EMAIL_CHECK, current_state
+
+        session[EMAIL_CHECK_KEY] = True
```
wasade, line 161: Isn't this already assured from the conditional on line 150?

AmandaBirmingham: @wasade  Errr ... I think no?  I think one (or both?) of us are confused here :)

Line 150 says `if not session.get(EMAIL_CHECK_KEY, False):` .  Assuming I understand get() (not a given :), this basically says "if EMAIL_CHECK_KEY is False or is not in the session object at all, then ..." In this situation, we do a bunch of stuff and then, to keep us from doing it again next time, we set EMAIL_CHECK_KEY in the session object to True.  Thus, the next time we come through, the "if" clause discussed above will not be activated.

Do you see something I don't about how this logic works (or doesn't!) ?

microsetta_private_api/example/client.yaml
```diff
@@ -98,6 +98,36 @@ paths:
...
+      operationId: microsetta_private_api.example.client_impl.post_workflow_update_email
+      requestBody:
+        content:
+          application/x-www-form-urlencoded:
+            schema:
+              type: object
+              additionalProperties: true
+      responses:
+        '200':
+          description: Error report
```
dhakim87, line 123: Error report on 200?

microsetta_private_api/example/client_impl.py
```diff
@@ -88,36 +114,52 @@ def authrocket_callback(token):
 
 def logout():
     if TOKEN_KEY_NAME in session:
-        del session[TOKEN_KEY_NAME]
+        # delete these keys if they are here, otherwise ignore
+        session.pop(TOKEN_KEY_NAME, None)
+        session.pop(KIT_NAME_KEY, None)
+        session.pop(EMAIL_CHECK_KEY, None)
```
dhakim87, line 120: Good for now, but we probably need to think about what variables are going to be kept in the session and what operations must clear them.

microsetta_private_api/example/client_impl.py
wasade: ohhhhhh... my bad, I see that now :)

microsetta_private_api/example/client_impl.py
```diff
@@ -88,36 +114,52 @@ def authrocket_callback(token):
...
         current_state["reroute"] = accts_output
         return NEEDS_REROUTE, current_state
+
     if len(accts_output) == 0:
-        return NEEDS_ACCOUNT, current_state
+        # NB: Overwriting outputs from get call above
+        needs_reroute, accts_output = ApiRequest.post("/accounts/legacies")
+        if needs_reroute:
+            current_state["reroute"] = accts_output
+            return NEEDS_REROUTE, current_state
```
dhakim87, line 141: Comment: Can we wrap the needs_reroute somehow?  It appears a bunch of times in this function.  I could see it being an exception handler maybe, but only if all the ApiRequest's that need reroutes can be handled by the same path.

microsetta_private_api/example/client_impl.py
```diff
@@ -88,36 +114,52 @@ def authrocket_callback(token):
...
     needs_reroute, accts_output = ApiRequest.get("/accounts")
+    # if there's an error, reroute to error page
     if needs_reroute:
         current_state["reroute"] = accts_output
         return NEEDS_REROUTE, current_state
+
     if len(accts_output) == 0:
-        return NEEDS_ACCOUNT, current_state
+        # NB: Overwriting outputs from get call above
+        needs_reroute, accts_output = ApiRequest.post("/accounts/legacies")
```
dhakim87, line 138: Hmm... the fact determine workflow state causes a POST action with side effects is a code smell.  I'm not sure what can go wrong with this, but something about it feels off.

microsetta_private_api/example/client_impl.py
```diff
@@ -88,36 +114,52 @@ def authrocket_callback(token):
...
+            return NEEDS_REROUTE, current_state
+        # if no legacy account found, need new account
+        if len(accts_output) == 0:
+            return NEEDS_ACCOUNT, current_state
 
     acct_id = accts_output[0]["account_id"]
     current_state['account_id'] = acct_id
 
+    # If we haven't yet checked for email mismatches and gotten user decision:
+    if not session.get(EMAIL_CHECK_KEY, False):
```
dhakim87, line 150: Comment: I believe Flask session is currently stored in an in memory cache.  Session is wiped every time the server restarts/is upgraded to a new version.  It will (may?) also be wiped every time the user closes their browser window I believe.  All in all, I'm not sure this is a robust place to store user preferences like whether or not they have already taken an action and should not be shown an operation again.  

We can play around with making sessions permanent, or and or setting session lifetimes.  Maybe we can get the flask session to fit our needs.  For example, are we supposed to ask the user once on detecting a change and never ask them again if they say no?  Are we supposed to ask them once each time their email is changed?  Are we supposed to ask them once a week if we detect a difference?

microsetta_private_api/example/client_impl.py
```diff
@@ -224,6 +268,45 @@ def post_workflow_create_account(body):
...
+    return render_template("update_email.jinja2")
+
+
+def post_workflow_update_email(body):
+    next_state, current_state = determine_workflow_state()
+    if next_state != NEEDS_EMAIL_CHECK:
+        return redirect(WORKFLOW_URL)
+
+    # if the customer wants to update their email:
+    update_email = body["do_update"] == "Yes"
```
dhakim87, line 285: Why "Yes"?  Is that a default in html forms?  I thought checkboxes were "on" (or none, they don't send "off"!)  Fine if this is hardcoded I guess.

microsetta_private_api/templates/create_acct.jinja2
```diff
@@ -44,7 +44,7 @@
     </script>
 </head>
 <body>
-    <a href="http://microsetta.ucsd.edu" title="microsetta.ucsd.edu">
+    <a href="https://microsetta.ucsd.edu" title="microsetta.ucsd.edu">
```
dhakim87, line 47: We should move this to server_config.json

microsetta_private_api/templates/error.jinja2
```diff
@@ -7,7 +7,7 @@
     <link rel="stylesheet" href="/static/css/minimal_interface.css" />
 </head>
 <body>
-    <a href="http://microsetta.ucsd.edu" title="microsetta.ucsd.edu">
+    <a href="https://microsetta.ucsd.edu" title="microsetta.ucsd.edu">
```
dhakim87, line 10: We should move this to server_config.json

microsetta_private_api/templates/home.jinja2
```diff
@@ -7,7 +7,7 @@
     <link rel="stylesheet" href="/static/css/minimal_interface.css" />
 </head>
 <body>
-    <a href="http://microsetta.ucsd.edu" title="microsetta.ucsd.edu">
+    <a href="https://microsetta.ucsd.edu" title="microsetta.ucsd.edu">
```
dhakim87, line 10: We should move this to server_config.json

microsetta_private_api/templates/kit_sample_association.jinja2
```diff
@@ -37,7 +37,7 @@
     </script>
 </head>
 <body>
-    <a href="http://microsetta.ucsd.edu" title="microsetta.ucsd.edu">
+    <a href="https://microsetta.ucsd.edu" title="microsetta.ucsd.edu">
```
dhakim87, line 40: Hmm, maybe we should put it in a config file AND put it into a jinja2 template so its not copied everywhere.

microsetta_private_api/templates/update_email.jinja2
```diff
@@ -0,0 +1,60 @@
...
+    <br />
+
+    <div class="content">
+        <nav aria-label="breadcrumb">
+          <ol class="breadcrumb">
+            <li class="breadcrumb-item"><a href="/home">Home</a></li>
+          </ol>
+        </nav>
+        <div class="container">
+            It appears that the email we have for your account is not the same as the email
```
dhakim87, line 33: Might be nice to show them the email that we have and the email that they are logged in with so they don't get confused.

microsetta_private_api/example/client_impl.py
AmandaBirmingham: Yes, absolutely.   The more we expand the minimal interface (heh, minimal ...), the more we are going to accumulate the need for session variables (or else considerable architecting to remove the need for them) ...

microsetta_private_api/example/client.yaml
AmandaBirmingham: Yes.  The reason is that we have two layer of APIs here.  Errors *from the private api* are routed to the error.jinja2 template in the *client api*, which (successfully, i.e., with a 200) returns the html created from that template.  If you want, I could change this to say "private api error report"?

microsetta_private_api/example/client_impl.py
AmandaBirmingham: Yes, as mentioned on the slack channel: "If I say NO, nothing is changed in the db, and I can move around the interface (go look at my samples, go home, etc) without being asked this again--UNTIL I log out. _Every time I log out and log back in, I will again be asked if I want to update my email_."

I am not prepared to implement long-term (e.g. across sessions) storage of any kind of state like this in the minimal interface.  If it is a major requirement that the user not be asked this more than once, then I think we would need to go the route of storing it in the db somewhere.

microsetta_private_api/example/client_impl.py
AmandaBirmingham: It is not a checkbox :) It is a hidden field, and hidden fields are text.

microsetta_private_api/templates/create_acct.jinja2
AmandaBirmingham: We should also move it to a sitebase template (see my earlier comment to @wasade )

microsetta_private_api/templates/kit_sample_association.jinja2
AmandaBirmingham: Indeed :)

microsetta_private_api/templates/update_email.jinja2
AmandaBirmingham: I have the power to do this but I didn't want to reveal anything here without explicit directions to do so.  @wasade , do you want this added?

microsetta_private_api/example/client_impl.py
AmandaBirmingham: Good point.  I could rearchitect it to return a new state (NEEDS_LEGACY_CHECK) and then add a new endpoint in the client_api (/workflow_claim_legacy) and have workflow() say 
```
if next_state == NEEDS_REROUTE:
        return redirect("/workflow_claim_legacy")
```
and then add a new function in client_impl.py to handle that endpoint ... I guess I'd probably do it like `get_workflow_claim_kit_samples()`, where there is a get that automatically routes to a post if update is required.  

Thoughts?

microsetta_private_api/example/client_impl.py
AmandaBirmingham: A very good point.  I can definitely raise a custom exception (RerouteException?) when a reroute is needed ... I'm torn as to how to handle it, though.  I could wrap the body of every function in client_impl.py in a try/catch RerouteException block, which is still a lot of repetition ... or I could add a statement like `app.app.register_error_handler(RerouteException, handle_reroute)` in server.py ... but if I follow the current model, I would define `RerouteException` in microsetta-private-api.exceptions.py and `handle_reroute()` in server.py ... which means really interweaving private api and minimal interface -related methods.  I don't love either of these .. thoughts?  Better ideas?

PR #120, opened 2020-04-27
Allow an admin to create a new project

wasade:
As the subject says. This PR also scopes to denote whether is project is to be considered a microsetta project.

### DIFF
--- a/microsetta_private_api/admin/admin_impl.py
+++ b/microsetta_private_api/admin/admin_impl.py
@@ -87,6 +87,26 @@ def validate_admin_access(token_info):
             raise Unauthorized()
 
 
+def create_project(body, token_info):
+    validate_admin_access(token_info)
+
+    project_name = body['project_name']
+    if len(project_name) == 0:
+        return jsonify(code=422, message="No project name provided"), 422
+
+    with Transaction() as t:
+        admin_repo = AdminRepo(t)
+
+        try:
+            admin_repo.create_project(project_name)
+        except:  # noqa
+            return jsonify(code=422, message="Could not create project"), 422
+        else:
+            t.commit()
+
+    return {}, 201
+
+
 def create_kits(body, token_info):
     validate_admin_access(token_info)
 
--- a/microsetta_private_api/admin/admin_impl.py
+++ b/microsetta_private_api/admin/admin_impl.py
@@ -94,17 +94,12 @@ def create_project(body, token_info):
     is_microsetta = body['is_microsetta']
 
     if len(project_name) == 0:
-        return jsonify(code=422, message="No project name provided"), 422
+        return jsonify(code=400, message="No project name provided"), 400
 
     with Transaction() as t:
         admin_repo = AdminRepo(t)
-
-        try:
-            admin_repo.create_project(project_name, is_microsetta)
-        except:  # noqa
-            return jsonify(code=422, message="Could not create project"), 422
-        else:
-            t.commit()
+        admin_repo.create_project(project_name, is_microsetta)
+        t.commit()
 
     return {}, 201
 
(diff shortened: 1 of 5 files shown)


microsetta_private_api/admin/admin_impl.py
```diff
@@ -87,6 +87,28 @@ def validate_admin_access(token_info):
...
 
 
+def create_project(body, token_info):
+    validate_admin_access(token_info)
+
+    project_name = body['project_name']
+    is_microsetta = body['is_microsetta']
+
+    if len(project_name) == 0:
+        return jsonify(code=422, message="No project name provided"), 422
```
dhakim87, line 97: Might consider a 400 error code for this.  I believe that's what would be thrown if this were detected via a regex in the yaml.

microsetta_private_api/admin/admin_impl.py
```diff
@@ -87,6 +87,28 @@ def validate_admin_access(token_info):
...
+
+    with Transaction() as t:
+        admin_repo = AdminRepo(t)
+
+        try:
+            admin_repo.create_project(project_name, is_microsetta)
+        except:  # noqa
```
dhakim87, line 104: Yes qa!  What exception does it throw?  You can take the unique constraint violation code out of the other repo classes.

dhakim87: Also, wrapping psycopg2 errors in RepoException and re-throwing them in the repo layer will automatically trigger 422s to be returned, which is convenient when you need to specify an error message close to the source

microsetta_private_api/admin/admin_impl.py
wasade: Done

microsetta_private_api/admin/admin_impl.py
wasade: Ah okay so dont need to wrap this business in a try/except then right?

microsetta_private_api/db/patches/0056.sql
```diff
@@ -0,0 +1,10 @@
+ALTER TABLE barcodes.project ADD COLUMN is_microsetta VARCHAR;
```
dhakim87, line 1: Why VARCHAR with Yes/No rather than the Postgres BOOLEAN type?  Then we wouldn't need to think about capitalization, which has burned us in a previous "Yes/no" field.

microsetta_private_api/admin/admin_impl.py
dhakim87: If its specifically a RepoException, it will turn into a 422 from the registered handler in server.py.  If its a different type of exception underneath, it'll do whatever is registered, or 500.  That lets you 404 from the repo layer with NotFound or throw 422s with RepoException.  You still have to catch the psycopg2 unique violations explicitly, but I do that at the repo layer and reraise them as RepoExceptions usually.

microsetta_private_api/db/patches/0056.sql
wasade: Not an issue, happy to address

microsetta_private_api/db/patches/0056.sql
```diff
@@ -1,10 +1,10 @@
...
-UPDATE barcodes.project SET is_microsetta='No' WHERE is_microsetta IS NULL;
-UPDATE barcodes.project SET is_microsetta='Yes' WHERE project in ('American Gut Project',
-                                                              'British Gut Project',
-                                                              'Daklapak W1',
-                                                              'canadian gut',
-                                                              'TMI - Daklapack W1',
-                                                              'TMI - Daklapack W2');
+ALTER TABLE barcodes.project ADD COLUMN is_microsetta BOOLEAN;
+UPDATE barcodes.project SET is_microsetta='yes' WHERE is_microsetta IS NULL;
+UPDATE barcodes.project SET is_microsetta='no' WHERE project in ('American Gut Project',
```
dhakim87, line 3: yes and no were switched here.  Looks unintentional.

PR #125, opened 2020-04-28
(WIP) St dh port vioscreen

dhakim87:
A work in progress for including vioscreen surveys.  Needs testing with real credentials to see if urls generated can access the vioscreen surveys.


microsetta_private_api/api/implementation.py
```diff
@@ -363,6 +365,10 @@ def submit_answered_survey(account_id, source_id, language_tag, body,
                            token_info):
     _validate_account_access(token_info, account_id)
 
+    # TODO:  What template id number is assigned for vioscreen?
```
wasade, line 368: None right now, but we should consider changing that

microsetta_private_api/api/implementation.py
```diff
@@ -653,3 +659,23 @@ def _validate_account_access(token_info, account_id):
...
         return account
+
+
+def _submit_vioscreen_status(key):
+    # get information out of encrypted vioscreen url
+    info = vioscreen.decode_key('key')
+    vio_info = {}
+    for keyval in info.split("&"):
+        key, val = keyval.split("=")
+        vio_info[key] = val
```
wasade, line 670: Would it make sense to replace this with urllib? https://stackoverflow.com/a/21584580

microsetta_private_api/api/implementation.py
```diff
@@ -653,3 +659,23 @@ def _validate_account_access(token_info, account_id):
...
+
+    with Transaction() as t:
+        vio_repo = VioscreenRepo(t)
+
+        # Add the status to the survey
+        vio_repo.update_vioscreen_status(vio_info["username"],
+                                         int(vio_info["status"]))
+        t.commit()
+
+    # TODO: Any reason to respond with anything?
```
wasade, line 680: open question: should we extract the FFQ survey results now, or later?

microsetta_private_api/repo/vioscreen_repo.py
```diff
@@ -0,0 +1,41 @@
...
+from werkzeug.exceptions import NotFound
+
+
+# This was ported from the american_gut_project's ag_data_access.py.
+class VioscreenRepo(BaseRepo):
+    def __init__(self, transaction):
+        super().__init__(transaction)
+
+    def update_vioscreen_status(self, survey_id, status):
+        # TODO:  Who is checking permissions for this?
```
wasade, line 11: Is there a away to restrict this to the account associated to the survey ID?

microsetta_private_api/util/vioscreen.py
```diff
@@ -0,0 +1,104 @@
...
+    # TODO: Is this the right way to do localization here?
+    if language_tag == "en_us":
+        text_locale = american_gut.text_locale
+    elif language_tag == "en_gb":
+        text_locale = british_gut.text_locale
+    else:
+        raise BadRequest("Unknown Locale: " + language_tag)
+
+    """Return a formatted text block and URL for the external survey"""
+    tl = text_locale['human_survey_completed.html']
```
wasade, line 21: Would it be okay to drop this?

microsetta_private_api/util/vioscreen.py
```diff
@@ -0,0 +1,104 @@
...
+from base64 import b64decode, b64encode
+from werkzeug.exceptions import BadRequest
+from werkzeug.urls import url_encode
+
+from microsetta_private_api.config_manager import AMGUT_CONFIG, SERVER_CONFIG
+from microsetta_private_api.LEGACY.locale_data import american_gut, british_gut
+
+
+def wrap_survey_url(survey_id, language_tag):
+    # TODO: Is this the right way to do localization here?
```
wasade, line 12: I dont think this is needed

microsetta_private_api/util/vioscreen.py
```diff
@@ -0,0 +1,104 @@
...
+    length = 16 - (len(in_message) % 16)
+    return in_message + chr(length) * length
+
+
+def pkcs7_unpad_message(in_message, ):
+    return in_message[:-ord(in_message[-1])]
+
+
+def encrypt_key(survey_id, language_tag):
+    # TODO: Is this the right way to do localization here?
```
wasade, line 54: I dont think this is needed

microsetta_private_api/util/vioscreen.py
```diff
@@ -0,0 +1,104 @@
...
+
+    """Encode minimal required vioscreen information to AES key"""
+    firstname = "NOT"
+    lastname = "IDENTIFIED"
+    gender_id = 2
+    dob = '01011800'
+
+    regcode = SERVER_CONFIG["vioscreen_regcode"]
+
+    # TODO: Specifying a callback url that goes back to our api server is, once
```
wasade, line 70: Can we just put `microsetta-rest.ucsd.edu`?

microsetta_private_api/util/vioscreen.py
```diff
@@ -0,0 +1,104 @@
...
+    #  again, super weird.  How is this supposed to work with callbacks?
+
+    # TODO: What is supposed to go here for return url!?
```
wasade, line 73: This originally just went back to a handler on our site, which (the vspassthrough) which then ultimately redirected them to the existing site homepage

microsetta_private_api/api/microsetta_private_api.yaml
```diff
@@ -1250,7 +1251,14 @@ components:
       description: Client url to which the customized consent form should be posted.
       schema:
         type: string
-        example: "http://www.microsetta.org/workflow_create_human_source"
+        example: "https://www.microsetta.org/workflow_create_human_source"
```
wasade, line 1254: what is microsetta.org?

microsetta_private_api/example/client.yaml
```diff
@@ -225,6 +240,63 @@ paths:
...
+        '200':
+          description: Error report
+          content:
+            text/html:
+              schema:
+                type: string
+        '302':
+          description: Redirecting to necessary action
+
+# TODO:  I'm not actually sure if vioscreen does a GET or a POST when its done.
```
wasade, line 277: It's a GET

microsetta_private_api/example/client_impl.py
```diff
@@ -60,6 +60,10 @@
 NEEDS_PRIMARY_SURVEY = "NeedsPrimarySurvey"
 ALL_DONE = "AllDone"
 
+# TODO FIXME HACK:  This is bullcrap.  VIOSCREEN_ID is just hardcoded.
```
wasade, line 63: If this is such a major concern, then please consider an alternative such as storing it in the database with other template IDs

microsetta_private_api/example/client_impl.py
```diff
@@ -431,20 +435,138 @@ def post_workflow_fill_primary_survey():
...
+        return redirect(survey_output['survey_template_text']['url'])
+
+    # Handle local surveys
+    return render_template("survey.jinja2",
+                           survey_schema=survey_output[
+                               'survey_template_text'])
+
+
+def finish_vioscreen(account_id, source_id, key):
+    # TODO FIXME HACK:  This is insanity.  I need to see the vioscreen docs
```
wasade, line 537: What specific information is necessary?

microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -15,6 +15,7 @@
 
 class SurveyTemplateRepo(BaseRepo):
 
+    VIOSCREEN_ID = 10001
```
wasade, line 18: This looks to be defined in two places?

microsetta_private_api/templates/source.jinja2
```diff
@@ -28,18 +27,46 @@
...
                     </div>
                     <div class="col-sm">
                       {{sample.sample_datetime |e}}
                     </div>
                     <div class="col-sm">
                       {{sample.sample_site |e}}
                     </div>
+                    <div class="col-sm">
+                      {% if sample.ffq %}
+                        <a href="/accounts/{{acct_id}}/sources/{{source_id}}/survey_templates/10001">Food Frequency Questionnaire</a><span class="fa fa-check"/>
```
wasade, line 43: Given that the number is defined, should 10001 be replaced by a variable?

microsetta_private_api/util/vioscreen.py
```diff
@@ -0,0 +1,95 @@
...
+    return embedded_text % url
+
+
+def gen_survey_url(survey_id, language_tag, survey_redirect_url):
+    if not survey_redirect_url:
+        raise BadRequest("Food Frequency Questionnaire Requires "
+                         "survey_redirect_url")
+
+    regcode = SERVER_CONFIG["vioscreen_regcode"]
+    # TODO: If we have problems getting the ciphertext to be accepted by
```
wasade, line 33: We've seen this work so this is a nonissue

microsetta_private_api/templates/source.jinja2
```diff
@@ -40,9 +40,9 @@
                     </div>
                     <div class="col-sm">
                       {% if sample.ffq %}
-                        <a href="/accounts/{{acct_id}}/sources/{{source_id}}/survey_templates/10001">Food Frequency Questionnaire</a><span class="fa fa-check"/>
+                        <a href="/accounts/{{acct_id}}/sources/{{source_id}}/samples/{{sample.sample_id}}/survey_templates/10001">Food Frequency Questionnaire</a><span class="fa fa-check"/>
```
wasade, line 43: Looks like the 10001 is still there?

wasade:
Thanks @dhakim87! It looks like one of the failures is [real](https://travis-ci.com/github/biocore/microsetta-private-api/jobs/325438740#L1276). The other is a dependency issue on py38 -- is pycrypto actually needed? It doesn't look like it's used here but maybe I'm not seeing it?

microsetta_private_api/example/client.yaml
```diff
@@ -225,6 +240,81 @@ paths:
               schema:
                 type: string
 
+  '/accounts/{account_id}/sources/{source_id}/survey_templates/{survey_template_id}':
```
wasade, line 243: this looks like a redefinition of a route in `microsetta_private_api.yaml`?

microsetta_private_api/api/microsetta_private_api.yaml
dhakim87: It's an example url string?  🤷‍♂️ All I did was make it https

microsetta_private_api/example/client_impl.py
dhakim87: Nothing in example/ folder is allowed to reference any other part of the codebase by design, all information has to come down through the api.  Otherwise we start adding dependencies that make it harder to tear this component out later.  Major issue is that vioscreen requires so much special handling: it implies the model we pass down through the api describing each survey is insufficient.

microsetta_private_api/example/client_impl.py
dhakim87: As an external survey, Vioscreen ends with a redirect to a specified url.  Our api does not specify how to integrate that callback method in any way, leaving it to the client_impl developer to figure out what fields ("key" for vioscreen) are returned by Vioscreen and how they must be packaged up to send back to the api.  A potential alternative is that Vioscreen's final redirect points right back to the api server, which then redirects back to the client, that would save some headache for the front end developer, but requires either persistent state or a clever mechanism for wrapping redirects for the entire round trip.  

This means that every external survey we add will require modification of any front ends to our api - which is not acceptable when a third party is developing them.

microsetta_private_api/repo/survey_template_repo.py
dhakim87: client_impl is not theoretically allowed to reference survey_template_repo, in the same way that any third party developing against our api would not be able to reference our constants.  The fact we now rely on magic numbers 1, 6, and 10001 in client_impl suggests at a minimum we need to document each of these numbers in or adjacent to the yaml.

microsetta_private_api/templates/source.jinja2
dhakim87: Magic number now passed in.

microsetta_private_api/util/vioscreen.py
dhakim87: Sure, wrap_survey_url is now gone.

microsetta_private_api/example/client_impl.py
wasade: Let's create an issue to fix the API and model later, and revise the comment to refer to the issue.

microsetta_private_api/example/client_impl.py
wasade: Allowing for users to take vioscreen surveys is non-optional. If our API does not support this that well, then we need to fix our API. A workaround near term is tolerable as the opportunity costs are high.

microsetta_private_api/repo/survey_template_repo.py
wasade: Since `SurveyTemplateRepo` can access the database. One alternative here is:

- issue a patch to include this ID there
- have this object query that table when appropriate
- update `client_impl` to obtain the stable vioscreen ID from the object

wasade: Or if that's going to take long, then we leave as is right now in favor of getting something online

microsetta_private_api/api/implementation.py
dhakim87: Done

microsetta_private_api/util/vioscreen.py
dhakim87: This requires more thought.  It's possible we could have the get return back to the rest api rather than the client, but it would then be up to the rest api to redirect back to the client.  It's a slight change from what I had (Vioscreen redirects to client, which then sends a message to the api), but would probably be cleaner for developer implementing the api.  Unfortunately, it requires us to hold state of how to redirect to the client on the api side, or alternately to encode a redirect url to the client inside of our encoding of a redirect to the api endpoint.  It's probably possible, but something I'd prefer to design out more

wasade: Okay makes sense

microsetta_private_api/example/client_impl.py
```diff
@@ -456,32 +462,48 @@ def post_workflow_fill_survey(survey_template_id, body):
...
+    do_return, surveys_output, _ = ApiRequest.get(
         '/accounts/%s/sources/%s/survey_templates' % (account_id, source_id))
     if do_return:
         return surveys_output
 
     # Limit to only the primary and COVID19 survey as that is the primary
     # data focus for TMI right now.
+    per_sample = []
     per_source = []
     restrict_to = [1, 6]
```
wasade, line 498: I think this should include `VIOSCREEN_ID`?

microsetta_private_api/api/implementation.py
dhakim87: Reverted this change because urllib turned all values into lists for no reason.

PR #127, opened 2020-05-01
COVID-19 survey support

wasade:
IMPORTANT: this borrows blocks of code from @dhakim87's #125 pull request. Dan, please let me know if you would like me to clearly denote code that was copied or derived from your PR.

Specifically, I brought over the ability to display additional survey choices, with a few changes:

* given the limitations of the UI, and the need to focus on COVID, the additional "local" surveys are limited to primary / COVID
* the ability to re-take a survey has been removed per Slack conversation
* FFQ support is _not_ in this pull request, so #125 is still needed

Question triggering is not yet included here. I am investigating that, however it may need to come in as a separate PR. 

This is my first major modification to the UI, so I apologize in advance if any unusual decisions were made. 

microsetta-covid


microsetta_private_api/example/client_impl.py
```diff
@@ -441,10 +460,53 @@ def get_source(account_id, source_id):
     if do_return:
         return samples_output
 
+    # Retrieve all surveys available to the source
+    do_return, surveys_output = ApiRequest.get(
+        '/accounts/%s/sources/%s/survey_templates' % (account_id, source_id))
+    if do_return:
+        return surveys_output
+
+    # Filter surveys to per sample and per source surveys
```
AmandaBirmingham, line 469: I don't understand the filtering; under our current underlying db model, aren't all surveys per-sample?

wasade: This was a lingering comment that was relevant to the vioscreen PR this code was pulled from, will remove

microsetta_private_api/example/client_impl.py
```diff
@@ -407,6 +407,25 @@ def get_workflow_fill_primary_survey():
                                'survey_template_text'])
 
 
+def finish_survey(account_id, source_id, survey_template_id):
+    model = {}
```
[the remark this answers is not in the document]
wasade: Added, and no observed issues -- was able to take the survey, and get answers out of the database for the source

microsetta_private_api/example/client_impl.py
```diff
@@ -445,10 +479,39 @@ def get_source(account_id, source_id):
...
+    # Retrieve all surveys available to the source
+    do_return, surveys_output = ApiRequest.get(
+        '/accounts/%s/sources/%s/survey_templates' % (account_id, source_id))
+    if do_return:
+        return surveys_output
+
+    # Limit to only the primary and COVID19 survey as that is the primary
+    # data focus for TMI right now.
+    per_source = []
+    restrict_to = ['Primary', 'COVID19 Questionnaire']
```
AmandaBirmingham, line 491: Everywhere else in this module we make per-survey decisions by checking the `survey_template_id`.  We should have that value in the survey object being worked with here, and the titles aren't being used elsewhere in this function, so I suggest we use survey_template_ids here too (`restrict_to = [1, 6]`, etc) for later maintainability.

microsetta_private_api/example/client_impl.py
```diff
@@ -445,10 +479,39 @@ def get_source(account_id, source_id):
...
+        if survey['survey_template_title'] in restrict_to:
+            per_source.append(survey)
+
+    # Identify answered surveys for the source
+    do_return, survey_answers = ApiRequest.get(
+        '/accounts/%s/sources/%s/surveys' % (account_id, source_id))
+    if do_return:
+        return survey_answers
+
+    # TODO: Would be nice to know when the user took the survey instead of a
```
AmandaBirmingham, line 502: @wasade , I see this PR includes a db patch to add a timestamp column in the db to record this info.  However, we aren't trying to actually retrieve or use that info yet, right?

microsetta_private_api/templates/source.jinja2
```diff
@@ -4,13 +4,13 @@
     <meta charset="utf-8"/>
     <title>Microsetta Account Samples</title>
     <link rel="stylesheet" href="/static/vendor/bootstrap-4.4.1-dist/css/bootstrap.min.css" />
-
+    <link href="https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css" rel="stylesheet"/>
```
AmandaBirmingham, line 7: For all the other static resources, we are serving them locally rather than using a cdn.  Should we do that here, too?

microsetta_private_api/templates/source.jinja2
```diff
@@ -4,13 +4,13 @@
...
     <link rel="stylesheet" href="/static/css/minimal_interface.css" />
 </head>
 <body>
     <a href="https://microsetta.ucsd.edu" title="microsetta.ucsd.edu">
         &nbsp;&nbsp;
-        <img src="/static/img/logo-co.png" class="resize">
+        <img src="/static/img/logo-co.png" class="resize"/>
```
AmandaBirmingham, line 13: Ugh ... good catch, but because we didn't take time to make a sitebase template and instead copy-pasted the chrome text, this also needs to be fixed on 8 other pages (error.jinja2, create_account.jinja2, home.jinja2, etc etc)

microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -45,6 +45,12 @@ class SurveyTemplateRepo(BaseRepo):
             "Personal Microbiome Information",
             "1.0",
             "local"
+        ),
+        6: SurveyTemplateLinkInfo(
+            6,
+            "COVID19 Questionnaire",
```
AmandaBirmingham, line 51: Is there a reason for titling it "COVID19 etc" rather than the more standard "COVID-19 etc"?

microsetta_private_api/example/client_impl.py
```diff
@@ -191,7 +192,13 @@ def determine_workflow_state():
...
         return NEEDS_PRIMARY_SURVEY, current_state
 
-    # ???COVID Survey??? -> covid_survey.html
+    has_covid = False
+    for survey in surveys_output:
+        if survey['survey_template_id'] == 6:
+            has_covid = True
+            current_state["answered_covid_survey_id"] = survey["survey_id"]
+    if not has_covid:
+        return NEEDS_COVID_SURVEY, current_state
```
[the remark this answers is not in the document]
wasade: A similar pattern was put in

microsetta_private_api/example/client_impl.py
```diff
@@ -389,20 +399,28 @@ def post_workflow_claim_kit_samples(body):
             if do_return:
                 return sample_output
 
+            do_return, sample_survey_output = ApiRequest.post(
```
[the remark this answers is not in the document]
wasade: Comment updated

microsetta_private_api/example/client_impl.py
```diff
@@ -389,20 +399,28 @@ def post_workflow_claim_kit_samples(body):
...
+                '/accounts/{0}/sources/{1}/samples/{2}/surveys'.format(
+                    acct_id, source_id, curr_sample_id
+                ), json={"survey_id": answered_covid_survey_id}
+            )
+
```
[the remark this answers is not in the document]
wasade: Yes and I think that makes sense right now

microsetta_private_api/example/client_impl.py
wasade: Not yet, too much complexity for this deployment

microsetta_private_api/templates/source.jinja2
wasade: probably. I don't know what this is used for, I brought it over when bringing changes from #125

microsetta_private_api/templates/source.jinja2
wasade: Fixing

microsetta_private_api/example/client_impl.py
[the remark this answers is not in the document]
dhakim87: I'm confused.  Where do we filter for old/new samples?  Does sample_locked prevent association of new surveys or something?  My read is that a legacy account user that takes the covid19 survey will associate that survey with all their old samples with this code.  Is that incorrect?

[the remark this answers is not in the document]
wasade: Unless i'm mistaken, the minimal interface is designed to only associate to unassigned samples, correct? If so, an "old" sample is unassigned so this behavior would be okay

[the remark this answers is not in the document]
AmandaBirmingham: @dhakim87  there is no filtering of old/new samples--but unless I am greatly mistaken we still won't be picking up ANY old samples :). The trick here is that the association is being done in the workflow_claim_kit_samples functionality, which is ONLY triggered if the human source on the account does NOT have any samples associated with it

So, imagine a legacy user logs in.  This user is forced to take the covid survey, but since they already have a human source and have samples associated with that human source, they are never routed to /workflow_claim_kit_samples and thus their covid survey is not associated with any samples.  

@wasade Note that as I think through this above, I realize that if a legacy user comes back to provide sample type/sample date/time, etc for samples they received in a past kit but are only now using, those old samples will NOT be associated with the covid survey (or the primary survey), because of the logic here:

https://github.com/wasade/microsetta-private-api/blob/3732b014672277c69fa9ef5cbb22b9905614bd88/microsetta_private_api/example/client_impl.py#L205-L212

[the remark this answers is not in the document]
wasade: Thanks. Let's leave as is. "old" kits are not suitable for the covid work anyway

PR #129, opened 2020-05-04
ajax survey post

AmandaBirmingham:
To accommodate vue-form-generator library limitations, change survey.jinja behavior to post form data via an ajax call rather than via a traditional form post.  Modify back-end to handle multiple surveys through the same jinja page by distinguishing survey template id.

### DIFF
--- a/microsetta_private_api/templates/survey.jinja2
+++ b/microsetta_private_api/templates/survey.jinja2
@@ -11,6 +11,7 @@
     <link rel="stylesheet" href="/static/css/minimal_interface.css" />
     <script>
         var result_txt = "";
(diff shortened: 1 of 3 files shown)


microsetta_private_api/templates/survey.jinja2
```diff
@@ -28,7 +60,7 @@
         <div class="container" id="app">
             <div class="panel panel-default">
                 <div class="panel-body">
-                    <form method="post" id="survey_form">
+                    <form id="survey_form" onsubmit="return postAccount(this);">
```
wasade, line 63: where is `postAccount` defined?

PR #135, opened 2020-05-05
Vioscreen

wasade:
* Remove pycrypto
* add pycryptodome (updating for str -> bytes)
* disallow repeating a FFQ for the same sample
* disallow retaking a primary or COVID survey


microsetta_private_api/templates/source.jinja2
```diff
@@ -25,6 +25,9 @@
         </nav>
         <div class="container">
             Samples:
+            {% if samples|length > 1 %}
+            <p><i>If collecting all samples on the same day, please only complete a single Food Frequency Questionnaire.</i></p>
```
dhakim87, line 29: Doesn't seem like we have any way to associate after the fact, are we just assuming we will link implicitly?

wasade: Yes. Getting a UI in place for this will introduce more delay...

microsetta_private_api/config_manager.py
```diff
@@ -30,3 +30,5 @@ def __init__(self):
 with pkg_resources.open_text('microsetta_private_api', "server_config.json") \
         as fp:
     SERVER_CONFIG = json.load(fp)
+    SERVER_CONFIG['vioscreen_cryptokey'] = \
+        SERVER_CONFIG['vioscreen_cryptokey'].encode('ascii')
```
dhakim87, line 34: Are you sure it should be 'ascii' rather than 'utf-8' or 'utf-16'?  As far as the crypto is concerned, we need a byte array, and my understanding is that ascii only has 7 bits of each byte.  This only matters if one of the vioscreen passwords has a non ascii character in it (like é or something), but if that ever happens, all our crypto probably explodes.

wasade: pycryptodome bombs on `str` and was only accepting `bytes`

ascii is 8-bit and this is the python way to go from utf-8 to `bytes`... and believe me, the transition from py2 -> py3 was super painful. 

```python
>>> x = 'foo'
>>> type(x)
<class 'str'>
>>> y = x.encode('ascii')
>>> type(y)
<class 'bytes'>
```

microsetta_private_api/util/vioscreen.py
```diff
@@ -30,7 +30,7 @@ def gen_survey_url(language_tag, survey_redirect_url):
 def pkcs7_pad_message(in_message):
     # http://stackoverflow.com/questions/14179784/python-encrypting-with-pycrypto-aes  # noqa
     length = 16 - (len(in_message) % 16)
-    return in_message + chr(length) * length
+    return (in_message + chr(length) * length).encode('ascii')
```
dhakim87, line 33: Again, why ascii over utf-8?

wasade: pycryptodome was bombing on `str`

wasade:
Thanks!

PR #137, opened 2020-05-05
Kiss fix for no survey answers

AmandaBirmingham:
Record at least ONE answer to each survey (or, failing even that, error in a way that explicitly describes the problem).  Also, show logout link on error page.

### DIFF
--- a/microsetta_private_api/repo/survey_answers_repo.py
+++ b/microsetta_private_api/repo/survey_answers_repo.py
@@ -178,13 +178,25 @@ class SurveyAnswersRepo(BaseRepo):
             for survey_template_group in survey_template.groups:
                 for survey_question in survey_template_group.questions:
                     survey_question_id = survey_question.id
+                    q_type = survey_question.response_type
+
+                    # TODO FIXME HACK: Modify DB to make this unnecessary!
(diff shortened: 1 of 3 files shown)


microsetta_private_api/repo/survey_answers_repo.py
```diff
@@ -178,13 +178,25 @@ def submit_answered_survey(self, ag_login_id, source_id,
...
+                    # (even if the user answered nothing)
+                    #  or we can't properly track the survey template id later.
+                    # Therefore, if the user answered NOTHING, store an empty
+                    # string for the first string or text question in the
+                    # survey, just so something is recorded.
+                    if len(survey_model) == 0 and \
+                            (q_type == "STRING" or q_type == "TEXT"):
+                        survey_model[str(survey_question_id)] = ""
+
                     if str(survey_question_id) not in survey_model:
```
wasade, line 194: Should this check come first?

dhakim87: No, if the survey model is empty, putting the one on 194 first would cause us to skip all questions, leaving us with an empty survey again

PR #138, opened 2020-05-05
Question Triggers

dhakim87:
Survey questions can now be triggered by responses to other survey questions.

### DIFF
--- a/microsetta_private_api/templates/survey.jinja2
+++ b/microsetta_private_api/templates/survey.jinja2
@@ -54,13 +54,13 @@
 
         function set_triggers(model, field)
         {
-            if ('triggered_by' in field && field.triggered_by != null && field.triggered_by.length > 0)
+            if ('triggered_by' in field && field.triggered_by !== null && field.triggered_by.length > 0)
             {
                 field.visible = function()
                 {
                   visible = false
                   for (var entry of field.triggered_by)
-                      visible |= (model[entry['q_id']] == entry['response'])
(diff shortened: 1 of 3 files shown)


microsetta_private_api/templates/survey.jinja2
```diff
@@ -51,6 +51,20 @@
             // always return false, preventing a traditional post of the form
             return false;
         }
+
+        function set_triggers(model, field)
+        {
+            if ('triggered_by' in field && field.triggered_by != null && field.triggered_by.length > 0)
```
wasade, line 57: This may be naive, but does it matter here if this is `!=` or `!==`?

wasade: ..same for the other javascript comparisons

AmandaBirmingham: "JavaScript has both strict and type–converting comparisons. A strict comparison (e.g., ===) is only true if the operands are of the same type and the contents match. The more commonly-used abstract comparison (e.g. ==) converts the operands to the same type before making the comparison."

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comparison_Operators

AmandaBirmingham: The javascript linter always recommends strict comparisons by default ...

dhakim87: Updated

PR #143, opened 2020-05-08
FAQ

AmandaBirmingham:
Creates an FAQ page using the content provided by Dom/Kat.

Since this was a new page, I decided to take the time to set up the sitebase.jinja more usefully and then base the new page off of it.  HOWEVER, I didn't want to take the risk of switching all the EXISTING pages over to use the sitebase.jinja, so there is still a lot of copy-pasted text in those pages--including the new text creating a link to the FAQ page. Let me know if this partial approach concerns you.

### DIFF
--- a/microsetta_private_api/templates/sitebase.jinja2
+++ b/microsetta_private_api/templates/sitebase.jinja2
@@ -2,10 +2,12 @@
 <html lang="en">
 <head>
     <meta charset="utf-8"/>
-    <title>Microsetta</title>
+    <title>Microsetta {{page_title}}</title>
+
 {%block head%}
 {%endblock%}
 
+    <link rel="stylesheet" href="/static/vendor/bootstrap-4.4.1-dist/css/bootstrap.min.css" />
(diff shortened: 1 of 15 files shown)


microsetta_private_api/templates/sitebase.jinja2
```diff
@@ -19,13 +21,32 @@
     <br />
 
     <div class="content">
+
+    {% if show_breadcrumbs %}
+        <nav aria-label="breadcrumb">
+          <ol class="breadcrumb">
+            <li class="breadcrumb-item"><a href="/home">Home</a></li>
```
dhakim87, line 28: Having just a single fixed entry here is temporary?  Presumably everything that uses the sitebase has to pass in its own list of breadcrumbs yeah?  That way we can do Home - FAQ and Home - Accounts - Sources - Samples or whatever.

AmandaBirmingham: What an excellent, excellent point that I had forgotten about.  I am making an issue for this and leaving it as is for now for just the FAQ and consent pages.

PR #145, opened 2020-05-12
Update sample status for existing samples

wasade:
@dhakim87 this should resolve biocore/microsetta-admin#22, able to review?


wasade:
Good catch. And I'll see if I can revise the query to do some of the other statuses. It doesn't need to address all samples, but resolving many would be pleasant

PR #146, opened 2020-05-13
Tweaks and additions

wasade:
The last IRB amendment brought with it two new questions, and addressing some typos. In addition, there were some suggested changes to the COVID survey. All of that is embodied here.

### DIFF
--- a/microsetta_private_api/api/tests/test_api.py
+++ b/microsetta_private_api/api/tests/test_api.py
@@ -1025,9 +1025,6 @@ class SurveyTests(ApiTests):
             "survey_template_type": "local",
             "survey_id": real_id_from_loc,
         }
-        print(json.dumps(expected_output, indent=2))
-        print("*****************")
-        print(json.dumps(get_resp_obj, indent=2))
         self.assertEqual(expected_output, get_resp_obj)
 
         # check dict keys
--- a/microsetta_private_api/api/tests/test_api.py
+++ b/microsetta_private_api/api/tests/test_api.py
@@ -1010,6 +1010,11 @@ class SurveyTests(ApiTests):
 
         # check all elements of object in body are correct
         expected_model = copy.deepcopy(input_model)
+
+        # the output order from the multiple choice question (30) from the
+        # database is not stable
+        observed_model = get_resp_obj.pop('survey_text')
+
         # TODO: determine whether the current behavior that returns
         #  fields input as numbers as strings is actually correct or not
         expected_model['108'] = "68"
@@ -1019,9 +1024,23 @@ class SurveyTests(ApiTests):
             "survey_template_version": "1.0",
             "survey_template_type": "local",
             "survey_id": real_id_from_loc,
-            "survey_text": expected_model
         }
+        print(json.dumps(expected_output, indent=2))
+        print("*****************")
+        print(json.dumps(get_resp_obj, indent=2))
         self.assertEqual(expected_output, get_resp_obj)
+
+        # check dict keys
+        self.assertEqual(set(observed_model), set(expected_model))
+        for k in observed_model:
+            obs = observed_model[k]
+            exp = expected_model[k]
+
+            if isinstance(obs, list):
+                obs = sorted(obs)
+                exp = sorted(exp)
+
+            self.assertEqual(obs, exp)
 
     def test_survey_create_success_empty(self):
         """Successfully create a new answered survey without any answers"""
--- a/microsetta_private_api/db/patches/0061.sql
+++ b/microsetta_private_api/db/patches/0061.sql
@@ -128,7 +128,7 @@ SELECT replace_response(235, 'Much', 'Quite interfering');
 SELECT replace_response(235, 'Very Much Interfering', 'Very interfering');
 
 -- now lets get rid of our helper function
-DROP FUNCTION replace_response;
+DROP FUNCTION replace_response(qid INTEGER, old_resp VARCHAR, new_resp VARCHAR);
 
 -- finally, create two new primary questions and one covid question
 INSERT INTO ag.survey_question (survey_question_id, american, british,  question_shortname, retired) 
--- a/microsetta_private_api/db/patches/0061.sql
+++ b/microsetta_private_api/db/patches/0061.sql
@@ -1,0 +1,202 @@
+-- add in a few approved questions to the primary survey, and one to covid
+-- in addition, apply some tweaks / cleanup to the covid survey
+
+-- Update the name of the survey. This was previously called "Primary Information"
+-- But it's what gets presented to the user. We don't really have a concept of the
+-- name of a survey, but we potentially should add that.
+UPDATE ag.survey_group 
+    SET american='Microsetta Initiative/American Gut Project' 
+    WHERE group_order = -1;
+
+UPDATE ag.survey_group 
+    SET british='Microsetta Initiative/American Gut Project' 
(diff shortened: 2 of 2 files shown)


wasade:
So very confused with this build failure. Working on it

microsetta_private_api/db/patches/0061.sql
```diff
@@ -0,0 +1,202 @@
...
+        -- ...and insert our response
+        INSERT INTO ag.survey_question_response (survey_question_id, response, display_index)
+            VALUES (qid, 'One or two days', 2);
+    END LOOP;
+END $do$;
+
+-- typos and response replacements
+
+-- create a means to replace a response as this pattern will be used a few times
+-- note: postgres9 does not have procedures...
```
dhakim87, line 68: At what point do we intend to upgrade postgres to latest?  I don't know why we should continue with 9.5

microsetta_private_api/db/patches/0061.sql
```diff
@@ -0,0 +1,202 @@
...
+        VALUES (qid, new_resp, 999);
+
+    UPDATE ag.survey_answers
+        SET response=new_resp
+        WHERE survey_question_id=qid
+            AND response=old_resp;
+
+    DELETE FROM ag.survey_question_response
+        WHERE survey_question_id=qid
+            AND response=old_resp;
```
dhakim87, line 93: If you move this above the INSERT INTO ag.survey_question_response, can you get rid of the 999 and the UPDATE that overrides the 999?

microsetta_private_api/db/patches/0061.sql
dhakim87: If the syntax here is painful/obsolete, we could consider writing this functionality in python and psycopg2 inside of migration_support.py

microsetta_private_api/db/patches/0061.sql
```diff
@@ -0,0 +1,202 @@
...
+SELECT replace_response(234, 'A Little Somewhat', 'Somewhat worried');
+SELECT replace_response(234, 'Much', 'Quite worried');
+SELECT replace_response(234, 'Very Much Worried', 'Very worried');
+
+SELECT replace_response(235, 'A Little Somewhat', 'Somewhat interfering');
+SELECT replace_response(235, 'Much', 'Quite interfering');
+SELECT replace_response(235, 'Very Much Interfering', 'Very interfering');
+
+-- now lets get rid of our helper function
+DROP FUNCTION replace_response(qid INTEGER, old_resp VARCHAR, new_resp VARCHAR);
```
dhakim87, line 131: Is this the first time we've needed such a function?  Are there pros/cons to keeping/deleting it?

wasade:
Thanks @dhakim87! I'm working on the test failure right now. It's particularly snarky: some database state is changing during the test suite. The failure only occurs on a fresh build of the database, and does not occur on a subsequent run...

microsetta_private_api/db/patches/0061.sql
wasade: We should migrate, and I'll raise that with Jeff. It's not something that can happen on a relatively short timeframe though.

This isn't a migration as it's in part addressing a recent IRB amendment. It's not obvious if this would have been easier in the land of python.

microsetta_private_api/db/patches/0061.sql
wasade: Would violate FK relationships unfortunately

microsetta_private_api/db/patches/0061.sql
wasade: First time. We can bring it back should we need it in the future, but I really hope we do not ever need it again

wasade:
And @dhakim87 thank you for the detailed review. I know this was *a lot* to grok. It would be valuable to examine how we can improve the database structure. Even just moving away from tuple primary keys to integer or UUIDs would greatly help. 

What I really like about the structure is we can be language agnostic, and it enforces referential integrity. However, as you note, having this outside of the database would greatly facilitate VC, maintenance and review, and it can still be setup in a language agnostic manner. And actually, we could set this up as a data dictionary which would be very nice for ease of distribution. 

All that said, I don't think we want to under take this within this PR :) It will be a pretty big change with careful migration of the existing survey data.

microsetta_private_api/api/tests/test_api.py
```diff
@@ -1019,10 +1024,24 @@ def test_survey_create_success(self):
             "survey_template_version": "1.0",
             "survey_template_type": "local",
             "survey_id": real_id_from_loc,
-            "survey_text": expected_model
         }
+        print(json.dumps(expected_output, indent=2))
```
wasade, line 1028: ...ugh will remove

wasade:
@dhakim87 Just did one more manual check on the patch against the individual survey question IDs and the responses and things seem good

microsetta_private_api/db/patches/0061.sql
AmandaBirmingham: @wasade I am amazed by your optimism ... ;)

microsetta_private_api/db/patches/0061.sql
```diff
@@ -0,0 +1,202 @@
+-- add in a few approved questions to the primary survey, and one to covid
+-- in addition, apply some tweaks / cleanup to the covid survey
+
+-- Update the name of the survey. This was previously called "Primary Information"
+-- But it's what gets presented to the user. We don't really have a concept of the
+-- name of a survey, but we potentially should add that.
```
AmandaBirmingham, line 6: +1 for adding a survey title to the db, since that concept is already built into the private api as `survey_template_title` ...

microsetta_private_api/db/patches/0061.sql
```diff
@@ -0,0 +1,202 @@
...
+    VALUES (236, 'Occasionally (1-2 times/week)', 3);
+INSERT INTO ag.survey_question_response (survey_question_id, response, display_index) 
+    VALUES (236, 'Regularly (3-5 times/week)', 4);
+INSERT INTO ag.survey_question_response (survey_question_id, response, display_index) 
+    VALUES (236, 'Daily', 5);
+
+INSERT INTO ag.survey_question (survey_question_id, american, british,  question_shortname, retired) 
+    VALUES (237, 
+           'In an average week, how often do you eat various plant sources of protein including tofu, tempeh, edamame, lentils, chickpeas, peanuts, almonds, walnuts, or quinoa?',
+           'In an average week, how often do you eat various plant sources of protein including tofu, tempeh, edamame, lentils, chickpeas, peanuts, almonds, walnuts, or quinoa?',
```
AmandaBirmingham, line 159: Ya know what would be nice some fine day?  Some concept in the code like a "default language", so that if the value in the db for a given field was null, it would fall back and use the value from the default language, which would presumably be American English.  I get distressed by so much exact duplication of user-facing text ... seems guaranteed to cause pain eventually ...

wasade: It definitely already causes pain :/ I think it's safe to say that we really should revisit the survey structure.

microsetta_private_api/db/patches/0061.sql
wasade: I could add this in right now, however the argument against it is that it sounds like there is desire to revisit the survey structures anyway, in which case it may make sense to defer this addition until then

AmandaBirmingham: Understood

PR #147, opened 2020-05-15
Allow participants to add samples from another kit

wasade:
Allows a user to associate samples from an additional kit. What's not shown in the video is that I also forcibly logged out by killing the server, and the samples from the additional kit were still present.

add_samples

### DIFF
--- a/microsetta_private_api/example/client_impl.py
+++ b/microsetta_private_api/example/client_impl.py
@@ -373,70 +373,61 @@ def get_workflow_claim_kit_samples():
         return render_template("kit_sample_association.jinja2")
 
 
-def claim_kit_samples(current_state, kit_name):
-    acct_id = current_state["account_id"]
-    source_id = current_state["human_source_id"]
-    answered_survey_id = current_state["answered_primary_survey_id"]
-    answered_covid_survey_id = current_state["answered_covid_survey_id"]
+def claim_kit_samples(expected_state, body):
+    next_state, current_state = determine_workflow_state()
+    if next_state == expected_state:
+        acct_id = current_state["account_id"]
+        source_id = current_state["human_source_id"]
+        answered_survey_id = current_state["answered_primary_survey_id"]
+        answered_covid_survey_id = current_state["answered_covid_survey_id"]
 
-    # get all the unassociated samples in the provided kit
-    do_return, sample_output, _ = ApiRequest.get(
-        '/kits', params={KIT_NAME_KEY: kit_name})
-    if do_return:
-        return sample_output
-
-    # for each sample, associate it to the human source
-    # and ALSO to the (single) primary and COVID survey for this human
-    # source
-    for curr_sample_obj in sample_output:
-        curr_sample_id = curr_sample_obj["sample_id"]
-        do_return, sample_output, _ = ApiRequest.post(
-            '/accounts/{0}/sources/{1}/samples'.format(acct_id, source_id),
-            json={"sample_id": curr_sample_id}
-        )
-
+        # get all the unassociated samples in the provided kit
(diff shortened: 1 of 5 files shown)


microsetta_private_api/templates/source.jinja2
```diff
@@ -6,6 +6,36 @@
...
+                    },
+                    // Make sure the form is submitted to the destination defined
+                    // in the "action" attribute of the form when valid
+                    submitHandler: function (form) {
+                        form.submit();
+                    }
+                }
+            });
+        });
+    </script>
```
AmandaBirmingham, line 38: I feel strongly that we need to break this out into a single new javascript file that is shared between the pages that use this code rather than copying it.  @wasade I can do this refactor if you don't feel confident on it.

microsetta_private_api/example/client_impl.py
```diff
@@ -373,51 +373,68 @@ def get_workflow_claim_kit_samples():
...
+
+    return redirect(WORKFLOW_URL)
+
+
+def post_workflow_claim_kit_samples(body):
+    next_state, current_state = determine_workflow_state()
+
+    if next_state == NEEDS_SAMPLE:
+        kit_name = body[KIT_NAME_KEY]
+        return claim_kit_samples(current_state, kit_name)
```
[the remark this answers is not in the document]
wasade: :+1:

microsetta_private_api/templates/source.jinja2
wasade: Very good call. I'll take a first pass -- I think I know what needs to be done

wasade:
@AmandaBirmingham I think these changes hit the refactors, able to re-review when you have a moment?

microsetta_private_api/static/kit_check.js
```diff
@@ -0,0 +1,25 @@
+function verify_kit_claim() {
+    // Initialize form validation on the registration form.
+    // It has the name attribute "registration"
+    $("form[name='kit_claim_form']").validate({
```
AmandaBirmingham, line 4: Suggest changing
```
function verify_kit_claim() {
    // Initialize form validation on the registration form.
    // It has the name attribute "registration"
    $("form[name='kit_claim_form']").validate({ 
```

to 

```
function verify_kit_claim(form_name) {
    // Initialize form validation 
    let form_selector = "form[name='" + form_name + "']";
    $(form_selector).validate({
```

and then changing the calls to this function to read 

```
$(document).ready(verify_kit_claim('kit_claim_form'));
```

I am nervous about hardcoding the form name in a shared script library.

microsetta_private_api/templates/kit_sample_association.jinja2
```diff
@@ -4,36 +4,13 @@
     <meta charset="UTF-8">
     <title>Microsetta Kit Claim</title>
     <link rel="stylesheet" href="/static/vendor/bootstrap-4.4.1-dist/css/bootstrap.min.css" />
+    <link rel="stylesheet" href="/static/css/minimal_interface.css" />
```
AmandaBirmingham, line 7: Oooo, good catch!!

microsetta_private_api/static/kit_check.js
wasade: :+1:

wasade: I made this change, but it does not seem to be triggering the API calls to do the verification. My javascript-foo is weak, and I apologize if this is naive, but should `verify_kit_claim('kit_claim_form')` return a `function`? Right now the thing passed to `$(document).ready(...)` is the result of the call to `verify_kit_claim` whereas before it was the `function` itself

PR #156, opened 2020-05-16
St dh issue 155

### LINKED ISSUE #155: Environmental sample support: update sample page to support non human source
modify sample.jinja2 page to limit what users put in by which kind of source the sample goes to

dhakim87:
Fix #155

### DIFF
--- a/microsetta_private_api/example/client_impl.py
+++ b/microsetta_private_api/example/client_impl.py
@@ -630,16 +630,38 @@ def get_sample(account_id, source_id, sample_id):
     if next_state != ALL_DONE:
         return redirect(WORKFLOW_URL)
 
+    do_return, source_output, _ = ApiRequest.get(
+        '/accounts/%s/sources/%s' %
+        (account_id, source_id)
+    )
+    if do_return:
+        return source_output
+
     do_return, sample_output, _ = ApiRequest.get(
         '/accounts/%s/sources/%s/samples/%s' %
         (account_id, source_id, sample_id))
     if do_return:
         return sample_output
 
+    # Human settings
     sample_sites = ["Blood (skin prick)", "Stool", "Mouth", "Nares",
                     "Nasal mucus", "Right hand", "Left hand",
                     "Forehead", "Torso", "Right leg",  "Left leg",
                     "Vaginal mucus", "Tears",  "Ear wax", "Hair", "Fur"]
+    site_hint = None
+    site_req = True
+
+    is_environmental = source_output['source_type'] == 'environmental'
+
+    if is_environmental:
+        # Environment settings
+        sample_sites = [None]
+        site_hint = "As we cannot enumerate all possible sampling sites for " \
+               "environmental sources, we recommend describing the site " \
+               "the sample was taken from in as much detail as " \
+               "possible below"
+        site_req = False
+
 
     factory = VueFactory()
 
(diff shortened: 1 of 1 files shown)


microsetta_private_api/example/client_impl.py
```diff
@@ -630,16 +630,37 @@ def get_sample(account_id, source_id, sample_id):
...
     sample_sites = ["Blood (skin prick)", "Stool", "Mouth", "Nares",
                     "Nasal mucus", "Right hand", "Left hand",
                     "Forehead", "Torso", "Right leg",  "Left leg",
                     "Vaginal mucus", "Tears",  "Ear wax", "Hair", "Fur"]
+    site_hint = None
+    site_req = True
+
+    is_environmental = source_output['source_type'] == 'environmental'
+
+    if is_environmental:
```
wasade, line 656: Could we change this to:

```python
source_type = source_output['source_type']
if source_type == 'environmental':
    # Environment settings
    sample_sites = [None]
    site_hint = "As we cannot enumerate all possible sampling sites for " \
        "environmental sources, we recommend describing the site " \
        "the sample was taken from in as much detail as " \
        "possible below"
    site_req = False
elif source_type == 'human':
    pass
else:
    raise SomethingAppropriate()

# for environmental samples, sample_site selection is not allowed
# meaning it is not required. As of right now, there is no known
# scenario where sample_site is optional (i.e., not required but still
# presented to the user
site_selection_is_disabled = site_req
```

...and then on line 676, make `disabled=site_selection_is_disabled`

dhakim87: I believe this is captured in 929b79b

microsetta_private_api/example/client_impl.py
```diff
@@ -650,8 +671,10 @@ def get_sample(account_id, source_id, sample_id):
                    .set(required=True,
                         validator="string"))\
         .add_field(VueSelectField("sample_site", "Site", sample_sites)
-                   .set(required=True,
-                        validator="string")) \
+                   .set(required=site_req,
+                        validator="string",
+                        disabled=is_environmental,
+                        hint=site_hint)) \
```
AmandaBirmingham, line 684: Especially since there is no scenario in which site selection is not disabled for environmental samples, I don't really see the point of having THREE boolean variables that are all set from the same piece of info.  Why not just do 

```
.set(required=!is_environmental,
       validator="string",
       disabled=is_environmental,
       hint=site_hint))
```

... and drop `site_req` and `site_selection_disabled`?  To me, this minimal approach makes it very clear what controls those input settings.

microsetta_private_api/example/client_impl.py
```diff
@@ -630,17 +632,43 @@ def get_sample(account_id, source_id, sample_id):
...
     if do_return:
         return sample_output
 
-    sample_sites = ["Blood (skin prick)", "Stool", "Mouth", "Nares",
-                    "Nasal mucus", "Right hand", "Left hand",
-                    "Forehead", "Torso", "Right leg",  "Left leg",
-                    "Vaginal mucus", "Tears",  "Ear wax", "Hair", "Fur"]
-
+    is_environmental = source_output['source_type'] == 'environmental'
+    is_human = source_output['source_type'] == 'human'
```
AmandaBirmingham, line 649: I would really like it if we tested against Source.SOURCE_TYPE_HUMAN ...

PR #159, opened 2020-05-17
St dh issue 153

### LINKED ISSUE #153: Environmental sample support: allow user to pick samples
expand the add samples from kit functionality to let user pick sample

dhakim87:
Some bootstrap to pick samples from a kit and pass on to a new function to claim those samples for the source.  Since this is designed to move away from single sources and single kits, I'm not sure what to do about association of surveys, so let's discuss in the review.  

Our error handling also doesn't work well when making multiple requests, but I suspect we're going to punt on that for now.  

Fix #153


microsetta_private_api/example/client_impl.py
```diff
@@ -687,32 +687,67 @@ def put_sample(account_id, source_id, sample_id):
...
+    if error is None:
+        return json.dumps(True)
+    else:
+        return json.dumps(error)
+
+
+def get_list_kit_samples(kit_name):
+    kit, error, code = _get_kit(kit_name)
+    if error is None:
+        return flask.jsonify(kit), code
```
wasade, line 729: why use `flask.jsonify` here but `json.dumps` a few lines above?

microsetta_private_api/example/client_impl.py
```diff
@@ -687,32 +687,67 @@ def put_sample(account_id, source_id, sample_id):
...
+    else:
+        return flask.jsonify({"error": error}), code
+
+
+def claim_samples(account_id, source_id, body):
+    # TODO:  they could theoretically successfully claim some samples and
+    #  fail out when trying to claim others.  And I have no transaction support
+    #  here.  Boo...
+
+    # TODO:  How do I know what surveys to associate if any???
```
wasade, line 739: What about associating to all surveys (if there are some) that are associated to the `source_id`?

microsetta_private_api/example/client_impl.py
```diff
@@ -687,32 +687,67 @@ def put_sample(account_id, source_id, sample_id):
...
+def get_list_kit_samples(kit_name):
+    kit, error, code = _get_kit(kit_name)
+    if error is None:
+        return flask.jsonify(kit), code
+    else:
+        return flask.jsonify({"error": error}), code
+
+
+def claim_samples(account_id, source_id, body):
+    # TODO:  they could theoretically successfully claim some samples and
```
wasade, line 735: The alternative is to attempt all sample claims, collect errors if they occur and then return back to the user. I *think* the situation of failing to claim others is pathological, so I'm not sure if we definitely need to address it right now.

microsetta_private_api/example/client_impl.py
dhakim87: Because every time I used json.dumps the server sent back garbage and I didn't catch that one.  Dang.

dhakim87: Actually...  Those json.dumps calls above in post_check_acct_inputs were preserving existing behavior of post_check_acct_inputs from when I split out _get_kit as a helper function.  I'll check if they work as jsonify calls with javascript.

microsetta_private_api/example/client_impl.py
dhakim87: Punted to issue #160

microsetta_private_api/example/client_impl.py
dhakim87: I'll move forward with this approach, but I think we'll need to revisit this one fairly soon.  

I suspect this policy is correct for primary survey and possibly covid survey but wrong for FFQ.  It's also very dependent on keeping the minimalInterface as restrictive as it is - different orderings of calls against the API would make this policy incorrect.

dhakim87: Newly claimed samples will now be associated with any primary and covid survey on the source.

microsetta_private_api/templates/source.jinja2
```diff
@@ -3,15 +3,50 @@
...
+                    console.log(jqXHR.error);
+                },
+            });
+        }
+
         // Wait for the DOM to be ready
-        $(document).ready(verify_kit_claim('kit_claim_form'));
+
+        $(document).ready(function(){
+            $('#list_kit_form').ajaxForm()
```
AmandaBirmingham, line 46: The usage note at the top of `http://malsup.github.io/jquery.form.js` says
``` 
    Do not use both ajaxSubmit and ajaxForm on the same form.  These
    functions are mutually exclusive.  Use ajaxSubmit if you want
    to bind your own submit handler to the form.  
```
It appears we are doing this; can you clarify?

microsetta_private_api/templates/source.jinja2
```diff
@@ -3,15 +3,50 @@
...
+                                     '<input type="checkbox" name="sample_id" id="' + bc + '" value="' + sample_id + '"/>' +
+                                     '</div>' +
+                                     '</div>');
+                        $('#samples').append(item);
+                    }
+                    var div = $('#choose_sample_div')
+                    div.collapse('show')
+                },
+                error: function (jqXHR, textStatus, errorThrown) {
+                    console.log(jqXHR.error);
```
AmandaBirmingham, line 38: Although I hope we never get here, I think we need to do something more robust in case we do (especially as a lot of users won't even know how to look at their console log).  I would like to route errors to the error page; I suggest we do this with something akin to https://github.com/biocore/microsetta-private-api/blob/8aae43e4b8940c3054a8dec600b0d7cd0ed0a782/microsetta_private_api/templates/survey.jinja2#L24-L45 , without the `if (result_txt !== ""){` clause.

microsetta_private_api/static/kit_check.js
```diff
@@ -14,14 +14,13 @@ function verify_kit_claim(form_name) {
...
                     }
-                },
-                // Make sure the form is submitted to the destination defined
-                // in the "action" attribute of the form when valid
-                submitHandler: function (form) {
-                    form.submit();
                 }
             }
-        });
+        }
```
AmandaBirmingham, line 19: I cannot see how this code can be working the way it was intended, as AFAICT this code block comes *after* the return (`return function() { `...).  Am I missing something key?

microsetta_private_api/example/client_impl.py
```diff
@@ -374,13 +373,47 @@ def get_workflow_claim_kit_samples():
...
 
+def _claim_kit_samples_helper(acct_id,
+                              source_id,
+                              sample_ids_to_claim,
+                              all_answered_survey_ids):
+
+    # TODO:  Any of these requests may fail independently, but we don't have
+    #  a good policy to deal with partial failures.  Currently, we abort early
+    #  but that will result in some set of associations being already made,
+    #  one association failing, and the remaining associations not attempted.
```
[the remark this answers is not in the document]
dhakim87: Already punted to issue #160

microsetta_private_api/example/client_impl.py
```diff
@@ -374,13 +373,47 @@ def get_workflow_claim_kit_samples():
...
+        # Claim sample
+        do_return, sample_output, _ = ApiRequest.post(
+            '/accounts/{0}/sources/{1}/samples'.format(acct_id, source_id),
+            json={"sample_id": curr_sample_id}
+        )
+
+        if do_return:
+            return sample_output
+
+        # Associate all surveys of the source with this sample.
```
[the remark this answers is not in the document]
wasade: I believe this was resolved

microsetta_private_api/example/client_impl.py
```diff
@@ -687,32 +702,80 @@ def put_sample(account_id, source_id, sample_id):
...
-        return json.dumps(response_info)
+        error_msg = unable_to_validate_msg
+
+    if error_msg is not None:
+        return None, error_msg, response.status_code
+    return response.json(), None, response.status_code
+
+
+def post_check_acct_inputs(body):
+    kit, error, code = _get_kit(body[KIT_NAME_KEY])
```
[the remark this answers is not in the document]
wasade: I believe this was resolved

microsetta_private_api/templates/source.jinja2
```diff
@@ -3,15 +3,50 @@
...
+    <script src="/static/vendor/js/jquery-3.4.1.min.js"></script>
+    <script src="/static/vendor/js/jquery.validate.min.js"></script>
+    <script src="/static/vendor/bootstrap-4.4.1-dist/js/bootstrap.min.js"></script>
     <link rel="stylesheet" href="/static/vendor/bootstrap-4.4.1-dist/css/bootstrap.min.css" />
     <link rel="stylesheet" href="/static/vendor/font-awesome-4.7.0/css/font-awesome.min.css"/>
     <link rel="stylesheet" href="/static/css/minimal_interface.css" />
-    <script src="/static/vendor/js/jquery-3.4.1.min.js"></script>
-    <script src="/static/vendor/js/jquery.validate.min.js"></script>
-    <script src="/static/kit_check.js"></script> 
+    <script src="http://malsup.github.com/jquery.form.js"></script>
```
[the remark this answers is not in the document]
wasade: If this library is being used then I think we would want to pull it down to the static directory. Note that it's necessary to update `setup.py` too to make sure the assets carry

microsetta_private_api/static/kit_check.js
dhakim87: So... note that I can't tell if this section was edited after you made this comment, so apologies if it was broken and something that you already fixed.  
My javascript is extremely rusty, but here's my intention.  

verify_kit_claim should create and return a function.  That function takes no arguments, starts on line 4 (return function() { ) and continues until line 24.  That function will be called when the document is ready.  The function initializes a dictionary called validateOptions with a key "rules" and values set to allow checking the kit_name field with a remote ajax post for verification.  If the caller of verify_kit_claim passed in a submitHandler callback, the validateOptions dictionary gets another key "submitHandler" added to it, which contains the callback.  

Jquery's .validate() specifically searches for both the "rules" and "submitHandler" keys in the validateOptions dictionary.  The "submitHandler" callback is called when the form passes validation.  

I wish github highlighted open and close brackets in diffs

microsetta_private_api/templates/source.jinja2
dhakim87: It feels really weird to me that a particular url could return either json or html.  Why wouldn't the json return something like {"error":<blahblahblah>, "redirect": "<blahblahblah>/error?error_message=<blahblahblahblah>"} or {"errorcode":<blahblah>, "error_msg":"blahblah"} and let the ajax write that error message underneath the button the same way the jquery.validate does?

microsetta_private_api/templates/source.jinja2
dhakim87: Removed that line, I thought it was necessary to initialize the form for the library but you're right that they're mechanisms.

microsetta_private_api/templates/source.jinja2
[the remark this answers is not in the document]
dhakim87: I've included the distribution of https://github.com/jquery-form/form which appears to be the latest version of what was hosted at malsup.  No change to setup.py because the files are in static/vendor/js

microsetta_private_api/templates/source.jinja2
```diff
@@ -32,11 +34,22 @@
...
                 },
                 error: function (jqXHR, textStatus, errorThrown) {
-                    console.log(jqXHR.error);
+                    error_txt = jqXHR.responseText;
                 },
+                complete: function (data, textStatus, outputObj) {
+                    if (successful)
+                        return;
+                    if (error_txt === "")
+                        error_txt = textStatus
```
wasade, line 47: semicolon plz?

microsetta_private_api/example/client_impl.py
```diff
@@ -644,7 +644,7 @@ def get_sample(account_id, source_id, sample_id):
         return sample_output
 
     source_type = source_output['source_type']
-    is_environmental = source_type == Source.SOURCE_TYPE_ENVIRONMENTAL
```
wasade, line 653: I think that was my bad... sorry

microsetta_private_api/templates/source.jinja2
dhakim87: Semicolon!

PR #164, opened 2020-05-19
St dh issue 162

### LINKED ISSUE #162: Do not display Food Frequency Questionnaire link for samples linked to environmental source
Currently the food frequency questionnaire link is shown on source.jinja2 for any sample that is not associated with an answered food frequency questionnaire:

Screen Shot 2020-05-19 at 3 42 33 PM

Desired behavior is that these items are shown only for any sample that is not associated with an answered food frequency questionnaire AND is associated with a source for which the food frequency questionnaire is possible.

We also will not want to put the text about not filling the ffq out more than once on the screen if ffq is not possible for the source in question.

dhakim87:
Fix #162 

FFQ not displayed for non human sources.


microsetta_private_api/templates/source.jinja2
```diff
@@ -81,7 +81,7 @@
         <div class="container">
             {% if samples|length > 0 %}
                 Samples:
-                {% if samples|length > 1 %}
+                {% if samples|length > 1 and source_type == "human" %}
                 <p><i>If collecting all samples on the same day, please only complete a single Food Frequency Questionnaire.</i></p>
```
AmandaBirmingham, line 85: I haaaaate hardcoded comparisons to strings that copy those defined elsewhere, especially repeated comparisons like this.  Can we please pass in `is_human_source` instead of `source_type`, where `is_human_source` can be set in `client_impl.get_source` via a comparison against Souce.SOURCE_TYPE_HUMAN?

PR #166, opened 2020-05-20
UI tweaks

wasade:
A few UI tweaks to draw attention help users

### DIFF
--- a/microsetta_private_api/templates/account.jinja2
+++ b/microsetta_private_api/templates/account.jinja2
@@ -25,8 +25,16 @@
 {% block content %}
         <div class="container">
             {%  if sources|length > 0 %}
-            Choose a source:
+            Choose a source by clicking on the name:
             <div class="list-group">
+                <div class="row">
+                    <div class="col-sm">
+                        <i>Name</i>
+                    </div>
+                    <div class="col-sm">
+                        <i>Source-type</i>
+                    </div>
+                </div>
                 {% for source in sources %}
                 <div class="container list-group-item {{loop.cycle('odd', 'even') }}">
                   <div class="row">
@@ -42,20 +50,18 @@
                 </div>
                 {% endfor %}
             </div>
-            <p>
-                or
-            </p>
+            <br/><br/>
             {% endif %}
         </div>
         <div class="container">
-            Create a new environmental source:
+            Optionally, you can create a new environmental source:
             <form method="post" name="new_source_form" action="/accounts/{{acct_id}}/sources/">
                 <div class="form-group">
-                    <label for="source_name" name="source_name_label">Name</label>
+                    <label for="source_name" name="source_name_label">Environment name (e.g., microwave):</label>
                     <input id="source_name" name="source_name" class="form-control" type="text"/>
                     <input id="source_type" name="source_type" type="hidden" value="environmental" />
                     <br />
-                    <label for="source_description" name="source_description_label">Description</label>
+                    <label for="source_description" name="source_description_label">Environment description (optional):</label>
                     <input id="source_description" name="source_description" class="form-control" type="text"/>
                  </div>
                 <input type="submit" class="btn btn-primary" value="Create Source"/>
--- a/microsetta_private_api/templates/account.jinja2
+++ b/microsetta_private_api/templates/account.jinja2
@@ -25,14 +25,14 @@
 {% block content %}
         <div class="container">
             {%  if sources|length > 0 %}
-            Choose a source by clicking on the name:
+            Choose a source by clicking on a name:
             <div class="list-group">
                 <div class="row">
                     <div class="col-sm">
                         <i>Name</i>
                     </div>
                     <div class="col-sm">
-                        <i>Source-type</i>
+                        <i>Source type</i>
                     </div>
                 </div>
                 {% for source in sources %}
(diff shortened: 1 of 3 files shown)


microsetta_private_api/templates/account.jinja2
```diff
@@ -25,8 +25,16 @@
 {% block content %}
         <div class="container">
             {%  if sources|length > 0 %}
-            Choose a source:
+            Choose a source by clicking on the name:
```
AmandaBirmingham, line 28: Suggest either "on *a* name" or "on *its* name" rather than "on the name"; "on a name" would be consistent with the wording use for samples below ("click *an* identifier").

microsetta_private_api/templates/account.jinja2
```diff
@@ -25,8 +25,16 @@
...
             <div class="list-group">
+                <div class="row">
+                    <div class="col-sm">
+                        <i>Name</i>
+                    </div>
+                    <div class="col-sm">
+                        <i>Source-type</i>
```
AmandaBirmingham, line 35: Can we make this `Source type`?  I don't think the hyphenation will be helpful/meaningful to end users

microsetta_private_api/templates/account.jinja2
```diff
@@ -42,23 +50,21 @@
...
             </div>
-            <p>
-                or
-            </p>
+            <br/><br/>
             {% endif %}
         </div>
         <div class="container">
-            Create a new environmental source:
+            Optionally, you can create a new environmental source:
```
AmandaBirmingham, line 57: It appears that all the other one-line instructions are in the form of commands "do x", whereas this switches over to a descriptive form "you can do x"; this jars me a bit.  Could it be "Optionally, create a new ..."?

microsetta_private_api/templates/account.jinja2
```diff
@@ -42,23 +50,21 @@
...
             <form method="post" name="new_source_form" action="/accounts/{{acct_id}}/sources/">
                 <div class="form-group">
-                    <label for="source_name" name="source_name_label">Name</label>
+                    <label for="source_name" name="source_name_label">Environment name (e.g., microwave):</label>
```
AmandaBirmingham, line 60: +1

microsetta_private_api/templates/source.jinja2
```diff
@@ -71,20 +71,42 @@
...
             {% if samples|length > 0 %}
-                Samples:
+                Samples (click an identifier to assign):
                 {% if samples|length > 1 and is_human %}
                 <p><i>If collecting all samples on the same day, please only complete a single Food Frequency Questionnaire.</i></p>
                 {% endif %}
                 <div class="list-group">
+                    <div class="row">
+                        <div class="col-sm">
+                            <i>Identifier</i>
```
AmandaBirmingham, line 96: Why not "Barcode", which is something the user could actually match up to the physical item they have?

microsetta_private_api/templates/source.jinja2
```diff
@@ -71,20 +71,42 @@
...
+                        </div>
+                        <div class="col-sm">
+                            <i>Collection date</i>
+                        </div>
+                        {% if is_human %}
+                        <div class="col-sm">
+                            <i>Sample type</i>
+                        </div>
+                        <div class="col-sm">
+                            <i>Sample specific survey</i>
```
AmandaBirmingham, line 106: Here we *do* need a hyphen: `Sample-specific` :)

microsetta_private_api/templates/source.jinja2
```diff
@@ -71,20 +71,42 @@
...
           <ol class="breadcrumb">
             <li class="breadcrumb-item"><a href="/home">Home</a></li>
             <li class="breadcrumb-item"><a href="/accounts/{{acct_id}}">Sources</a></li>
+            <li class="breadcrumb-item"><a>{{source_name}}</a></li>
           </ol>
         </nav>
         <div class="container">
             {% if samples|length > 0 %}
-                Samples:
+                Samples (click an identifier to assign):
```
AmandaBirmingham, line 89: Ok, this wording is confusing me and I wrote half this stuff :) "Assign" is not a term we have used in the interface (except 2x in the faq, which however doesn't explain what that term *means* ... :( ) so it seems problematic to use it here without a definition.   AFAICT, what we want the user to do here is not "assigning" a sample in the sense that we use the word on the back-end (which is probably contributing to my confusion) ... we want the user to provide info on what a particular sample is, right?  See above comment re need for more description ...

Trivially, to match formulation on Sources page, should be "click *on* an identifier ..."

microsetta_private_api/templates/source.jinja2
```diff
@@ -71,20 +71,42 @@
     <br />
     <br />
 
+    {% if needs_assignment %}
+    <div class="alert alert-warning" role="alert">
+        IMPORTANT: samples must be assigned in order for the lab to process them.
```
AmandaBirmingham, line 76: I don't think this is likely to make things much clearer to users ... Honestly, I think we need considerably more descriptive text on these pages. It is unrealistic to expect we will find one word or one phrase ("assign samples", etc) that will make these somewhat complicated concepts clear to a person using this website only once or twice ... I think we need brief but full-sentence instructions, e.g., "Once you have set up a source, you can specify which swabs were used to sample that source; you can then fill in the details about when and how these samples were taken", yadda yadda). Can we either generate such text ourselves or ask Dom/Edgar/whoever understands what the users know (and don't know) to generate it?

PR #168, opened 2020-05-22
Bugfix for legacy users with null last name

dhakim87:
Due to crash on logging in legacy users, we set nullable string fields of account table that can appear in yaml object to empty string

### DIFF
--- a/microsetta_private_api/db/patches/0062.sql
+++ b/microsetta_private_api/db/patches/0062.sql
@@ -11,7 +11,7 @@
 -- last_name, Yes
 -- street, Yes
 -- city, Yes
--- state, Yes
+-- state, No
 -- post_code, Yes
 -- country_code, Yes
 -- latitude, No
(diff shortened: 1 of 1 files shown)


wasade:
Should we also alter table to require these fields are not null?

microsetta_private_api/db/patches/0062.sql
```diff
@@ -24,21 +24,28 @@
...
     SET street=''
     WHERE street is null;
+ALTER TABLE ag.account ALTER COLUMN street SET NOT NULL;
 UPDATE ag.account
     SET city=''
     WHERE city is null;
+ALTER TABLE ag.account ALTER COLUMN city SET NOT NULL;
 UPDATE ag.account
     SET state=''
     WHERE state is null;
```
wasade, line 42: state should be nullable as that isn't a generalized concept across countries

PR #169, opened 2020-05-22
POSSIBLE BUG FIX: Nullable account fields in yaml

AmandaBirmingham:
Makes account's first_name and last_name, and address's street, city, state, post_code, and country_code nullable.  Adds test showing this change allows loading of existing account record in which these fields are null.

NOTE: We will merge this PR *or* #169 (Bugfix for legacy users with null last name) but not both; which depends on requirements decision by @wasade .


wasade:
Closing per Slack discussion on bug resolution

PR #170, opened 2020-05-27
WIP: State rework

AmandaBirmingham:
Let's discuss this before considering any merging.  It is a major rework of client_impl.py to make the state machine capable of handling multiple human sources (in the way @dhakim87 and I discussed on Friday), with lots of knock-on changes in the client.yaml and the jinja2 templates.

Note that it is failing build right now because of lint failures in files that HAVEN'T changed between this branch and master, which has me scratching my head ...


microsetta_private_api/example/client_impl.py
```diff
@@ -683,133 +761,93 @@ def get_sample(account_id, source_id, sample_id):
...
 
 
-def put_sample(account_id, source_id, sample_id):
-    next_state, current_state = determine_workflow_state()
-    if next_state != ALL_DONE:
-        return redirect(WORKFLOW_URL)
+# TODO: guess we should also rewrite as ajax post for sample vue form?
+def put_update_sample(account_id, source_id, sample_id):
+    prereqs_step, curr_state = _check_relevant_prereqs(account_id, source_id)
+    # TODO: here we might want a "sample prereqs met"?
```
wasade, line 774: What type of checks might be performed?

microsetta_private_api/example/client_impl.py
```diff
@@ -683,133 +761,93 @@ def get_sample(account_id, source_id, sample_id):
...
-                             error_msg=error_msg,
-                             endpoint=SERVER_CONFIG["endpoint"],
-                             authrocket_url=SERVER_CONFIG["authrocket_url"])
-
-    return output
+    result = kit if error is None else error
+    return flask.jsonify(result), code
+
+
+# TODO: Note that associating surveys with samples when samples are claimed
```
wasade, line 805: This is the correct behavior here. If we need to associate samples to other surveys after the fact, I think that would fall into a separate UI element and backend method to resolve.

microsetta_private_api/example/client_impl.py
```diff
@@ -683,133 +761,93 @@ def get_sample(account_id, source_id, sample_id):
...
+#  means that any surveys added to this source AFTER these samples are claimed
+#  will NOT be associated with these samples.  This might be the right behavior
+#  but we should probably make an explicit policy decision about that :)
+def post_claim_samples(account_id, source_id, body):
+    prereqs_step, curr_state = _check_relevant_prereqs(account_id, source_id)
+    if prereqs_step == SOURCE_PREREQS_MET:
+        sample_ids_to_claim = body.get('sample_id', None)
```
wasade, line 812: `, None)` isn't needed

microsetta_private_api/example/client_impl.py
```diff
@@ -683,133 +761,93 @@ def get_sample(account_id, source_id, sample_id):
...
+        if sample_ids_to_claim is None:
+            # User claimed no samples ... shrug
+            return _route_to_closest_sink(prereqs_step, curr_state)
+
+        has_error, survey_output, _ = ApiRequest.get(
+            '/accounts/{0}/sources/{1}/surveys'.format(account_id, source_id))
+        if has_error:
+            return survey_output
+
+        # TODO: this will have to get more nuanced when we add animal surveys?
```
wasade, line 822: Ah good catch. Let's defer animals to a separate PR as there are a few changes in here already...

microsetta_private_api/templates/home.jinja2
```diff
@@ -28,21 +28,15 @@
...
-                        <a href="/workflow" class="list-group-item list-group-item-action">{{user}}'s account</a>
-                    {% endif %}
-                {% else %}
-                <a href="/workflow" class="list-group-item list-group-item-action">
+            {% if accounts|length > 0 %}
+                {% for account in accounts %}
+                    <a href="/accounts/{{ account.account_id }}" class="list-group-item list-group-item-action">{{ account.email }}'s account</a>
+                {% endfor %}  <!-- end of for account in accounts -->
+            {% else %}
+                <a href="/create_account" class="list-group-item list-group-item-action">
```
wasade, line 36: Just to check, a user can only create a new account if there aren't any already in the system? I think this makes sense

microsetta_private_api/templates/home.jinja2
```diff
@@ -28,21 +28,15 @@
...
-                        and we’re still in the process of handling the different type of scenarios
-                        that can happen on the website. We apologize for the inconvenience.
-                    {% else %}
-                        <a href="/workflow" class="list-group-item list-group-item-action">{{user}}'s account</a>
-                    {% endif %}
-                {% else %}
-                <a href="/workflow" class="list-group-item list-group-item-action">
+            {% if accounts|length > 0 %}
+                {% for account in accounts %}
+                    <a href="/accounts/{{ account.account_id }}" class="list-group-item list-group-item-action">{{ account.email }}'s account</a>
```
wasade, line 33: what happens if the emails aren't unique?

microsetta_private_api/templates/source.jinja2
```diff
@@ -10,19 +10,43 @@
...
+                        // on the right side
+                        kit_name: {
+                            required: true,
+                            remote: {
+                                url: "/check_kit_valid",
+                            }
+                        }
+                    }
+                };
+                if (submitHandler != null){
```
wasade, line 32: should this be `!==`?

microsetta_private_api/templates/source.jinja2
```diff
@@ -164,7 +187,7 @@
                 <p>Surveys taken:</p>
                 <div class="list-group">
                     {% for survey in surveys %}
-                    <a>
+                    <a> <!--TODO: Err, should this link go somewhere? -->
```
wasade, line 190: if/when we allow editing a survey...

microsetta_private_api/example/client_impl.py
```diff
@@ -26,219 +27,371 @@
...
-        "/accounts/%s/sources" % (acct_id,), params={"source_type": "human"})
-    if needs_reroute:
-        current_state["reroute"] = sources_output
-        return NEEDS_REROUTE, current_state
-    if len(sources_output) == 0:
-        return NEEDS_HUMAN_SOURCE, current_state
-    elif len(sources_output) > 1:
-        # we do not currently support displaying multiple human sources
-        return TOO_MANY_HUMAN_SOURCES, current_state
+    # IF we decide that every acct needs at least one source,
```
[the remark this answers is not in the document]
dhakim87: Admin accounts probably won't...  But I don't think that's really relevant to our policies for regular users.

microsetta_private_api/templates/home.jinja2
dhakim87: I think the "idx_account_email" UNIQUE, btree (email) on ag.account guarantees account emails are unique.

microsetta_private_api/example/client.yaml
```diff
@@ -98,49 +86,57 @@ paths:
...
+          description: Display of account
           content:
             text/html:
               schema:
                 type: string
-        '302':
-          description: Redirecting to necessary action
 
-  '/workflow_create_human_source':
+  '/accounts/{account_id}/update_email':
```
dhakim87, line 118: Comment that's more related to my account information editing branch.  Do we want users to be able to edit their email addresses whenever they want?  I disabled it, but it felt weird not letting them set their own email address.

microsetta_private_api/example/client.yaml
```diff
@@ -189,38 +189,22 @@ paths:
...
             text/html:
               schema:
                 type: string
-
-  '/accounts/{account_id}/sources/':
     post:
-      operationId: microsetta_private_api.example.client_impl.post_source
+      operationId: microsetta_private_api.example.client_impl.post_create_nonhuman_source
       tags:
         - Sources
```
dhakim87, line 209: I might be getting tripped up by the github diff tool, but it seems this line should be updated to match singular style in naming convention?

microsetta_private_api/example/client_impl.py
```diff
@@ -26,219 +27,371 @@
...
+    SURVEY_TEMPLATE_ID_KEY = "survey_template_id"
+    current_state = {} if current_state is None else current_state
+    current_state['source_id'] = source_id
 
-    source_id = sources_output[0]["source_id"]
-    current_state['human_source_id'] = source_id
+    # Get the input source
+    has_error, source_output, _ = ApiRequest.get(
+        '/accounts/%s/sources/%s' %
+        (acct_id, source_id))
```
dhakim87, line 151: Should this use _make_path?

microsetta_private_api/example/client_impl.py
```diff
@@ -26,219 +27,371 @@
...
+
+    # Get all required survey template ids for this source type
+    req_survey_template_ids = _get_required_survey_templates_by_source_type(
+        source_output["source_type"])
 
-    # Have you taken the primary survey? NO-> main_survey.html
+    # Get all the current answered surveys for this source
     needs_reroute, surveys_output, _ = ApiRequest.get(
-        "/accounts/{0}/sources/{1}/surveys".format(acct_id, source_id))
+        '/accounts/{0}/sources/{1}/surveys'.format(acct_id, source_id))
```
dhakim87, line 161: Should this use _make_path?

microsetta_private_api/example/client_impl.py
```diff
@@ -26,219 +27,371 @@
...
+        # Does this source LACK an answered survey with this template id?
+        if curr_req_survey_template_id not in template_ids_of_answered_surveys:
+            current_state["needed_survey_template_id"] = \
+                curr_req_survey_template_id
+            return NEEDS_SURVEY, current_state
+
+    return SOURCE_PREREQS_MET, current_state
+
+
+def _check_relevant_prereqs(acct_id=None, source_id=None):
```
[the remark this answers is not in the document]
dhakim87: Inverting the if statements matches my intuition as well.  I have trouble following this method even though it's completely linear.  

I also wonder if there's a way to switch up the function prototypes to build something like:
`
if not _check_home_prereqs(output)
    return output
if not _check_acct_prereqs(acct_id, output)
    return output
_check_source_prereqs(acct_id, source_id, output)
return output
`

where output just denotes some mutable object (list/dict) that is modified in place.

microsetta_private_api/example/client_impl.py
```diff
@@ -26,219 +27,371 @@
...
+# get the token, validate it, and pull email out of it.
+def _parse_jwt(token):
+    decoded = jwt.decode(token, PUB_KEY, algorithms=['RS256'], verify=True)
+    email_verified = decoded.get('email_verified', False)
+    return decoded["email"], email_verified
 
-    return ALL_DONE, current_state
 
+def _route_to_closest_sink(prereqs_step, current_state):
+    print("Current Prereq Step:", prereqs_step)
```
dhakim87, line 206: Leftover print statement

microsetta_private_api/example/client_impl.py
```diff
@@ -26,219 +27,371 @@
...
+# Client might not technically care who the user is, but if they do, they
+# get the token, validate it, and pull email out of it.
+def _parse_jwt(token):
+    decoded = jwt.decode(token, PUB_KEY, algorithms=['RS256'], verify=True)
+    email_verified = decoded.get('email_verified', False)
+    return decoded["email"], email_verified
 
-    return ALL_DONE, current_state
 
+def _route_to_closest_sink(prereqs_step, current_state):
```
dhakim87, line 205: It might be coming up on the time to turn this function into a flow chart.  I don't fully know how to interpret the two classes of steps: NEEDS_XXX, and XXX_PREREQS_MET.  I understand why I would interrupt the user to say "No, you NEED to do this first"  I don't understand why I would redirect the user to say "Okay, you have an account, so go look at it", which makes me have trouble following this function.

microsetta_private_api/example/client_impl.py
```diff
@@ -26,219 +27,371 @@
...
+    else:
+        # TODO: theoretically could route this to the front end /error ..
+        # but kind of a waste?  Do we need to reroute all these to front end?
+        # Guess doing so means we can't accidentally change front end behavior
+        # but not back-end behavior here ...
+        return get_show_error_page(
+            "Unknown prereq_step: '{0}'".format(prereqs_step))
+
+
+def _update_state_and_route_to_sink(account_id=None, source_id=None):
```
dhakim87, line 241: Naming?  Update what state?  Is curr_state visible outside the scope of this function in some way?

microsetta_private_api/example/client_impl.py
```diff
@@ -26,219 +27,371 @@
...
+        error_msg = unable_to_validate_msg
+
+    if error_msg is not None:
+        return None, error_msg, response_status_code
+
+    return response.json(), None, response_status_code
+
+
+def _check_survey_allowed(account_id, source_id, survey_template_id,
+                          addtl_allowed_steps=None):
```
dhakim87, line 280: Needs a doc string, I can't tell what this function is for or what to put into addtl_allowed_steps

microsetta_private_api/example/client_impl.py
```diff
@@ -26,219 +27,371 @@
...
+            # session, so if our token goes back we need to force them to login
+            # again.
+            user, email_verified = _parse_jwt(session[TOKEN_KEY_NAME])
+        except jwt.exceptions.ExpiredSignatureError:
+            return redirect('/logout')
+
+        if email_verified:
+            home_step, curr_state = _check_home_prereqs()
+            if home_step == NEEDS_REROUTE:
+                return curr_state[REROUTE_KEY]
```
dhakim87, line 347: This usage pattern is a little verbose.  Is there a way we can make it read more like:
```
if not _check_home_prereqs(output):
    return output
```

dhakim87: I miss C#'s out parameters 😢

microsetta_private_api/example/client_impl.py
```diff
@@ -26,219 +27,371 @@
...
+        # delete these keys if they are here, otherwise ignore
+        session.pop(TOKEN_KEY_NAME, None)
+        session.pop(KIT_NAME_KEY, None)
+        session.pop(EMAIL_CHECK_KEY, None)
+    return redirect(HOME_URL)
+
+
+def get_create_account():
+    prereqs_step, curr_state = _check_relevant_prereqs()
+    if prereqs_step != NEEDS_ACCOUNT:
```
dhakim87, line 384: I think my main confusion is right here.  I don't fundamentally understand why we are checking if a user needs to create an account rather than whether a user is allowed to create an account based on their prerequisites.  

I think when we discussed this, we talked about duplicate post operations, and the user getting out of sync with their current place in the workflow.  Is this trying to address the problem of duplicate post operations?  I can understand a worry that a user might hit the back button and post twice to the same route, but I don't think that should be addressed within client_impl.  If that's the problem we're trying to solve, we need to move that inside the api layer to check the database and ensure individual sources can't take two primary surveys for example.

microsetta_private_api/example/client_impl.py
```diff
@@ -256,258 +409,261 @@ def post_workflow_create_account(body):
...
+        has_error, consent_output, _ = ApiRequest.post(
+            "/accounts/{0}/consent".format(account_id), json=body)
+        if has_error:
+            return consent_output
 
-def _claim_kit_samples_helper(acct_id,
-                              source_id,
-                              sample_ids_to_claim,
-                              answered_survey_ids_to_associate):
+    new_source_id = consent_output["source_id"]
```
dhakim87, line 503: Going a bit blind with github's diffs, but I believe that if prereqs_step is not ACCT_PREREQS_MET, this line will crash as consent_output is not declared

microsetta_private_api/example/client_impl.py
```diff
@@ -256,258 +409,261 @@ def post_workflow_create_account(body):
...
-        acct_id = current_state["account_id"]
-        source_id = current_state["human_source_id"]
-        primary_survey_id = current_state["answered_primary_survey_id"]
-        covid_survey_id = current_state["answered_covid_survey_id"]
+def get_fill_local_source_survey(account_id, source_id, survey_template_id):
+    # if we are filling out a source-level survey, it must come before the
+    # source prerequisites are met; if a sample-level one, it can come after
+    reroute = _check_survey_allowed(account_id, source_id, survey_template_id)
+    if reroute is not None:
+        return reroute
```
dhakim87, line 532: I'm so sad that python doesn't have out parameters right now

microsetta_private_api/templates/home.jinja2
dhakim87: I believe attempting to make a second account with the same authrocket token errors out at the database layer due to "idx_account_issuer_sub" UNIQUE, btree (auth_issuer, auth_sub)

microsetta_private_api/templates/source.jinja2
```diff
@@ -10,19 +10,43 @@
     <link rel="stylesheet" href="/static/vendor/font-awesome-4.7.0/css/font-awesome.min.css"/>
     <link rel="stylesheet" href="/static/css/minimal_interface.css" />
     <script src="/static/vendor/js/jquery.form-4.2.2/jquery.form.min.js"></script>
-    <script src="/static/kit_check.js"></script>
```
dhakim87, line 13: Why was this functionality moved to source.jinja2 from being a separate file?  I thought this was used in both account registration and in sample claiming.

microsetta_private_api/templates/home.jinja2
wasade: Existing users with multiple source possibly used the same email address on the consent document, so this may be violated

microsetta_private_api/example/client_impl.py
AmandaBirmingham: I didn't use _make_path for any private API requests--I only used it for client URLs.

Right now it is in fact the case that the paths for the client URLs and the paths for the private API URLs have the same structure (/accounts/{acct_id}/sources/{source_id}/<etc>) but since we had not made any formal decision that committed to them mirroring each other, I was hesitant to use _make_path for both.  Would be happy to put that in, though, if we want to commit to that mirroring.

microsetta_private_api/example/client_impl.py
AmandaBirmingham: See above

microsetta_private_api/example/client_impl.py
AmandaBirmingham: Yeah, we've had this one for basically the entire lifetime of minimalInterface; this is the print statement that is near the top of `workflow()` in the existing master (refactored into the function that one became):

https://github.com/biocore/microsetta-private-api/blob/9d7195d2552db0fe3f194185a8e07bc8ebf752ed/microsetta_private_api/example/client_impl.py#L214-L217

I have found it pretty helpful so I left it ... but I suppose in reality it should be logging instead ...

microsetta_private_api/example/client_impl.py
AmandaBirmingham: Fair point re naming!  This method does not change state in any way :) Rather, it calls _check_relevant_prereqs, which returns updated info about what the state of the user's account/source/etc is, and then passes that updated "state" information to _route_to_closest_sink to send the user to the location appropriate to that state.  Maybe it would be clearer to refer to this as *refreshing* the state rather than *updating* it?

microsetta_private_api/example/client_impl.py
AmandaBirmingham: Definitely open to suggestions on how to do that!

microsetta_private_api/example/client_impl.py
AmandaBirmingham: Note that this is on the *get*, not the post (the post *also* has a check on it, but it is a little different in form and function).  This check is here to guard against doing a "refresh" or a "back" to display a form for step that it turns out has (through some out-of-band action) already been performed.  E.g., if someone copies the url that points to the create account form and then pastes it into their browser at some future point after having actually created their account, they won't even see that create account form again--they will be redirected straight to the page for the account they already made.  

I didn't introduce this sort of checking in this WIP; it has been in minimalInterface since the beginning (see examples below from the current master).  Guarding against improper form gets in the client via the private api would not be feasible as the client's `get` endpoint sometimes doesn't make an api call at all (see e.g. `get_workflow_create_account` and `get_workflow_update_email` in the current master).  Therefore, I think we would instead be deciding whether we want to keep doing such aggressive checking on gets.  Another option would be to go ahead and let the user potentially do stupid things, like `getting` create_account form AGAIN and filling it out AGAIN, and then simply e.g., ignore its contents when they are submitted because, well, the user already had an account :D 

https://github.com/biocore/microsetta-private-api/blob/9d7195d2552db0fe3f194185a8e07bc8ebf752ed/microsetta_private_api/example/client_impl.py#L243-L246

https://github.com/biocore/microsetta-private-api/blob/9d7195d2552db0fe3f194185a8e07bc8ebf752ed/microsetta_private_api/example/client_impl.py#L281-L284

https://github.com/biocore/microsetta-private-api/blob/9d7195d2552db0fe3f194185a8e07bc8ebf752ed/microsetta_private_api/example/client_impl.py#L320-L323

etc :)

microsetta_private_api/example/client_impl.py
AmandaBirmingham: Yesss ...

microsetta_private_api/templates/source.jinja2
AmandaBirmingham: > I thought this was used in both account registration and in sample claiming
Nope :) This functionality was used in kit_sample_association.jinja2 and in source.jinja2 (but *not* in create_account.jinja2). However, once @wasade had added this functionality to source.jinja2, kit_sample_association.jinja2 was no longer needed.  It has been removed from the codebase (you can see it isn't there in master anymore) -- which means the functionality in kit_check.js was used in only one place, so I put into that place directly :)

microsetta_private_api/example/client_impl.py
dhakim87: Well... It's not pretty, but you can do
```
if ((home_step, curr_state) := _check_home_prereqs())[0] == NEEDS_REROUTE:
    return curr_state[REROUTE_KEY]
```
🤮 
And only as of python 3.8

If curr_state was passed in and mutated it would be two lines on average for the repeated calls, but that doesn't help all the functions that then have to declare curr_state to begin with.

microsetta_private_api/example/client_impl.py
dhakim87: Hmm... I think we agree on what it does and what the other option is then.  I just think that if the only way the user gets to these pages is hitting back or copy pasting links, they deserve to get an error message on form submission, at least if doing so would allow us to simplify the design.  That and we need to ensure that they actually do get an error on form submission out of the API, rather than anything enforced by the client impl - otherwise a sufficiently motivated user can still do dumb things like submit double covid surveys.

microsetta_private_api/templates/home.jinja2
dhakim87: I think the ag_consent emails are in ag.source's participant_email field, which doesn't have a uniqueness constraint on it.

microsetta_private_api/example/client_impl.py
```diff
@@ -26,219 +27,371 @@
...
-def home():
-    user = None
-    email_verified = False
-    acct_id = None
-    has_multiple_hs_sources = False
+def _get_required_survey_templates_by_source_type(source_type):
+    if source_type == Source.SOURCE_TYPE_HUMAN:
+        return [1, 6]
+    elif source_type == Source.SOURCE_TYPE_ANIMAL:
+        return []
```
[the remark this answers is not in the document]
AmandaBirmingham: @wasade Good to know, but I think splitting the current non-human source creation into Animal vs Environmental is a whole 'nother kettle of fish.  I have created issue #193 to capture this need.

microsetta_private_api/example/client_impl.py
AmandaBirmingham: Excellent catch, thank you!!

microsetta_private_api/example/client_impl.py
AmandaBirmingham: @dhakim87 I'm tabling this one in light of the rewrite we discussed in our 1pm meeting Friday.  

One final thought, though: You have a good point about not enforcing "hard" constraints on back-end behavior through the client, since it is possible to make an end-run around the client by simply using the private api directly.  However, I am not sure that the primary survey is an example of such a hard constraint--RIGHT NOW we aren't allowing multiple primary surveys, but as I understand it that is a temporary limitation.  The private api explicitly supports this because we really *do* want users to be able to update info about e.g. where they live and what diseases they have been diagnosed with.  I guess we could temporarily hobble the private api to remove this functionality, but I feel like wherever we put that limit, it is going to be a temporary patch that we remove once we are ready to fully realize our data model.

microsetta_private_api/example/client_impl.py
AmandaBirmingham: I think this comment will be resolved by the work we discussed on Friday to rejigger the handling of errors originating from the private api.  I have captured that in new issue #198 Revise client_impl.py's handling of private-api errors

microsetta_private_api/example/client.yaml
AmandaBirmingham: @dhakim87 I don't understand this comment. Could you clarify so I know what to fix? :D

microsetta_private_api/example/client.yaml
AmandaBirmingham: @dhakim87  I think you handled letting them update their email in the account editing branch, right?  

Given that, do you want this endpoint to renamed to something else, like "auto-update_email" or "align_emails" or something, to differentiate it from the email update capability in the account editing?

microsetta_private_api/templates/source.jinja2
AmandaBirmingham: Indeed it should :) Thanks

microsetta_private_api/example/client.yaml
dhakim87: The account editing branch shows the email, but the user isn't able to edit it, the input is marked readonly.  I wasn't sure what to do with it since you already had another mechanism to update the email elsewhere.  

Outside the scope of this branch, but I think we should just put a yellow bootstrap alert on top of the account_overview and account_details pages stating that their email doesn't match and they might want to update it, then delete the old mechanism from the workflow.

microsetta_private_api/example/client.yaml
dhakim87: I couldn't tell if this was supposed to be tagged "Sources" or "Source", I thought I saw a lot of other plural tags becoming singular.  Either way is fine with me

microsetta_private_api/example/client_impl.py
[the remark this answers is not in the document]
wasade: @dhakim87 do you feel strongly about this, or should it be deferred?

microsetta_private_api/example/client_impl.py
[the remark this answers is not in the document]
AmandaBirmingham: Punted to issue #209 Require that each account have at least one source

microsetta_private_api/example/client_impl.py
AmandaBirmingham: Added docstring and renamed function to hopefully be more self-explanatory ...

PR #174, opened 2020-06-02
Allow for tracking which kit ID an account was created using

wasade:
In brief: modify account table, recover kit IDs from old accounts, and modify the model/repo code to allow for getting a kit name in and out from the `accounts` table.

### DIFF
--- a/microsetta_private_api/api/implementation.py
+++ b/microsetta_private_api/api/implementation.py
@@ -158,7 +158,6 @@ def update_account(account_id, body, token_info):
             body['address']['post_code'],
             body['address']['country_code']
         )
-        acc.created_with_kit_name = body.get('kit_name')
 
         # 422 handling is done inside acct_repo
         acct_repo.update_account(acc)
--- a/microsetta_private_api/api/implementation.py
+++ b/microsetta_private_api/api/implementation.py
@@ -70,6 +70,7 @@ def find_accounts_for_login(token_info):
 
         if acct is None:
             return jsonify([]), 200
+
         return jsonify([acct.to_api()]), 200
 
 
@@ -157,6 +158,7 @@ def update_account(account_id, body, token_info):
             body['address']['post_code'],
             body['address']['country_code']
         )
+        acc.created_with_kit_name = body.get('kit_name')
 
         # 422 handling is done inside acct_repo
         acct_repo.update_account(acc)
@@ -687,7 +689,6 @@ def _validate_account_access(token_info, account_id):
         token_associated_account = account_repo.find_linked_account(
             token_info['iss'],
             token_info['sub'])
-
         account = account_repo.get_account(account_id)
         if account is None:
             raise NotFound(ACCT_NOT_FOUND_MSG)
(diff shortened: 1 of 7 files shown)


microsetta_private_api/api/implementation.py
```diff
@@ -157,6 +158,7 @@ def update_account(account_id, body, token_info):
             body['address']['post_code'],
             body['address']['country_code']
         )
+        acc.created_with_kit_name = body.get('kit_name')
```
dhakim87, line 161: This doesn't make sense to me.  If the purpose of this field is to track what was sent when the user created the kit, we shouldn't allow them to edit the field later when they update account information.  Unless you want this to be possible for admins for some reason (in which case it should validate admin permission)?

microsetta_private_api/repo/account_repo.py
```diff
@@ -14,11 +14,11 @@ def __init__(self, transaction):
                 "account_type, auth_issuer, auth_sub, " \
                 "first_name, last_name, " \
                 "street, city, state, post_code, country_code, " \
-                "creation_time, update_time"
+                "creation_time, update_time, created_with_kit_id"
```
dhakim87, line 17: The column ordering is a little bit inconsistent in this file.  In the reads it follows update time, in the writes it follows last name.  I'd prefer it immediately followed country_code (and preceded creation_time, update_time if applicable) in all reads and writes in the file.

microsetta_private_api/repo/account_repo.py
```diff
@@ -14,11 +14,11 @@ def __init__(self, transaction):
...
 
     write_cols = "id, email, " \
                  "account_type, auth_issuer, auth_sub, " \
-                 "first_name, last_name, " \
+                 "first_name, last_name, created_with_kit_id, " \
```
dhakim87, line 21: Ideally would follow country code

microsetta_private_api/repo/account_repo.py
```diff
@@ -41,14 +41,15 @@ def _row_to_account(r):
...
+            r['created_with_kit_id'],
             r['creation_time'], r['update_time'])
 
     @staticmethod
     def _account_to_row(a):
         return (a.id, a.email,
                 a.account_type, a.auth_issuer, a.auth_sub,
-                a.first_name, a.last_name) + \
-               AccountRepo._addr_to_row(a.address)
+                a.first_name, a.last_name, a.created_with_kit_id) + \
```
dhakim87, line 51: Ideally would follow country code

microsetta_private_api/repo/account_repo.py
```diff
@@ -146,6 +146,7 @@ def update_account(self, account):
                             "auth_sub = %s, "
                             "first_name = %s, "
                             "last_name = %s, "
+                            "created_with_kit_id = %s, "
```
dhakim87, line 149: If you decide to make my requested change in ordering of columns, this line has to be shifted.

microsetta_private_api/api/implementation.py
wasade: Makes sense, will remove from update

microsetta_private_api/repo/account_repo.py
wasade: Okay, ill take a look

PR #182, opened 2020-06-03
St dh update account info

dhakim87:
Gives users the ability to edit their account information from the account page.


microsetta_private_api/example/client_impl.py
```diff
@@ -473,15 +473,63 @@ def get_account(account_id):
...
+                           account=account)
+
+
+def post_account_info(account_id):
+    next_state, current_state = determine_workflow_state()
+    if next_state != ALL_DONE:
+        return redirect(WORKFLOW_URL)
+
+    acct = {
+        'email': flask.request.form['email'],
```
wasade, line 509: do we need to verify this corresponds to their authrocket profile?

microsetta_private_api/templates/update_account.jinja2
```diff
@@ -0,0 +1,135 @@
+{% extends "sitebase.jinja2" %}
+{% set page_title = "Account" %}
+{% set show_breadcrumbs = True %}
+{% block head %}
+    <script src="/static/vendor/js/jquery-3.4.1.min.js"></script>
+    <script src="/static/vendor/js/jquery.validate.min.js"></script>
+    <script>
+        // Wait for the DOM to be ready
+        $(function() {
```
wasade, line 9: This looks largely replicated from `creaete_acct.jinja2`, can the JS be centralized?

microsetta_private_api/templates/update_account.jinja2
```diff
@@ -0,0 +1,135 @@
...
+            });
+        });
+    </script>
+{% endblock %}
+{% block breadcrumb %}
+    <li class="breadcrumb-item"><a href="/accounts/{{account.account_id}}">Account</a></li>
+    <li class="breadcrumb-item active" aria-current="page">Account Information</li>
+{% endblock %}
+{% block content %}
+        <div class="container">
```
wasade, line 39: Same with much of the content here -- it is largely similar with `create_acct.jinja2`. Would it make sense to have this template centralized and use `{% include ... %}` to place here?

microsetta_private_api/templates/update_account.jinja2
```diff
@@ -0,0 +1,135 @@
...
+                        <input id="city" name="city" class="form-control" value="{{account.address.city |e}}" type="text" />
+                    </div>
+                    <div class="form-group">
+                        <p>
+                        <strong>Note to non-US users:</strong> We are working on support for additional countries and
+                        expect to have that available soon.
+                        </p>
+                        <label for="state" name="state_label">State/Estado</label>
+                        <select id="state" name="state" class="form-control">
+                            <option value="CA"{% if account.address.state == 'CA' %} selected {% endif %}>California</option>
```
wasade, line 72: For handling the `selected` situation, we may want to fall back on a small piece of JS. One possibly simple mechanism is [here](https://stackoverflow.com/a/21419051).

microsetta_private_api/example/client_impl.py
```diff
@@ -473,15 +473,63 @@ def get_account(account_id):
...
 
+    do_return, account, _ = ApiRequest.get('/accounts/%s' % account_id)
+    if do_return:
+        return account
+
     do_return, sources, _ = ApiRequest.get('/accounts/%s/sources' % account_id)
     if do_return:
         return sources
 
     return render_template('account.jinja2',
```
AmandaBirmingham, line 484: I am sort of struggling with our template names ... I find it confusing that I wouldn't go to the "account.jinja2" template to look at the details of my account, but instead go to the "update_account.jinja2" template. ... Any chance we could make these something like "account_overview.jinja2" and "account_details.jinja2" or something? If we went that way, might be nice to change the new endpoint name from /accounts/{account_id}/info to /accounts/{account_id}/details, etc ...

microsetta_private_api/templates/update_account.jinja2
AmandaBirmingham: second @wasade's thought

microsetta_private_api/example/client_impl.py
```diff
@@ -473,15 +473,63 @@ def get_account(account_id):
...
+
+
+def post_account_info(account_id):
+    next_state, current_state = determine_workflow_state()
+    if next_state != ALL_DONE:
+        return redirect(WORKFLOW_URL)
+
+    acct = {
+        'email': flask.request.form['email'],
+        'first_name': flask.request.form['first_name'],
```
AmandaBirmingham, line 510: I would strongly prefer to avoid pulling this information from `flask.request.form` here and pulling it from `body` in `post_create_account`.  Also, I would like to use `ACCT_FNAME_KEY`, `ACCT_LNAME_KEY`, etc here as is done in that function.

microsetta_private_api/templates/update_account.jinja2
dhakim87: I split out the definition of US states, but this data still has to come from somewhere.  We should address this when we expand to support more countries

microsetta_private_api/example/client_impl.py
dhakim87: Now renamed account_overview and account_details at /details as you specified.

microsetta_private_api/example/client_impl.py
dhakim87: Done

microsetta_private_api/example/client_impl.py
dhakim87: User has to do some hackery to change the incoming email field.  In both create and update workflows, the field is either hidden or readonly.  As such, the only way the two differ is if the user is trying to bypass our input validation, in which case they can just bypass the minimal interface altogether.  We can potentially put this validation on the API side, but I don't think putting it into the client_impl makes sense.  

I believe there are potentially valid states where the two emails differ due to the way AuthRocket federates multiple email address sources, so I think this may be a can of worms.

microsetta_private_api/example/client_impl.py
```diff
@@ -246,8 +246,23 @@ def get_workflow_create_account():
         return redirect(WORKFLOW_URL)
 
     email, _ = parse_jwt(session[TOKEN_KEY_NAME])
-    return render_template('create_acct.jinja2',
-                           authorized_email=email)
+    default_account_values = {
+            'email': email,
+            'first_name': '',
```
wasade, line 251: Can these entries also use the standard variables (e.g. `ACCT_FNAME_KEY`)?

microsetta_private_api/example/client_impl.py
```diff
@@ -473,15 +488,64 @@ def get_account(account_id):
...
+    next_state, current_state = determine_workflow_state()
+    if next_state != ALL_DONE:
+        return redirect(WORKFLOW_URL)
+
+    acct = {
+        ACCT_FNAME_KEY: body['first_name'],
+        ACCT_LNAME_KEY: body['last_name'],
+        ACCT_EMAIL_KEY: body['email'],
+        ACCT_ADDR_KEY: {
+            "street": body['street'],
```
wasade, line 528: We've centralized things like `ACCT_FNAME_KEY`, can we (should we) also do that for the address keys?

AmandaBirmingham: @wasade I'd definitely like that but I would vote *against* doing that in this PR.  

My reason for that is these string literals are in fact defined by input field names in the jinja2 template, so if we are *really* centralizing them correctly, we need to define each field name constant somewhere in python, then pass each constant into the templates that use it and put it into the html--and relevant javascript!--with `{{ }}`  syntax so that it is guaranteed the same everywhere it is used. That kind of change makes the templates harder to read and maintain, so I'd want do to it thoughtfully.

microsetta_private_api/example/client_impl.py
```diff
@@ -526,7 +541,8 @@ def post_account_info(account_id):
     if do_return:
         return sample_output
 
-    # Do we want this to return to the account info page or the account page?
+    # Do we want this to return to the account details page or the
+    # account overview page?
```
AmandaBirmingham, line 553: I vote account overview (which is already a sink, so it will be easy to integrate with the changes eventually coming in from the WIP PR :)

microsetta_private_api/templates/account_details.jinja2
```diff
@@ -0,0 +1,163 @@
...
+                        <label for="kit_name" name="kit_name_label">Kit ID/ID del kit</label>
+                        <input id="kit_name" name="kit_name" class="form-control" type="text"/>
+                    </div>
+                    {% endif %}
+                    <input id="country_code" name="country_code" type="hidden" value="{{account.address.country_code |e}}"/>
+                    {% if CREATE_ACCT %}
+                    <button type="submit" class="btn btn-primary">Create Account</button>
+                    {% else %}
+                    <button type="submit" class="btn btn-primary">Save</button>
+                    {% endif %}
```
AmandaBirmingham, line 158: Would we possibly consider having the button text just say "Submit" in both cases?  I doubt that would confuse users, since that word is common in web forms, and it would let us drop one of these fiddly "if create_account ... else" checks ...

microsetta_private_api/templates/account_details.jinja2
```diff
@@ -0,0 +1,163 @@
+{% extends "sitebase.jinja2" %}
```
AmandaBirmingham, line 1: +1 really happy to see some of the existing pages migrate over to using the sitebase :D

microsetta_private_api/example/client_impl.py
AmandaBirmingham: @wasade  please see my comment about why I'd prefer we didn't do that centralization in *this* PR

microsetta_private_api/example/client_impl.py
wasade: @dhakim87 got the commit in before this comment came up :/ given these issues, should it be reverted?

microsetta_private_api/example/client_impl.py
dhakim87: Great, done.

microsetta_private_api/templates/account_details.jinja2
dhakim87: I found two blog posts telling me to use "action verbs".  Agree that it's fiddly in jinja2 the way I did it, but I think the user experience is better

microsetta_private_api/example/client_impl.py
dhakim87: I kinda already made the change to keep these as constants at the top of client_impl before I saw this comment.  I honestly find the use of string constants for these to be less readable than the constant strings themselves.  To me it feels the same as when people write 
`5 * SECONDS_TO_MILLISECONDS`
rather than 
`5 * 1000`

The advantage I see to this style is the ability to command-B to trace references in PyCharm, but I could almost as easily Find in Project.

I think extending it down into the jinja2 template is probably unnecessary, it should be known that anything that touches an html form has to also make changes to the get/post it is submitted to.  I think renaming the fields will be less frequent than adding new ones, which will still require modifications in two places even with the constants centralized.

microsetta_private_api/templates/account_details.jinja2
AmandaBirmingham: "Submit" is an action verb :D ... but I can let this go ;)

Issue #141, opened 2020-05-05
Clear session on logout


AmandaBirmingham:
https://github.com/biocore/microsetta-private-api/pull/119#discussion_r414819189

Replace per-key `del` statements with `session.clear()`

PR #183, opened 2020-06-03
St dh admin toolbar

dhakim87:
Adds administrator functionality to the minimalInterface.  For the most part, this is just a new home page and navbar that allows searching accounts by email address.  But in order to get to that point, many of our pages had to be updated to extend sitebase.


dhakim87:
Note: Don't merge this until the user account info editing goes in, otherwise parts of the diff won't make much sense since it was spun out of that branch.

microsetta_private_api/example/client.yaml
```diff
@@ -450,6 +487,25 @@ paths:
...
+          description: Email to search for
+          schema:
+            type: string
+      responses:
+        '200':
+          description: List of identified accounts
+          content:
+            text/html:
+              schema:
+                type: string
```
AmandaBirmingham, line 508: Add an empty line here to match the empty line added before "components" in private api yaml?

microsetta_private_api/example/client_impl.py
```diff
@@ -14,6 +14,10 @@
 
 import flask
 from flask import render_template, session, redirect
+# TODO FIXME HACK:  It is convenient to expose some admin functionality,
+#  but requires knowing whether or not the current user is an admin, which is
+#  not currently part of our api.  This import couples client and admin modules
```
AmandaBirmingham, line 19: Do we have an issue created to make this part of our api? This  coupling seems currently necessary but like something we would like to be able to get rid of in the future?

microsetta_private_api/example/client_impl.py
```diff
@@ -119,13 +133,16 @@ def home():
...
+    decoded = jwt.decode(token, PUB_KEY, algorithms=['RS256'], verify=True)
+    session[ADMIN_MODE_KEY] = token_grants_admin_access(decoded)
     return redirect("/home")
 
 
 def logout():
     if TOKEN_KEY_NAME in session:
         # delete these keys if they are here, otherwise ignore
         session.pop(TOKEN_KEY_NAME, None)
+        session.pop(ADMIN_MODE_KEY, None)
```
AmandaBirmingham, line 145: Is now the time to change this over to session.clear()? (Issue #141 )

microsetta_private_api/example/client.yaml
```diff
@@ -450,6 +487,25 @@ paths:
             text/html:
               schema:
                 type: string
+
+  '/admin/account_search':
+    get:
+      operationId: microsetta_private_api.example.client_impl.interactive_account_search
```
AmandaBirmingham, line 493: Can we name the client_impl function get_interactive_account_search?

microsetta_private_api/example/client_impl.py
```diff
@@ -833,6 +910,22 @@ def render_signup_intermediate():
...
 
 
+# Administrator Mode Functionality
+def interactive_account_search(email_query):
+    do_return, email_diagnostics, _ = ApiRequest.get(
+        '/admin/search/account/%s' % (email_query,))
+    if do_return:
+        return email_diagnostics
+
+    accounts = [{"email": acct['email'], "account_id": acct['id']}
```
AmandaBirmingham, line 920: I'm not entirely sold on the necessity of doing this ... jinja templates can take objects, so perhaps we could just pass in the `email_diagnostics['accounts']` value to the template and access the fields in it that are needed directly from the template?

microsetta_private_api/example/client_impl.py
dhakim87: Yup, see my other PR 😉

microsetta_private_api/example/client.yaml
dhakim87: Done

microsetta_private_api/example/client.yaml
dhakim87: Done

microsetta_private_api/example/client_impl.py
dhakim87: Yeah...  But the admin diagnostics objects are enormous and don't have a schema specified.  I can push the diagnostics object to jinja if you really want, but I was worried I would stupidly leave a comment like `<!-- {{account}} -->` in the jinja2 and leave web pages showing commented out admin diagnostics in them.  I just found it easier to figure out what fields I actually needed on the python side than the jinja side.

microsetta_private_api/example/client_impl.py
dhakim87: Ooh, a GET to /api/accounts already puts the account_type in, so I was able to remove the import and check that.  Even if that account_type field in the response was inexplicably wrong, all the admin operations exposed by the client are still guarded by the token, so this should be safe.

microsetta_private_api/templates/sample.jinja2
```diff
@@ -1,8 +1,9 @@
...
-<html lang="en">
-<head>
-    <meta charset="utf-8"/>
-    <title>Microsetta Sample Information</title>
+{% extends "sitebase.jinja2" %}
+{% set page_title = "Sample Information" %}
+{% set show_breadcrumbs = True %}
+{% set disable_bootstrap = True %}
+{% block head %}
+    <!--  Argh, Bootstrap 3 and 4 don't mix.  So the header and footer look slightly different on this page -->
```
wasade, line 6: This is resolved in #202 as bootstrap 3 is no longer needed

microsetta_private_api/templates/sample.jinja2
```diff
@@ -14,40 +15,27 @@
...
+              <form method="post" id="sample_form">
+                <vue-form-generator :schema="schema" :model="model" :options="formOptions"></vue-form-generator>
+              </form>
             </div>
+          </div>
         </div>
     </div>
 
+    <!-- These should really be in the head, but don't easily work there.
+    Probably needs a document.load handler or something?  -->
```
wasade, line 38: Same, should be resolved in #202

microsetta_private_api/templates/sitebase.jinja2
```diff
@@ -4,13 +4,40 @@
     <meta charset="utf-8"/>
     <title>Microsetta {{page_title}}</title>
 
-{%block head%}
-{%endblock%}
-
+    <!-- jquery must be enabled before bootstrap for collapsible functionality -->
+    {% if enable_jquery %}
+    <script src="/static/vendor/js/jquery-3.4.1.min.js"></script>
```
wasade, line 9: ...we're using multiple versions of jquery. that's awkward, I'm going to see if I can clean that up...

PR #197, opened 2020-06-08
Open FAQ in new tab

### LINKED ISSUE #187: Have FAQ open in a new tab
This would avoid the situation where a user looses all their survey responses if they click on the FAQ link part of the way through

dhakim87:
Fix #187 
187 FAQs walk into a bar and the bartender says:
Each of you has to open your own tab.

Sets the <a> target property to "_blank" causing user browsers to open new tabs.  For a full list of options, see https://www.w3schools.com/tags/att_a_target.asp

This was changed in sitebase and all other jinja2 files that contain FAQ links.  Many of these should be migrated to just use sitebase.


wasade:
😆 

...ya, we really should migrate to sitebase but I think out of scope of this PR

PR #202, opened 2020-06-09
Fixes #178

### LINKED ISSUE #178: The calendar is confusing
Users are reporting that it's not obvious they can click on the calendar, and if modifying the text field manually they are getting a date in 1969. The behavior is contributing to user frustration.

wasade:
fixes #178 

...also put in some additional changes to the UI page which I hope make it a little nicer. 

Note: the different js and css files are not represented locally. I can pull to static if that would be useful. 

I don't know enough about jQuery to know if this matters, but the version we're using is rather old. 

NOTE: I just noticed that if a sample site is missing AND a date or a time is missing, it's necessary to click "Update" twice. I don't know why yet, but that needs to be fixed before merge

*IMPORTANT*: I cleaned up and standardized the JS libraries used. There was a mix of versions. I did not notice anything odd on the pages or on the javascript console so I think all is good.  

datetime


wasade:
Okay, the validation piece is fixed

datetime_validation

microsetta_private_api/example/client_impl.py
```diff
@@ -778,6 +779,12 @@ def put_sample(account_id, source_id, sample_id):
     for x in flask.request.form:
         model[x] = flask.request.form[x]
```
dhakim87, line 780: I wonder, if we added a parameter called "body" to this function, would it be automatically populated with the same contents as this model object we're filling in here?  That's how we do it everywhere else, but I know this one uses slightly different mechanisms in the yaml...

microsetta_private_api/templates/sample.jinja2
```diff
@@ -3,19 +3,70 @@
 <head>
```
dhakim87, line 3: This file was updated in master to use sitebase.  Do you want to use sitebase here and ensure your jquery libraries are updated in that file, or push in the PR with the changes as is and we'll do the sitebase extension again as a separate PR?

dhakim87:
Actually, don't merge this yet.  I tried testing locally and something might be wrong?  Any time I add notes, they just disappear when I save and load.  Suspect one of the form parameter names is wrong or something?  

Is there also a setting to make the sample notes multi line?

microsetta_private_api/templates/sample.jinja2
```diff
@@ -36,34 +87,59 @@
...
+                  </select>
+                </div>
+                <label for="sample_site" generated="true" class="error"></label>
+              </div>
+              {% endif %}
+
+              <div class="form-group">
+                <label for="sample_notes" name="sample_notes_label">Notes</label>
+                <div class="col-sm-10">
+                    <input id="sample_notes" name="sample_notes" class="form-control" value="{{ sample_notes |e}}" autocomplete="off" type="text" />
```
dhakim87, line 134: I think this should say: 
value="{{ sample.sample_notes |e }}"
That's why notes is always blank to start with.

microsetta_private_api/example/client_impl.py
wasade: I'm really not sure...

microsetta_private_api/templates/sample.jinja2
wasade: I'll resolve here, thanks

microsetta_private_api/templates/sample.jinja2
wasade: Ah! good catch

wasade:
@dhakim87 I think this is good now. Good catch on the bug. And using a `textarea` now for notes too

microsetta_private_api/templates/sitebase.jinja2
```diff
@@ -7,10 +7,8 @@
     <!-- jquery must be enabled before bootstrap for collapsible functionality -->
     <script src="/static/vendor/js/jquery-3.4.1.min.js"></script>
     <script src="/static/vendor/js/jquery.validate.min.js"></script>
-    {% if not disable_bootstrap %}
```
dhakim87, line 10: Yaaay!

PR #203, opened 2020-06-09
Fixes #180

### LINKED ISSUE #180: Suggest the kit ID when a user needs to enter one to claim samples
After selecting a source, a user needs to then enter a kit ID to claim samples from. In most situations, the kit ID they'll use is the one they created their account with. Once #174 is merged, we should be able to suggest a reasonable kit ID to use here so they don't have to re-type it

wasade:
Provides a suggested kit ID based on the one the account was created with. Fixes #180 

kit_hint

### DIFF
--- a/microsetta_private_api/templates/source.jinja2
+++ b/microsetta_private_api/templates/source.jinja2
@@ -142,8 +142,13 @@
         <div class="container">
             <form method="get" name="list_kit_form" action="/list_kit_samples">
                 <div class="form-group">
+                    {% if claim_kit_name_hint is not none %}
+                    <label for="kit_name" name="kit_name_label">The kit ID used to create your account has unclaimed samples. To claim samples from this kit, please click "List samples". To claim samples from another kit ID, please enter it here and then click "List Samples":</label>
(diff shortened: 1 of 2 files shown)


microsetta_private_api/templates/source.jinja2
```diff
@@ -142,8 +142,13 @@
         <div class="container">
             <form method="get" name="list_kit_form" action="/list_kit_samples">
                 <div class="form-group">
+                    {% if claim_kit_name_hint is not none %}
+                    <label for="kit_name" name="kit_name_label">The kit ID used to create your account has unclaimed samples. To claim samples from this kit, please click "List samples". To claim samples from another kit ID, please enter it here and then click "List Samples":</label>
+                    <input id="kit_name" name="kit_name" class="form-control" type="text" value="{{ claim_kit_name_hint }}"/>
```
dhakim87, line 147: Not sure you if you tried this and prefer the current approach, but any reason to set the value rather than using it as a hint/placeholder?  
https://www.w3schools.com/tags/att_input_placeholder.asp

wasade: I thought about that, but I also wanted to give people the option to just hit enter. The placeholder value isn't sent over the wire :/

PR #207, opened 2020-06-10
St dh issue 144

### LINKED ISSUE #144: Pass variable breadcrumb options to sitebase.jinja2
https://github.com/biocore/microsetta-private-api/pull/143#discussion_r422311940

dhakim87:
All jinja2 templates now extend sitebase.  I've looked at all templates to ensure they correctly display breadcrumb as necessary - Several specialized templates (FAQ, Error Page, ...) by design will not show breadcrumbs or will not have specific breadcrumb links for their own pages.  All templates corresponding to the viewing of objects within our api will show full breadcrumbs.  

Fix #144


microsetta_private_api/templates/home.jinja2
```diff
@@ -1,87 +1,56 @@
...
-    <div class="content">
-        <nav aria-label="breadcrumb">
-          <ol class="breadcrumb">
-            <li class="breadcrumb-item active" aria-current="page">Home</li>
-          </ol>
-        </nav>
-
-        {% if user %}
+{% block content %}
+{% if user %}
```
wasade, line 10: Should (can?) this be `if user is not none`?

microsetta_private_api/templates/home.jinja2
```diff
@@ -1,87 +1,56 @@
...
-                {% else %}
-                <a href="/workflow" class="list-group-item list-group-item-action">
-                    Click here to create a new account...
-                </a>
-                {% endif %}
-            </div>
+    </div>
+    {% if email_verified %}
+    <div class="list-group">
+        {% if acct_id %}
```
wasade, line 16: same here with `is not none`

microsetta_private_api/templates/sitebase.jinja2
```diff
@@ -51,7 +51,11 @@
     {% if show_breadcrumbs %}
         <nav aria-label="breadcrumb">
           <ol class="breadcrumb">
+            {% if not is_home %}
```
wasade, line 54: minor, but possible to do the uninverted form? eg:

```
{% if is_home %}
<li class="breadcrumb-item active" aria-current="page">Home</li>
{% else %}
<li class="breadcrumb-item"><a href="/home">Home</a></li>
{% endif %}
```

microsetta_private_api/templates/source.jinja2
```diff
@@ -57,120 +57,113 @@
...
-                        </div>
-                    </a>
-                    {% endfor %}
-                {% endif %}
-            </div>
+            </form>
         </div>
     </div>
+    <p></p>
+    <hr style="border-top: dotted 1px"/>
```
wasade, line 144: 👍

microsetta_private_api/templates/source.jinja2
```diff
@@ -57,120 +57,113 @@
...
+    </div>
+    {% endif %}  <!-- end of if samples|length > 0 -->
+    {% if samples|length > 1 and is_human %}
+        <br>
+        <u><i>Do I need to take more than one Food Frequency Questionnaire (FFQ)</u></i>?
+If you are taking all of your samples at the same time, you will only need to complete one. If you are collecting samples over time (e.g., once per week), then please consider taking an FFQ per time point. The FFQs provide valuable dietary research data for the project, so researchers can better understand the relationship between the food you eat and your microbes.
+    {% endif %}  <!-- end of if samples|length > 1 and is_human -->
+
+    {% if samples|length > 0 %}
+    <hr style="border-top: dotted 1px"/>
```
wasade, line 120: :+1:

wasade: This might be out of scope, but any chance these separators could be added into `account_overview.jinja2` as well?

wasade:
And actually, @dhakim87 I haven't pulled this down to test locally due to the conflicts. Would you prefer I pull down and explore this directly?

microsetta_private_api/templates/source.jinja2
dhakim87: I think that's where I took them from?  There's one in between Account Details and the Sources list, did you want another one somewhere else?

wasade: ah okay, sorry, i couldn't remember if that had gone in

microsetta_private_api/templates/home.jinja2
dhakim87: It's not quite python and it's not quite javascript.  Jinja2 is very confusing.  

In the case where {{ user }} displays as None:

"{% if user is not none %}" Works identically to "{% if user %}"
"{% if user is not None %}" crashes ("No test named None")
"{% if user is not null %}" crashes ("No test named null")
"{% if user != none %}"  Works identically
"{% if user != null %}" Doesn't crash, produces a wrong result
"{% if user != None %}" Works identically

So it -can- be "if user is not none", but I don't like making that change very much since it sounds like none is a "test" that it runs that I really don't understand...

microsetta_private_api/templates/sitebase.jinja2
dhakim87: Sure

dhakim87:
Up to you, conflicts are resolved for the moment.  Everything should look more or less the same except for the source page, and sporadic indentation issues from the container divs so you can check out whether or not you like the layout changes.

microsetta_private_api/templates/home.jinja2
wasade: The jinja2 [docs](https://jinja.palletsprojects.com/en/2.11.x/templates/) dont seem to explicitly state it from searching around, but I found this note: "The special constants true, false, and none are indeed lowercase. Because that caused confusion in the past, (True used to expand to an undefined variable that was considered false), all three can now also be written in title case (True, False, and None). However, for consistency, (all Jinja identifiers are lowercase) you should use the lowercase versions." My impression is `is not none` is more inline with jinja2 norms. 

In python, `is` and `==` are different. Former tests identity, latter tests value. I'm not sure of the jinja2 semantics.

I'm not going to fight this one though -- if you prefer `if user` then let's move fwd!

PR #208, opened 2020-06-10
No spaces in kit_name

### LINKED ISSUE #190: Strip contained, leading and trailing whitespace in kit IDs
Users are accidentally putting spaces before and within the kit IDs causing confusion. It would be nice to remove this for them in the field, and before the request is made to the API

dhakim87:
Fix #190 

Added a jquery event handler to kit_name input in source and create_acct templates.  Any time the user modifies the kit_name field, all whitespace will be removed.


microsetta_private_api/templates/source.jinja2
```diff
@@ -53,9 +53,17 @@
             });
         };
 
+
         // Wait for the DOM to be ready
 
         $(document).ready(function(){
+            <!--https://stackoverflow.com/questions/12010275/strip-white-spaces-on-input-->
+            $('#kit_name').bind('input', function(){
```
wasade, line 61: As this is used twice, would it be possible to centralize the logic? Placing something like

```javascript
function sanitize_kit_name(unused, value) {
    return value.replace(/\s+/g, '');
}
```

...into `static/kit_check.js`?

dhakim87: I'm actually super confused by the status of kit_check.js.  I could have sworn it was deleted in the state rework PR and moved into another file.  Are we going to have 1 copy or 2 copies of kit name entry in the final version?  

Regardless, I'll make a javascript util we can start using for setting common jquery event handlers.

wasade: I think it is removed by the state rework PR, but that hasn't been merged yet. Thanks for the util for common handlers!

wasade:
Conflict hell... :(

microsetta_private_api/templates/sitebase.jinja2
```diff
@@ -14,6 +14,10 @@
     {%block head%}
     {%endblock%}
 
+    <link rel="stylesheet" href="/static/vendor/bootstrap-4.4.1-dist/css/bootstrap.min.css" />
```
wasade, line 17: this line and next look like duplicates of lines 11 and 12?

dhakim87: Ugh.  I need to see if Pycharm has some ignore whitespace options on its merges, the jinja2 files are a nightmare to merge if I have to change indentation in a block.

dhakim87: We could also look into purchasing diff tools like BeyondCompare 😄

Issue #192, opened 2020-06-06
Mechanism to notify users of maintenace


wasade:
It would be helpful to provide users notice of the site going down to facilitate deploys

PR #214, opened 2020-06-13
Admin Panel For Systemwide Messaging

dhakim87:
Adds a new administrator panel for sending messages user wide.  Can be used to notify users of maintenance, solving part of #192

### DIFF
--- a/microsetta_private_api/example/client_impl.py
+++ b/microsetta_private_api/example/client_impl.py
@@ -59,6 +59,13 @@ SOURCE_PREREQS_MET = "SourcePrereqsMet"
 #  in some way, as well as any special handling for external surveys.
 VIOSCREEN_ID = 10001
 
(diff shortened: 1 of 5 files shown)


microsetta_private_api/example/client_impl.py
```diff
@@ -834,6 +859,8 @@ def get_source(account_id, source_id):
     is_human = source_output['source_type'] == Source.SOURCE_TYPE_HUMAN
     return render_template('source.jinja2',
                            admin_mode=session.get(ADMIN_MODE_KEY, False),
+                           system_msg_text=SYSTEM_MSG,
```
wasade, line 862: All this makes me wonder if `render_template` should be a `functools.partial`, or if flask has a mechanism to intercept `render_template` calls?  (Not blocking here, but if you agree we should make an issue)

PR #216, opened 2020-06-15
Dissociate sample

### LINKED ISSUE #181: Allow a user to disassociate a sample from a source
If a user accidentally associates a sample to a source, they do not have a means to disassociate. A disassociation should only be allowed if sample details are not present.

AmandaBirmingham:
Fixes #181 Allow a user to disassociate a sample from a source 

Based on the manual testing I have done, this seems to work just fine. However, I have listed this PR as WIP because it required an expansion to the private api (not just the minimal interface client) and I haven't yet added unit tests to officially verify the api change.

** New!  Now with unit tests! **


README.md
```diff
@@ -16,10 +16,14 @@ Install connexion version 2.0 (which supports the OpenAPI Specification 3.0) as
...
 `pip install connexion[swagger-ui]`
 
-Also install the JSON Web Tokens library with cryptography support:
+Install the JSON Web Tokens library with cryptography support:
 
 `pip install pyjwt[crypto]`
 
+Install redis:
+
+`sudo pip install redis`
```
dhakim87, line 25: oops.  Good call.

microsetta_private_api/api/implementation.py
```diff
@@ -528,9 +528,19 @@ def dissociate_sample(account_id, source_id, sample_id, token_info):
...
 
     with Transaction() as t:
+        answers_repo = SurveyAnswersRepo(t)
+        answered_survey_ids = answers_repo.list_answered_surveys_by_sample(
+            account_id, source_id, sample_id)
+
+        for curr_answered_survey_id in answered_survey_ids:
+            answers_repo.dissociate_answered_survey_from_sample(
+                account_id, source_id, sample_id, curr_answered_survey_id)
+
```
dhakim87, line 538: I might prefer that the dissociation of answered surveys is moved into SampleRepo.dissociate_sample.  My understanding is that this would need to occur every time a sample is dissociated, and there are other fields in the database (sample information) that are already wiped, so I don't see a reason to keep dissociation of answered surveys as a separate operation in the implementation.py file.  

Thoughts?

microsetta_private_api/example/client.yaml
```diff
@@ -385,6 +385,23 @@ paths:
               schema:
                 type: string
 
+  '/accounts/{account_id}/sources/{source_id}/samples/{sample_id}/remove':
+    post:
+      operationId: microsetta_private_api.example.client_impl.post_remove_sample_from_source
```
dhakim87, line 390: Any reason to make this a POST rather than a DELETE the way it is in the api layer?  Is this so we can do it in response to a form or something?

microsetta_private_api/repo/survey_answers_repo.py
```diff
@@ -378,3 +383,17 @@ def _unlocalize(self, answer, language_tag):
                 raise BadRequest("Invalid survey response: %s" % answer)
             normalized_answer = row[0]
             return normalized_answer
+
+    def _get_survey_sample_associations(self, answered_survey_id):
+        """ Do not use from api layer, you must validate account and source."""
```
dhakim87, line 388: Who calls this function?  I didn't see it used in implementation (which is good), but I also didn't see it used in the SampleRepo or anywhere else...  (I wish github diff had a jump to usages :D)

dhakim87: Ahhh, it enables the test code, nm.

README.md
AmandaBirmingham: Hmm, I admit I didn't try it *without* sudo.  I tried installing it with conda, which was a miserable failure ... even though the conda environment claimed it was there, the code couldn't see it.  So I looked up "installing redis", found instructions that said "use this line", used it, and it worked.   The thing I read had some reason why sudo was included but I didn't pay much attention once it worked.

I can take the sudo out, but can't claim to have verified that the redis install works without it.  Are you confident it does?

microsetta_private_api/example/client.yaml
AmandaBirmingham: That is a very good question, and there actually is an answer :) The answer is: this method is used by an ajax call.  Ajax can do get and post but NOT delete or put, so post is the better of two bad options.

microsetta_private_api/api/implementation.py
AmandaBirmingham: The SurveyAnswersRepo already depends on the SampleRepo.  If I move this functionality (which depends on  into SurveyAnswersRepo) SampleRepo.dissociate_sample, then we will have a circular dependency.  I figure that's a Bad Thing.

I suppose there is some restructuring we could do that would work around the need for a circular dependency, but I think it would require changes in 3 different modules (survey_answers_repo, implementation, and admin_repo). Given that, my inclination is to think that pushing this functionality out of implementation.py isn't worth it at this point. Opinions?

PR #225, opened 2020-06-16
Redis Backend for Global State in client

dhakim87:
Added a redis backend!  You have to install and run a redis server in order to run the server now.  

You can do that in your favorite conda environment with a 

```
conda install -c conda-forge redis
redis-server
```

We now include redis-py as a requirement for accessing that server from python.  

There's a simple wrapper for the redis interactions in microsetta_private_api/client_state/redis_cache, which gives you dictionary like semantics and a set of keys with comments on expected formats of stored objects.  

As a side effect, I went through the render_template calls and moved all the sitebase related jinja2 variables out into a wrapper function which cleans up the code and may fix some broken logout buttons.  

This should fix the issues with the admin system message banners when running with multiple gunicorn workers.

### DIFF
--- a/microsetta_private_api/client_state/redis_cache.py
+++ b/microsetta_private_api/client_state/redis_cache.py
@@ -1,0 +1,31 @@
+import json
+import redis
+
+
+class RedisCache:
+    # System Banner Format: ["Blah blah blah", "primary"]
+    SYSTEM_BANNER = "system_banner"
+
+    def __init__(self):
+        self.r = redis.Redis(host='localhost',
+                             port=6379,
+                             db=0,
+                             decode_responses=True)
+
+    def __getitem__(self, key):
+        val = self.r.get(key)
+        if val is None:
+            return None
+        return json.loads(val)
+
+    def get(self, key, default):
+        val = self.r.get(key)
+        if val is None:
+            return default
+        return json.loads(val)
+
+    def __setitem__(self, key, value):
+        self.r.set(key, json.dumps(value))
+
+    def __delitem__(self, key):
+        self.r.delete(key)
--- a/microsetta_private_api/example/client_impl.py
+++ b/microsetta_private_api/example/client_impl.py
@@ -1,5 +1,3 @@
-import json
-
 import flask
 from flask import render_template, session, redirect
 import jwt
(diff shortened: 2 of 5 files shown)


microsetta_private_api/example/client_impl.py
```diff
@@ -59,12 +59,28 @@
...
+    msg, style = client_state.get(RedisCache.SYSTEM_BANNER, (None, None))
+    defaults["system_msg_text"] = msg
+    defaults["system_msg_style"] = style
+
+    endpoint = SERVER_CONFIG["endpoint"]
+    authrocket_url = SERVER_CONFIG["authrocket_url"]
+    defaults["endpoint"] = endpoint
+    defaults["authrocket_url"] = authrocket_url
+
+    for key in context:
```
wasade, line 80: would `defaults.update(context)` work here?

microsetta_private_api/client_state/redis_cache.py
```diff
@@ -0,0 +1,31 @@
...
+        val = self.r.get(key)
+        if val is None:
+            return default
+        return json.loads(val)
+
+    def __setitem__(self, key, value):
+        self.r.set(key, json.dumps(value))
+
+    def __delitem__(self, key):
+        self.r.delete(key)
```
wasade, line 31: what happens if the key is not present?

dhakim87: Case with key present:

```
>>> import redis
>>> r = redis.Redis()
>>> r.set("x", "twenty-seven")
True
>>> r.delete("x")
1
```

Case with no key present:
```
>>> import redis
>>> r = redis.Redis()
>>> print(r.delete("No Such Key"))
0
```

Looks like it just silently ignores failing to remove that key.  I'm fine with that as our api, but if you wanted it to be more dict like, we could throw a KeyError

microsetta_private_api/example/client_impl.py
dhakim87: I swear I won't remember how to write a for loop in a couple more years :D

wasade: :)

microsetta_private_api/client_state/redis_cache.py
wasade: I think this is perfect

PR #228, opened 2020-06-19
Email support

wasade:
An object to support sending jinja2 templated emails. This is provides partial support for biocore/microsetta-admin#17

### DIFF
--- a/microsetta_private_api/util/email.py
+++ b/microsetta_private_api/util/email.py
@@ -45,7 +45,7 @@ class SendEmail:
             raise smtplib.SMTPException("Unable to connect")
 
     @classmethod
-    def send(cls, to, email_template, **email_template_args):
+    def send(cls, to, email_template, email_template_args=None, from_=None):
         """Send a message
 
         Parameters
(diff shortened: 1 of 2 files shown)


microsetta_private_api/util/email.py
```diff
@@ -0,0 +1,81 @@
...
+        return connection
+
+    @classmethod
+    def connect(cls):
+        """Connect or reconnect if a connection has timed out"""
+        def reconnect():
+            if cls.connection is None or cls.connection.noop()[0] == 421:
+                return True
+            else:
+                return False
```
dhakim87, line 34: Either this method, or connection.noop() has a really unfortunate name.  It's not clear to me how this if statement can be connecting or reconnecting via side effects from a function called no op :).  

Maybe this could be renamed to requires_reconnect() or is_active() with appropriate changes to booleans returned

dhakim87: Ahh, wait, I'm misunderstanding what the comment is attached to :D

microsetta_private_api/util/email.py
```diff
@@ -0,0 +1,80 @@
...
+    @staticmethod
+    def _connect(cls):
+        """Establish a SMTP connection"""
+        connection = smtplib.SMTP(cls.host, port=cls.port)
+        if cls.user is not None:
+            connection.login(cls.user, cls.password)
+        return connection
+
+    @classmethod
+    def connect(cls):
```
dhakim87, line 28: Is this synchronous?

microsetta_private_api/util/email.py
wasade: `.noop()` is part of `smtplib.SMTP` and issues a `NOOP` command server side. This will report a timeout if the connection has died.

microsetta_private_api/util/email.py
wasade: Ya, per slack though, I think we defer async for now

microsetta_private_api/util/email.py
```diff
@@ -0,0 +1,80 @@
...
+        html = email_template.html.render(**email_template_args or {})
+
+        first = MIMEText(plain, "plain")
+        second = MIMEText(html, "html")
+
+        message.attach(first)
+        message.attach(second)
+
+        cls.connect()
+        cls.connection.send_message(message)
```
dhakim87, line 80: Coming from a mobile development background, where connections unexpectedly drop all the freaking time, I think this pattern is slightly wrong, though I doubt we'll ever notice while the server is colocated.  If the connect call succeeds, but then the connection immediately drops out, what happens when you call send_message?

wasade: fair. happy to wrap and catch?

PR #232, opened 2020-06-23
Remove interface

### LINKED ISSUE #175: Separate interface into a separate Github repository
Given the relative stability of the API, it may simplify development and deployment to have the interface code separated into a distinct repository, something like "microsetta-interface".

wasade:
Fixes #175 
cc biocore/microsetta-interface#1


README.md
```diff
@@ -24,10 +24,6 @@ Install the JSON Web Tokens library with cryptography support:
 
 `pip install pyjwt[crypto]`
 
-Install redis:
```
dhakim87, line 27: Did this make it into the migrated README?

microsetta_private_api/api/_consent.py
```diff
@@ -13,10 +13,9 @@ def render_consent_doc(account_id, language_tag, consent_post_url, token_info):
...
     localization_info = localization.LANG_SUPPORT[language_tag]
-    consent_html = render_template(
-        "new_participant.jinja2",
-        tl=localization_info[localization.NEW_PARTICIPANT_KEY],
-        lang_tag=language_tag,
-        post_url=consent_post_url
-    )
-    return jsonify({"consent_html": consent_html}), 200
+    payload = {'lang_tag': language_tag,
+               'post_url': consent_post_url}
```
dhakim87, line 17: What?  Why are we just echoing language_tag and consent_post_url?  Do we need to update the yaml to remove consent_post_url?  Will language_tag be going into the localization_info lookup in the future?

README.md
wasade: No need, just added it to conda reqs so the user doesn't need to explicitly install it

microsetta_private_api/api/_consent.py
wasade: Good catch. I figured `lang_tag` was used in the template for rendering but it does not seem so. The `post_url` is of course used

microsetta_private_api/api/microsetta_private_api.yaml
```diff
@@ -194,9 +194,62 @@ paths:
...
+                    type: string
+                  consent_your_child:
+                    type: string
+                  assent_13_17:
+                    type: string
+                  text_i_have_read_1:
+                    type: string
+                  consent_18:
+                    type: string
+
```
dhakim87, line 252: Do the consent docs work post migration?  It looks like this collapses down the dictionary of values of the jinja2 variable "tl", but I didn't see all references updated on the other side.

wasade: yes. Which reference is missing?

wasade: it does seem that openapi (or connexion) doesn't care about case of the keys though

microsetta_private_api/api/_consent.py
dhakim87: post_url is just echoed from whatever the interface passed up to the api here, that seems strange to me.

microsetta_private_api/api/microsetta_private_api.yaml
dhakim87: I have no idea how jinja2 managed to pull the correct value for {{ post_url }} given my current read, but I'm glad that it does :D

README.md
dhakim87: Did we tell the user to start the redis server in the old readme?  I'm not sure.

wasade: nope

microsetta_private_api/api/microsetta_private_api.yaml
wasade: it didn't...

PR #233, opened 2020-06-25
St dh issue 199

### LINKED ISSUE #199: Streamline step-checking in client_impl.py
Per zoom discussion w @wasade @dhakim87 @AmandaBirmingham on 20200605:
4) Modify _route_to_closest_sink to take in account_id, source_id, and needed_template_survey_id as nullable arguments (instead of the curr_state dictionary)
5) Modify all code that calls _route_to_closest_sink to pass the relevant arguments
6) Modify _check_relevant_prereqs to incorporate currently-separate _check_survey_allowed (we didn't discuss this in our meeting but it will be required to make the changes we did discuss work :) ), including adding needed_template_survey_id as nullable argument
7) Modify _check_relevant_prereqs (and its subsidiary functions) not to fill or return curr_state
8) Rework _check_relevant_prereqs into a decorator that takes in a list of acceptable states and, if input doesn't meet one of those states, routes to the closest sink
9) Modify existing code that calls _check_relevant_prereqs or _check_survey_allowed to use decorator instead

dhakim87:
Fix #199 

Hides a lot of the complexity of checking the workflow in client_impl behind the new @prerequisite decorator.

### DIFF
--- a/microsetta_private_api/example/client_impl.py
+++ b/microsetta_private_api/example/client_impl.py
@@ -7,6 +7,8 @@ from urllib.parse import quote
 from os import path
 from datetime import datetime
 import base64
+import functools
+import inspect
 
 # Authrocket uses RS256 public keys, so you can validate anywhere and safely
 # store the key in code. Obviously using this mechanism, we'd have to push code
@@ -1048,6 +1050,65 @@ def post_system_message(body):
     client_state[RedisCache.SYSTEM_BANNER] = (text, style)
 
     return _render_with_defaults('admin_system_panel.jinja2')
+
+
+# To send arguments to a decorator, you define a factory method with those args
+# that returns a decorator.  The decorator then takes a function and returns
+# a wrapper that you want to call instead.  Thus there should be three layers.
+def get_requires(target_state):
+    def decorator(func):
+        # Since we don't know what arguments the function to wrap will take
+        # we need to figure out how to parse an account_id and source_id out
+        # of the function when it's called.  To do that we inspect the function
+        # signature and find the locations of the arguments named 'account_id'
+        # and 'source_id'.  (Yes, this will break if arguments are renamed)
+
+        # NOTE:  Rather than inspecting the function signature,
+        # we could potentially switch on the target_state value and decide how
+        # many arguments to take from the beginning of *args, this is because
+        # by our current design, if an account_id is required, it's the first
+        # argument, and if a source_id is required, it's the second argument.
+
+        # Determine whether/where wrapped function takes account_id, source_id
+        sig = inspect.signature(func)
+        param_names = sig[0]
+        acct_idx = None
+        source_idx = None
+        acct_default = None
+        source_default = None
+        if 'account_id' in param_names:
+            acct_idx = param_names.index('account_id')
+            acct_default = sig[1 + acct_idx]
+        if 'source_id' in param_names:
+            source_idx = param_names.index('source_id')
+            source_default = sig[1 + source_idx]
+
+        # Calling functools.wraps preserves information about the wrapped func
+        # so introspection/reflection can pull the original documentation even
+        # when passed the wrapper function.
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            # Parse account_id/source_id
+            acct_id = acct_default
+            src_id = source_default
+            if acct_idx is not None:
+                if acct_idx < len(args):
+                    acct_id = args[acct_idx]
+            if source_idx is not None:
+                if source_idx < len(args):
+                    src_id = args[source_idx]
+
+            # Check relevant prereqs from those arguments
+            prereqs_step, curr_state = _check_relevant_prereqs(acct_id, src_id)
+
+            # Route to closest sink if state doesn't match a required state
+            if prereqs_step != target_state:
+                return _route_to_closest_sink(prereqs_step, curr_state)
+            return func(*args, **kwargs)
+
+        return wrapper
+
+    return decorator
 
 
 class BearerAuth(AuthBase):
--- a/microsetta_private_api/example/client_impl.py
+++ b/microsetta_private_api/example/client_impl.py
@@ -234,15 +234,15 @@ def _check_relevant_prereqs(acct_id=None, source_id=None):
 # To send arguments to a decorator, you define a factory method with those args
 # that returns a decorator.  The decorator then takes a function and returns
 # a wrapper that you want to call instead.  Thus there should be three layers.
-def get_requires(target_state):
+def prerequisite(target_state):
     """
     Usage
-    @get_requires(NEEDS_EMAIL_CHECK)
+    @prerequisite(NEEDS_EMAIL_CHECK)
     def get_update_email(account_id):
         # If client is not in the NEEDS_EMAIL_CHECK state, they will be
         # redirected rather than reaching get_update_email.
 
-    @get_requires([State1, State2, State3])
+    @prerequisite([State1, State2, State3])
     def crazy_function(account_id, source_id):
         # If client is not in one of State1, State2, State3 stats, they will
         # be redirected.
@@ -535,7 +535,7 @@ def get_logout():
     return redirect(HOME_URL)
 
 
-@get_requires(NEEDS_ACCOUNT)
+@prerequisite(NEEDS_ACCOUNT)
 def get_create_account():
     email, _ = _parse_jwt(session[TOKEN_KEY_NAME])
     # TODO:  Need to support other countries
@@ -559,7 +559,7 @@ def get_create_account():
                                  account=default_account_values)
 
 
-@get_requires(NEEDS_ACCOUNT)
+@prerequisite(NEEDS_ACCOUNT)
 def post_create_account(body):
     kit_name = body[KIT_NAME_KEY]
     session[KIT_NAME_KEY] = kit_name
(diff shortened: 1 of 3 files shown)


microsetta_private_api/example/client_impl.py
```diff
@@ -211,21 +214,142 @@ def _check_source_prereqs(acct_id, source_id, current_state=None):
...
+    The wrapped function itself will never see these overridden values.
+    """
+    def decorator(func):
+        # TODO:  This probably isn't robust to the myriad ways python
+        #  has to call functions with mixtures of positional and keyword args.
+        #  Appears that connexion always calls in with keyword arguments,
+        #  but should also work if we try to call a wrapped function with
+        #  positional arguments
+
+        # Since we don't know what arguments the function to wrap will take
```
wasade, line 275: Can we just restrict the use to kwargs so we don't have to care about positional? That should reduce the complexity below

microsetta_private_api/example/client_impl.py
```diff
@@ -211,21 +214,142 @@ def _check_source_prereqs(acct_id, source_id, current_state=None):
...
+                bound_map.get('account_id'),
+                bound_map.get('source_id')
+            )
+
+            # Route to closest sink if state doesn't match a required state
+            # TODO: Do we actually need contains/set functionality?
+            #  It looks like it is only in use for the vioscreen calls, and
+            #  neither of those is technically in the NEEDS_SURVEY state ever.
+            #  They always run from SOURCE_PREREQS_MET
+            if isinstance(target_state, list) or isinstance(target_state, set):
```
wasade, line 318: `if isinstance(target_state, collections.abc.Iterable) and not isinstance(target_state, str)`

...this would cover `tuple`s and derived classes of `list`, etc etc

wasade: i think this could all be rewritten as:

```python
if isinstance(target_state, str):
    target_state = [target_state, ]

if prereqs_step not in target_state:
    return _route_to_closest_sink(prereqs_step, curr_state)
```

microsetta_private_api/example/client_impl.py
```diff
@@ -211,21 +214,142 @@ def _check_source_prereqs(acct_id, source_id, current_state=None):
...
+                if prereqs_step not in target_state:
+                    return _route_to_closest_sink(prereqs_step, curr_state)
+            else:
+                if prereqs_step != target_state:
+                    return _route_to_closest_sink(prereqs_step, curr_state)
+
+            # For any states that require checking additional parameters, we do
+            # so here.
+            # TODO:  This section feels like it could be removed with some more
```
wasade, line 327: This sounds cool :) but perhaps a future refactor

microsetta_private_api/example/client_impl.py
```diff
@@ -211,21 +214,142 @@ def _check_source_prereqs(acct_id, source_id, current_state=None):
...
+            #  parameterized state object from the function's parameters
+            #  and compare directly, rather than having state specific checks.
+
+            # TODO:  Please check you agree this logic correctly replaces the
+            #  survey rerouting.
+            if prereqs_step == NEEDS_SURVEY:
+                passed_id = bound_map.get('survey_template_id')
+                needed_id = curr_state.get("needed_survey_template_id")
+                passed_is_correct = passed_id == needed_id
+                if not passed_is_correct:
```
wasade, line 339: is there a reason to have the extra variable? `if passed_id != needed_id` seems reasonably expressive?

microsetta_private_api/util/decorator_util.py
```diff
@@ -0,0 +1,35 @@
...
+def build_param_map(func, param_names):
+    sig = inspect.signature(func)
+    params = sig.parameters
+
+    desired_set = set(param_names)
+
+    # Argh.  OrderedDict doesn't know how to find the indexes of its keys.
+    output = {}
+    idx = 0
+    for key in params:
```
wasade, line 13:
```suggestion
    for idx, (key, value) in enumerate(params.items()):
        if key in desired_set:
            output[key] = (idx, value.default)
```

microsetta_private_api/util/decorator_util.py
```diff
@@ -0,0 +1,35 @@
...
+        if key in desired_set:
+            output[key] = (idx, params[key].default)
+        idx += 1
+    return output
+
+
+def bind_param_map(param_map, parameter_overrides, func_args, func_kwargs):
+    output = {}
+    for key in param_map:
+        idx = param_map[key][0]
```
wasade, line 23:
```suggestion
        idx, default = param_map[key]
```

microsetta_private_api/util/decorator_util.py
```diff
@@ -0,0 +1,35 @@
...
+        default = param_map[key][1]
+        if idx < len(func_args):
+            output[key] = func_args[idx]
+        else:
+            output[key] = func_kwargs.get(key, default)
+
+    # TODO: We only use this to set survey_template_id on vioscreen callback
+    #  would we prefer defaults instead of overrides?
+    if parameter_overrides is not None:
```
wasade, line 32: this test shouldn't be needed, and it should be okay to just go into the forloop as `parameter_overrides` should be `{}` if not used:

```python
>>> def foo(**bar):
...   print(type(bar))
...   print(bar)
...
>>> foo()
<class 'dict'>
{}
```

microsetta_private_api/util/tests/test_decorator_util.py
```diff
@@ -0,0 +1,52 @@
...
+
+    assert func1("z", "a", "b", "x", "d", "y") == ("x", "y", "z", "Fig Newton")
+    assert func1("zz2", "a", "b", d=27, y="foobar") == \
+        (7, "foobar", "zz2", "Fig Newton")
+
+    # Hah, you'd think this wouldn't compile, and it would throw a TypeError
+    # if the wrapper method ever called into func1, but the wrapper itself
+    # has no idea what the args and kwargs are, and doesn't care if they
+    # overlap.  Bound params happens to prefer args over kwargs arbitrarily.
+    assert func1("z", "a", "b", "x", x=15) == ("x", None, "z", "Fig Newton")
```
wasade, line 34: I feel like we're fighting against the language here, which is risky. I'm not sure we need to support positional arguments, which would I believe avoid the introspection work. 

```python
>>> def foo(a, b, **kwargs):
...   print(a, b, kwargs)
...
>>> foo(1, 2, b=3)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: foo() got multiple values for argument 'b'
```

Would it be valid with connexxion, for example, to do the following?

```python
@prerequisite(SOURCE_PREREQS_MET)
def post_remove_sample_from_source(account_id=None, source_id=None, sample_id=None):
    # ... do stuff
```

or

```python
@prerequisite(SOURCE_PREREQS_MET)
def post_remove_sample_from_source(**kwargs):
    # validate kwargs in prerequisite
    # ... do stuff
```

if we *do* need positional arguments, then we may want to take a closer look at what `click` does at call resolution, which it does very well. It would entail, I think, enumerating the individual arguments in the decorator calls

microsetta_private_api/example/client_impl.py
dhakim87: Switched target_state to allowed_states and enforced that it is always a list.

microsetta_private_api/example/client_impl.py
dhakim87: Was a direct variable name refactor on code that already existed, removed the passed_is_correct variable.

microsetta_private_api/util/decorator_util.py
dhakim87: 👍

microsetta_private_api/util/decorator_util.py
dhakim87: Since I've moved this out to a separate utility class, I can't guarantee this will be enforced by all callers.  I'd rather leave a null check in.

microsetta_private_api/example/client_impl.py
dhakim87: I've implemented this as a separate PR against this branch so you can see what the change would look like.  It's feasible and reduces complexity in the decorator - but I think it increases the complexity for other developers starting with client_impl.  Devs who are new to python (like me), may find keyword arguments to be black magic at first, so putting asterisks in front of arguments is just terrifying.  I'm also slightly worried that this will couple us to Connexion, or break should connexion or flask decide to change their calling conventions ever so slightly.  I would lean towards the version using the decorator_util methods.

microsetta_private_api/util/tests/test_decorator_util.py
dhakim87: See my PR to this PR :)

microsetta_private_api/example/client_impl.py
wasade: On the other hand, regular python developers are comfortable with kwargs, while `inspect` and reflection in general are less frequent to encounter.

The use of kwargs actually increases protection from API changes. This allows those libraries to alter the order of the arguments, including positional and it won't break our code:

```python
>>> def foo(a, b=10):
...   print(a, b)
...
>>> foo(b=6, a=2)
2 6
```

The scenario that does present a problem is if the library changes a variable name. In explicitly typed languages like C, that would not represent an API change as the compiler sees `foo(int, int) -> null` irrespective of variable name. However, in Python, the ducktyping means these signatures are not (typically) reliable. In python, a change in variable name in a function signature *is an API change*. A properly maintained package will clearly note in the changelog and trigger at least a minor version bump if not major. 

This is part of why well used libraries like `flask`, `pandas`, `numpy`, etc make extensive use of kwargs within their codebases. A few examples are [`save_session`](https://github.com/pallets/flask/blob/master/src/flask/sessions.py#L375-L383) in `flask.sessions`, [`read_csv`](https://github.com/pandas-dev/pandas/blob/master/pandas/io/parsers.py#L540-L598) in `pandas.io.parsers`, and [`count_nonzero`](https://github.com/numpy/numpy/blob/master/numpy/core/numeric.py#L392) in `numpy`. Internally, these libraries also regularly use kwargs _to reduce bugs_ from API changes. 

All this said, we are flying somewhat blind with the client interface as we don't have integration tests, and we regardless of positional or kwargs, we would not catch issues until runtime.

PR #234, opened 2020-06-26
A patch for 199 showing kwargs only decorator

dhakim87:
Implementation restricting decorator to kwargs only functions.  It cleans up the decorator, but I feel it significantly increases the barrier to entry of client_impl for devs who are new to python (like me).  It's here as an option, but I prefer the decorator_util bound mappings version

### DIFF
--- a/microsetta_private_api/util/decorator_util.py
+++ b/microsetta_private_api/util/decorator_util.py
@@ -7,34 +7,3 @@ def has_non_keyword_arguments(func):
     for p in params:
         if params[p].kind != inspect.Parameter.KEYWORD_ONLY:
             return True
-
-
-def build_param_map(func, param_names):
-    sig = inspect.signature(func)
-    params = sig.parameters
-
-    desired_set = set(param_names)
-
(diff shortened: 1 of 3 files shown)


wasade:
The more I look at this, the more this looks like the safest long term approach

microsetta_private_api/util/decorator_util.py
```diff
@@ -1,6 +1,14 @@
...
 
+def has_non_keyword_arguments(func):
+    sig = inspect.signature(func)
+    params = sig.parameters
+    for p in params:
+        if params[p].kind != inspect.Parameter.KEYWORD_ONLY:
+            return True
+
+
 def build_param_map(func, param_names):
```
wasade, line 12: This is no longer used, right?

wasade: ...same with `bind_param_map`

dhakim87: Removed.

PR #237, opened 2020-06-27
St dh event log

dhakim87:
Adds schema change, new repo and new model object for representing an Event Log.  This will be used initially to track emails we send out to end users.  

Relevant to https://github.com/biocore/microsetta-admin/issues/17

### DIFF
--- a/microsetta_private_api/db/patches/0065.sql
+++ b/microsetta_private_api/db/patches/0065.sql
@@ -2,7 +2,7 @@ CREATE TABLE ag.event_log(
     id uuid PRIMARY KEY NOT NULL,
     event_type varchar(100) NOT NULL,
     event_subtype varchar(100) NOT NULL,
-    event_time timestamptz default current_timestamp,
+    event_time timestamptz default current_timestamp NOT NULL,
     event_state jsonb);
 
 -- Full event log sorted by time
--- a/microsetta_private_api/db/patches/0065.sql
+++ b/microsetta_private_api/db/patches/0065.sql
@@ -1,0 +1,11 @@
+CREATE TABLE ag.event_log(
+    id uuid PRIMARY KEY NOT NULL,
+    event_type varchar(100) NOT NULL,
+    event_time timestamptz default current_timestamp,
+    event_subtype varchar(100) NOT NULL,
+    event_state jsonb);
+
+CREATE UNIQUE INDEX idx_event_log_event_time ON ag.event_log (event_time);
+CREATE UNIQUE INDEX idx_event_log_event_type_event_time ON ag.event_log (event_type, event_time);
+CREATE UNIQUE INDEX idx_events_type_time ON ag.event_log (event_type, event_time);
+
--- a/microsetta_private_api/db/patches/0065.sql
+++ b/microsetta_private_api/db/patches/0065.sql
@@ -1,11 +1,17 @@
 CREATE TABLE ag.event_log(
     id uuid PRIMARY KEY NOT NULL,
     event_type varchar(100) NOT NULL,
-    event_time timestamptz default current_timestamp,
     event_subtype varchar(100) NOT NULL,
+    event_time timestamptz default current_timestamp,
     event_state jsonb);
 
-CREATE UNIQUE INDEX idx_event_log_event_time ON ag.event_log (event_time);
-CREATE UNIQUE INDEX idx_event_log_event_type_event_time ON ag.event_log (event_type, event_time);
-CREATE UNIQUE INDEX idx_events_type_time ON ag.event_log (event_type, event_time);
-
+-- Full event log sorted by time
+CREATE INDEX idx_event_log_event_time ON ag.event_log (event_time);
+-- Event log filtered by type sorted by time
+CREATE INDEX idx_event_log_event_type_event_time ON ag.event_log (event_type, event_time);
+-- Event log filtered by type and subtype sorted by time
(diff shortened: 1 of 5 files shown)


microsetta_private_api/db/patches/0065.sql
```diff
@@ -0,0 +1,17 @@
+CREATE TABLE ag.event_log(
+    id uuid PRIMARY KEY NOT NULL,
```
wasade, line 2: just curious, why never rely in postgres to create these as done [here](https://github.com/biocore/microsetta-private-api/blob/8c9ce8e85f79842b5f0b21d97ef6204737a58a0f/microsetta_private_api/db/patches/0055.sql#L13)?

microsetta_private_api/db/patches/0065.sql
```diff
@@ -0,0 +1,17 @@
...
+    event_type varchar(100) NOT NULL,
+    event_subtype varchar(100) NOT NULL,
+    event_time timestamptz default current_timestamp,
```
wasade, line 5: can this be not-null as well?

microsetta_private_api/db/patches/0065.sql
```diff
@@ -0,0 +1,17 @@
...
+    event_state jsonb);
+
+-- Full event log sorted by time
+CREATE INDEX idx_event_log_event_time ON ag.event_log (event_time);
+-- Event log filtered by type sorted by time
+CREATE INDEX idx_event_log_event_type_event_time ON ag.event_log (event_type, event_time);
+-- Event log filtered by type and subtype sorted by time
+CREATE INDEX idx_event_log_event_type_event_subtype_event_time ON ag.event_log (event_type, event_subtype, event_time);
+-- Event log filtered by user email sorted by time
+CREATE INDEX idx_events_state_email_time ON ag.event_log ((event_state->>'email'), event_time);
```
wasade, line 15: ohhhh... that's nifty. Just to check, does this force all `event_state` entries to have an `email`?

microsetta_private_api/repo/event_log_repo.py
```diff
@@ -0,0 +1,122 @@
...
+                        "event_subtype = %s "
+                        "ORDER BY "
+                        "event_time DESC",
+                        (event_type.value, event_subtype.value))
+            return [_row_to_event(row) for row in cur.fetchall()]
+
+    # See https://www.postgresql.org/docs/9.5/functions-json.html#FUNCTIONS-JSON-OP-TABLE  # noqa
+    # to understand referencing email field from jsonb representation
+
+    # TODO: I believe the LIKE operator can make use of the btree index i've
```
wasade, line 87: Should we force emails in the db to be lowercase?

microsetta_private_api/db/patches/0065.sql
dhakim87: No particularly strong reason I don't think.  Uuids are great because anybody can make them.  It's kind of nice for the writer not having to fetch the newly created row to know the ID (and how would you even write such a fetch without knowing the ID?).  But at the same time, the writer doesn't know the timestamp without doing a fetch anyway, so 🤷‍♂️.

microsetta_private_api/db/patches/0065.sql
dhakim87: Yep, I'll make that explicit

microsetta_private_api/db/patches/0065.sql
dhakim87: Nope, see the last test case, search works fine even if some entries don't have nulls.  I'm not sure if this is a partial index or it just counts the ones that don't have it as null in some sense.  It should be good enough to do prefix searches based on those fields I believe.  There's also some extreme black magic with gin indexes that might allow for arbitrary text search, but it looked to have random hard to describe limitations on what fields were searchable and how it interacted with other indexes.  Best way to figure out what index will be used by a query remains to run EXPLAIN on it. 

`ag_test=# EXPLAIN select * from event_log where event_state->>'email' LIKE 'abc' ORDER BY event_state->>'email', event_time DESC;
                                             QUERY PLAN                                              
-----------------------------------------------------------------------------------------------------
 Sort  (cost=8.18..8.19 rows=1 width=492)
   Sort Key: ((event_state ->> 'email'::text)), event_time DESC
   ->  Index Scan using idx_events_state_email_time on event_log  (cost=0.14..8.17 rows=1 width=492)
         Index Cond: ((event_state ->> 'email'::text) = 'abc'::text)
         Filter: ((event_state ->> 'email'::text) ~~ 'abc'::text)
(5 rows)
`

microsetta_private_api/repo/event_log_repo.py
dhakim87: Nah.  If someone really does enter an uppercase email, might as well persist whatever they enter.  I don't know enough about email to ensure it really is case insensitive, so I'd rather not break whatever pathological edge case arises.

dhakim87: I'll switch it to a -slow- ILIKE operator instead, since the EXPLAIN query indicates its doing a full table scan anyway.  We can always make a separate function call for pulling exact email match, which does use the indexes.

microsetta_private_api/db/patches/0065.sql
wasade: fair. let's leave as is, though i think `returning` can be used in a query if needed in the future

wasade:
looks like something with travis is still running but am +1 after

PR #241, opened 2020-07-01
Project banking support

AmandaBirmingham:
Supports fixing microsetta-admin repo's issues 16 and 18.

This PR adds functionality to the private api's admin endpoints to (a) pass into project creation info on whether or not a project should be banked and if so until what (optional) plating start date and (b) adds this banked-or-not and plating start date info to the project info returned when searching for info on a barcode.  Tests are updated/added to exercise this functionality.

It also changes the /api/admin/scan/{barcode} endpoint that updates sample info on scanning from a post to a put (and changes the http return code from 201 to 200) because this endpoint isn't creating a new record but rather updating an existing one.

### DIFF
--- a/microsetta_private_api/api/microsetta_private_api.yaml
+++ b/microsetta_private_api/api/microsetta_private_api.yaml
@@ -918,7 +918,7 @@ paths:
         '422':
           $ref: '#/components/responses/422UnprocessableEntity'
 
-  # START ADMINISTRATOR PATHS
+  # ADMINISTRATOR PATHS
   '/admin/search/samples/{sample_barcode}':
     get:
       operationId: microsetta_private_api.admin.admin_impl.search_barcode
@@ -939,6 +939,7 @@ paths:
           $ref: '#/components/responses/401Unauthorized'
         '404':
           $ref: '#/components/responses/404NotFound'
+
   '/admin/search/kit/{kit_id}':
     get:
       operationId: microsetta_private_api.admin.admin_impl.search_kit_id
@@ -964,6 +965,7 @@ paths:
           $ref: '#/components/responses/401Unauthorized'
         '404':
           $ref: '#/components/responses/404NotFound'
+
   '/admin/search/account/{email}':
     get:
       operationId: microsetta_private_api.admin.admin_impl.search_email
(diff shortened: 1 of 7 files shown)


microsetta_private_api/api/microsetta_private_api.yaml
```diff
@@ -990,8 +992,9 @@ paths:
           $ref: '#/components/responses/401Unauthorized'
         '404':
           $ref: '#/components/responses/404NotFound'
+
   '/admin/scan/{sample_barcode}':
-    post:
+    put:
```
wasade, line 997: We'll need to make sure a deploy with this commit is timed with an updated deploy of microsetta-admin

microsetta_private_api/api/tests/test_admin_api.py
```diff
@@ -0,0 +1,186 @@
...
+
+        # execute project post (create)
+        response = self.client.post(
+            "/api/admin/create/project",
+            content_type='application/json',
+            data=input_json,
+            headers=MOCK_HEADERS
+        )
+
+        # NB: the date format SHOULD be checked by connexion
```
wasade, line 177: ugh!! should we `try/except` around `datetime.strptime` prior to the sql query?

microsetta_private_api/repo/admin_repo.py
```diff
@@ -148,10 +157,26 @@ def create_project(self, project_name, is_microsetta):
...
+    def delete_project_by_name(self, project_name):
+        """Delete a project entry in the database
+
+        Parameters
+        ----------
+        project_name : str
+            The name of the project to delete
+        """
+        with self._transaction.cursor() as cur:
+            cur.execute("DELETE FROM barcodes.project WHERE project = %s",
```
wasade, line 176: Doesn't this trigger a FK issue with the many-many join table `barcodes.project_barcode`?

AmandaBirmingham: Good catch--I didn't see that bc I only wrote this for use by tests, and the tests I was using it for didn't include any barcode associations!  Added code to delete the barcode associations first ...

microsetta_private_api/api/tests/test_admin_api.py
AmandaBirmingham: Added try/except

microsetta_private_api/api/microsetta_private_api.yaml
AmandaBirmingham: Correct

PR #247, opened 2020-07-27
Update Email Templates - User can't edit locked samples

dhakim87:
Rewording in email templates since user can't currently edit their samples directly.

### DIFF
--- a/microsetta_private_api/admin/email_templates.py
+++ b/microsetta_private_api/admin/email_templates.py
@@ -34,6 +34,22 @@ class EmailMessage(Enum):
         EventType.EMAIL,
         EventSubtype.EMAIL_MISSING_SAMPLE_INFO
     )
+    sample_is_valid = (
+        "Sample is valid",
+        "email/sample_is_valid.jinja2",
+        "email/sample_is_valid.plain",
+        ("contact_name", "sample_barcode", "received_type", "resolution_url"),
+        EventType.EMAIL,
+        EventSubtype.EMAIL_SAMPLE_IS_VALID
+    )
+    no_associated_source = (
+        "No associated source",
+        "email/no_associated_source.jinja2",
+        "email/no_associated_source.plain",
+        ("contact_name", "sample_barcode", "received_type", "resolution_url"),
+        EventType.EMAIL,
+        EventSubtype.EMAIL_NO_SOURCE
+    )
 
     def __init__(self, subject, html, plain, required, event_type, event_sub):
         self.subject = subject
--- a/microsetta_private_api/admin/email_templates.py
+++ b/microsetta_private_api/admin/email_templates.py
@@ -38,7 +38,7 @@ class EmailMessage(Enum):
         "Sample is valid",
         "email/sample_is_valid.jinja2",
         "email/sample_is_valid.plain",
-        ("contact_name", "sample_barcode", "received_type", "resolution_url"),
+        ("contact_name",),
         EventType.EMAIL,
         EventSubtype.EMAIL_SAMPLE_IS_VALID
     )
@@ -46,7 +46,7 @@ class EmailMessage(Enum):
         "No associated source",
         "email/no_associated_source.jinja2",
         "email/no_associated_source.plain",
-        ("contact_name", "sample_barcode", "received_type", "resolution_url"),
+        ("contact_name", "sample_barcode", "resolution_url"),
         EventType.EMAIL,
         EventSubtype.EMAIL_NO_SOURCE
     )
--- a/microsetta_private_api/api/_sample.py
+++ b/microsetta_private_api/api/_sample.py
@@ -11,6 +11,7 @@ from microsetta_private_api.repo.source_repo import SourceRepo
 from microsetta_private_api.repo.survey_answers_repo import SurveyAnswersRepo
 from microsetta_private_api.repo.transaction import Transaction
 from microsetta_private_api.util.util import fromisotime
+from microsetta_private_api.admin.admin_impl import token_grants_admin_access
 
 
 def read_sample_associations(account_id, source_id, token_info):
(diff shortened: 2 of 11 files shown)


wasade:
Thanks! A proposed email for "sample-is-valid":

```text
Hello {{contact_name |e}},
    Thank you for your interest and participation in The Microsetta Initiative. We have received your sample, and it is now in the processing queue.

    If you have any questions, please reply to us at microsetta@ucsd.edu

    Thank you,
    The Microsetta Team
```

And I think "no-associated-source", I think the following:

```text
Hello {{contact_name |e}},
    Thank you for your interest and participation in The Microsetta Initiative. We are writing as we've encountered an issue processing your sample, and we would appreciate your assistance in helping to resolve this matter.

    Specifically, the following sample sent to the lab is not associated with a source:
        Barcode {{sample_barcode |e}}

    We require this information in order to process this sample in compliance with our human subjects research protocol. Please reconnect to the participant website, create a source, add the sample to the source, and record the sample specific information (e.g., date / time of collection). You can access your account by clicking on or copying the following URL:

    {{ account_url}}

    If you have any questions, please reply to us at microsetta@ucsd.edu

    Thank you,
    The Microsetta Team
```

@specsdom, these seem good?

specsdom:
> ```
> Hello {{contact_name |e}},
>     Thank you for your interest and participation in The Microsetta Initiative. We have received your sample, and it is now in the processing queue.
> 
>     If you have any questions, please reply to us at microsetta@ucsd.edu
> 
>     Thank you,
>     The Microsetta Team
> ```

For the sample-is-valid email, I'm trying to think of how to reword this to be inclusive of studies where the samples are being banked (since saying a sample is in the processing queue could suggest that a sample will be processed soon), but it inevitably just sounds vague if we say "some samples will be banked and processed later" without being clear on which ones will be affected... which could be okay. If people really are curious, they could ask the help account.

I know we discussed previously having two methods in place: One for when samples are received, but get banked, and another when a sample are actually being plated for processing.  This adds some more complexity, but could we instead have two different templates that go out for "sample-is-valid" samples?  Like what we suggested with the samples that have inconsistencies, there could be an additional menu drop-down that lets the technician choose which email to send out: One for samples that are actively being plated, and one for samples that are being banked (which would also change the workflow for technicians, as that would effectively mean samples should be re-scanned once they do get plated, so participants know that their sample is now processing).

Alternatively, on the Microsetta public website, we wanted to have a page dedicated to active cohorts and specify processing times (looks like [this right now](https://www.dropbox.com/s/ljgczcghg072ae8/TMI-19-005%20Website%20V1-%20Participant%20Resources%20-%20D1-Active%20Cohort%20Info.jpg?dl=0)). We could link to this page in the sample-is-valid email so users can be aware the status of studies and when we're processing samples from a specific study, then keep it simplified to just one email. Only problem is this page isn't up yet, but I could put something really simple together until Departures has it up.

> And I think "no-associated-source", I think the following:
> 
> ```
> Hello {{contact_name |e}},
>     Thank you for your interest and participation in The Microsetta Initiative. We are writing as we've encountered an issue processing your sample, and we would appreciate your assistance in helping to resolve this matter.
> 
>     Specifically, the following sample sent to the lab is not associated with a source:
>         Barcode {{sample_barcode |e}}
> 
>     We require this information in order to process this sample in compliance with our human subjects research protocol. Please log back into the participant website by clicking or copying on the following URL:
>     {{ account_url}}
>
> After accessing your account using the URL above, you will need to:
> 1.) **Create a source**: Click on one of the buttons under the "Sources" section
> 2.) **Add the sample to the source**: After creating a source, claim the sample using the associated kit ID
> 3.) **Record the sample specific information (e.g., date / time of collection)**: After claiming the sample, click on its button to update its information
> 
>     If you have any questions, please reply to us at microsetta@ucsd.edu
> 
>     Thank you,
>     The Microsetta Team
> ```

Modified this a bit--when I list out the steps with instructions to participants, they usually have a much easier time understanding what to do.

Also RE the email templates for samples that are missing info or have inconsistencies: The ones currently being used should still be good.

wasade:
@dhakim87 and @specsdom, templates added in, I think good for review

wasade:
Annoyingly that push brought over some other commits. This branch must have been created before the admin override piece was merged?

specsdom:
> @dhakim87 and @specsdom, templates added in, I think good for review

Looks good!  Guessing the .plain file for the sample-is-valid email is the correct one to check.  Presumably then it'll be easier to link to the active cohorts page (once available) in this template instead so participants can check the processing status of their project?

wasade:
The .plain and .jinja2 templates are for supporting both plaintext and HTML email clients -- the content should be the same.

microsetta_private_api/admin/email_templates.py
```diff
@@ -34,6 +34,22 @@ class EmailMessage(Enum):
...
+        EventSubtype.EMAIL_SAMPLE_IS_VALID
+    )
+    no_associated_source = (
+        "No associated source",
+        "email/no_associated_source.jinja2",
+        "email/no_associated_source.plain",
+        ("contact_name", "sample_barcode", "received_type", "resolution_url"),
+        EventType.EMAIL,
+        EventSubtype.EMAIL_NO_SOURCE
+    )
```
dhakim87, line 52: Looked over the jinja2 templates associated with these, and I don't see received_type required in either of the new templates.  Did we want to pass that info along to the user, or should we drop it from the list of required fields when filling out the templates?  

Same question for resolution_url on sample_is_valid

Same question for sample_barcode on sample_is_valid

wasade: Ah we should probably remove, good call

dhakim87: Updated.

PR #249, opened 2020-08-08
Admin override for associate sample

### LINKED ISSUE #248: Sample locking override does not work for "no-registered-account" samples
@specsdom is reporting that the sample lock override is not working for samples that have a scan status of "no-registered-account"

wasade:
Fixes #248 

The code path used for claiming a sample did not get the lockout override added in. This PR addresses that.

### DIFF
--- a/microsetta_private_api/repo/sample_repo.py
+++ b/microsetta_private_api/repo/sample_repo.py
@@ -176,7 +176,8 @@ class SampleRepo(BaseRepo):
                             sample_info.id
                         ))
 
-    def associate_sample(self, account_id, source_id, sample_id):
+    def associate_sample(self, account_id, source_id, sample_id,
+                         override_locked=False):
         with self._transaction.cursor() as cur:
             cur.execute("SELECT "
                         "ag_kit_barcode_id, "
@@ -193,7 +194,6 @@ class SampleRepo(BaseRepo):
             if row is None:
                 raise werkzeug.exceptions.NotFound("No sample ID: %s" %
                                                    sample_id)
-
             if row[2] is not None:
                 if row[1] != account_id:
                     # This is the case where the sample is already assigned in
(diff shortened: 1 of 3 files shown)


microsetta_private_api/repo/sample_repo.py
```diff
@@ -205,9 +205,8 @@ def associate_sample(self, account_id, source_id, sample_id):
                     raise RepoException("Sample is already assigned")
             else:
                 # This is the case where the sample is not yet assigned
-                # NOTE: we are not passing override_locked here as a sample
```
wasade, line 208: :/ this comment didn't hold up

wasade:
cc @specsdom

microsetta_private_api/repo/sample_repo.py
dhakim87: I don't follow this change.  The if statement above indicates that this particular sample is not assigned to a source.  The two raise RepoException lines would be the ones that prevent changing locked samples.  Why don't we have to replace the logic on those RepoExceptions to allow being overridden by the flag if the problem is that admins can't associate samples that are already associated?

wasade: The problem here is that admins cannot override unassigned samples that have been scanned in. A scanned sample is locked.

wasade:
Samples that are scanned are locked, and this is true when sample status is "no-registered-account"

microsetta_private_api/repo/sample_repo.py
AmandaBirmingham: @dhakim87 the two RepoExceptions will prevent changing samples, but not because the samples are "locked" -- just because doing so would make a dog's breakfast of the data :). The error Dom reported that says you can't edit a sample because it is "locked" comes from within the `_update_sample_association` call:

https://github.com/biocore/microsetta-private-api/blob/7f7148a09c49368a970a0b0c8da72dbf3f72e41e/microsetta_private_api/repo/sample_repo.py#L73-L79 

I think that passing the override variable here (same as for `update_sample_association` and `dissociate_sample`) indeed should take care of this.

dhakim87:
Ahhhh, that makes sense then.  Okay so separate change in the future if we ever need an admin to move samples that are in the wrong account/source

microsetta_private_api/repo/sample_repo.py
```diff
@@ -205,9 +205,8 @@ def associate_sample(self, account_id, source_id, sample_id):
...
-                # that is not yet assigned cannot be locked.
-                self._update_sample_association(sample_id, source_id)
+                self._update_sample_association(sample_id, source_id,
+                                                override_locked=override_locked)  # noqa
```
AmandaBirmingham, line 209: curious--what lint complaint is the noqa here overriding?

wasade: It's like 81 characters, and I thought it would be less readable to use a `\`

PR #253, opened 2020-08-12
Support to fix admin issue 42

AmandaBirmingham:
Primarily extends admin_repo.py's _get_ids_relevant_to_barcode to identify cases where barcode is not attached to a source but IS associated with an account in the sense that it is part of a kit that has been used to set up an account; this case will get (only) the 'no-associated-source' warning in admin interface as part of forthcoming issue 42 fix.  Extends tests to separately exercise each of the cases that will lead to different warnings in admin interface.

### DIFF
--- a/microsetta_private_api/repo/admin_repo.py
+++ b/microsetta_private_api/repo/admin_repo.py
@@ -23,6 +23,10 @@ class AdminRepo(BaseRepo):
 
     def _get_ids_relevant_to_barcode(self, sample_barcode):
         with self._transaction.dict_cursor() as cur:
+            # First, look for this barcode in the set of barcodes
+            # associated with AGP kits; if it is there, get its kit id and
+            # follow the source id associated with that AGP barcode to its
+            # source and account ids.
             cur.execute(
                 "SELECT "
                 "ag_kit_barcodes.ag_kit_barcode_id as sample_id, "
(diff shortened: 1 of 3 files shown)


microsetta_private_api/repo/admin_repo.py
```diff
@@ -31,18 +35,45 @@ def _get_ids_relevant_to_barcode(self, sample_barcode):
...
+            if result is not None:
+                return result
+
+            # else if the barcode is not associated with a source id, it may
+            # still be (more tenuously) associated with an account if it in
+            # the set of barcodes associated with AGP kits and the kit it is
+            # in has been used to open an AGP account; in this case, we can get
+            # an account id and a kit id but NOT a source id.
+            # Even if there is no such account association, we can still get
+            # the (official, UUID) kit id if this is an AGP kit
```
wasade, line 60: I don't think this is returning `barcodes.kit.kit_uuid`?

wasade: Oh, I see, the UUID reference here is to `ag.ag_kit.ag_kit_id`?

wasade: (for clarify, the former refers to the UUID placed on the outside of the physical kit someone receives)

AmandaBirmingham: Comments modified.  I can't help feeling that there are waaaay too many kinds of kit ids floating around ... for example, it is very confusing to me that the "kit_id" used in ag_kit_barcodes.ag_kit_id is a UUID (presumably from the ag.ag_kits table, although I don't see a foreign key relation ...?) but the "kit_id" used in ag.account.created_with_kit_id is an external-user-facing string like 'ju-gtbfd' (I think this matches to ag.ag_kit.supplied_kit_id, although, again, I don't see a foreign key relation.)  The "kit_id" in barcodes.barcode.kit_id is also this external-user-facing string, but the same table also has a kit_uuid ... Ohhhh, my head ...

AmandaBirmingham:
@wasade I think I've covered your accepted change; if so, can you accept?

PR #254, opened 2020-08-20
Extend project handling, including computed statistics

AmandaBirmingham:
I think I have the major functionality down here--all the stuff I had two weeks ago plus considerably extended computed metrics about kits/samples/etc requested by Edgar.  I would like to get feedback on concerns/issues before writing up the unit tests for all this ...


microsetta_private_api/api/microsetta_private_api.yaml
```diff
@@ -1760,6 +1769,130 @@ components:
...
+          nullable: true
+        contact_email:
+          type: string
+          format: email
+          nullable: true
+        deadlines:
+          type: string
+          nullable: true
+        num_subjects:
+          # no data type specified--this can be integer or string :-|
```
wasade, line 1822: Like "8" vs "eight"?

microsetta_private_api/api/microsetta_private_api.yaml
```diff
@@ -1760,6 +1769,130 @@ components:
...
+          nullable: true
+        num_timepoints:
+          # no data type specified--this can be integer or string :-|
```
wasade, line 1825: Same q here

microsetta_private_api/db/patches/0069.sql
```diff
@@ -0,0 +1,36 @@
...
+-- extend the project table to hold lots more info
+ALTER TABLE barcodes.project ADD COLUMN subproject_name varchar;
+ALTER TABLE barcodes.project ADD COLUMN alias varchar;
+ALTER TABLE barcodes.project ADD COLUMN sponsor varchar;
+ALTER TABLE barcodes.project ADD COLUMN coordination varchar;
+ALTER TABLE barcodes.project ADD COLUMN contact_name varchar;
+ALTER TABLE barcodes.project ADD COLUMN additional_contact_name varchar;
+ALTER TABLE barcodes.project ADD COLUMN contact_email varchar;
+ALTER TABLE barcodes.project ADD COLUMN deadlines varchar;
+ALTER TABLE barcodes.project ADD COLUMN num_subjects varchar;
```
wasade, line 14: This feels weird...

microsetta_private_api/model/project.py
```diff
@@ -0,0 +1,181 @@
...
+# get_computed_stats_keys() below.
+NUM_KITS_KEY = 'num_kits'
+NUM_SAMPLES_KEY = 'num_samples'
+NUM_SAMPLES_RECEIVED_KEY = 'num_samples_received'
+NUM_UNIQUE_SOURCES_KEY = 'num_unique_sources'
+NUM_PARTIALLY_RETURNED_KITS_KEY = 'num_partially_returned_kits'
+NUM_FULLY_RETURNED_KITS_KEY = 'num_fully_returned_kits'
+NUM_KITS_W_PROBLEMS_KEY = "num_kits_w_problems"
+
+# I hate these status strings. They are independently (re)defined
```
wasade, line 57: Yea, at some point we should really centralize this somehow

microsetta_private_api/model/project.py
```diff
@@ -0,0 +1,181 @@
...
+        self.computed_stats = kwargs.get(COMPUTED_STATS_KEY, {})
+
+        project_name = kwargs.get(PROJ_NAME_KEY)
+        db_project_name = kwargs.get(DB_PROJ_NAME_KEY)
+
+        if project_name is None:
+            if db_project_name is None:
+                raise ValueError("No project name provided")
+            self.project_name = db_project_name
+        elif project_name is not None:
```
wasade, line 151: I think this is equivalent to `else:`

microsetta_private_api/model/project.py
```diff
@@ -0,0 +1,181 @@
...
+        if not self.bank_samples and self.plating_start_date is not None:
+            raise ValueError("Plating start date cannot be set for"
+                             " unbanked projects")
+
+    @classmethod
+    def from_db(cls, row):
+        row_dict = dict(row)
+        return Project(**row_dict)
+
+    @staticmethod
```
wasade, line 176: Was there a reason to make this a `staticmethod` rather than a `classmethod` like `from_db`?

microsetta_private_api/model/project.py
```diff
@@ -0,0 +1,181 @@
...
+                        "YYYY-MM-DD format".format(self.plating_start_date))
+
+        if not self.bank_samples and self.plating_start_date is not None:
+            raise ValueError("Plating start date cannot be set for"
+                             " unbanked projects")
+
+    @classmethod
+    def from_db(cls, row):
+        row_dict = dict(row)
+        return Project(**row_dict)
```
wasade, line 174: For `classmethod`'s, I think the convention would be to `return cls(**row_dict)`

microsetta_private_api/repo/admin_repo.py
```diff
@@ -16,11 +23,237 @@
 
 from microsetta_private_api.repo.survey_answers_repo import SurveyAnswersRepo
 
+# TODO: Refactor repeated elements in project-related sql queries?
+PROJECTS_BASICS_SQL = """
+    SELECT
+    p.*,
+    count(distinct barcode) as {0},  -- number of samples
```
wasade, line 30: These column names aren't being filled in to avoid replicating the field names, right?

microsetta_private_api/repo/admin_repo.py
```diff
@@ -147,38 +380,158 @@ def _rows_to_dicts_list(rows):
...
+                id_ = cur.fetchone()[0]
 
             cur.execute("INSERT INTO barcodes.project "
                         "(project_id, project, is_microsetta, bank_samples, "
-                        "plating_start_date) "
-                        "VALUES (%s, %s, %s, %s, %s)",
-                        [id_, project_name, tmi, bank_samples,
-                         plating_start_date])
-        return True
+                        "plating_start_date, contact_name, "
```
wasade, line 402: Should these also be derived from the constants? One thing that might be interesting to consider are f-strings here as they'd describe the field rather than the `.format` which is index based. If you haven't seen them:

```python
>>> a_variable = 'stuff and stuff'
>>> f'things {a_variable} and other things'
'things stuff and stuff and other things'
```

microsetta_private_api/repo/admin_repo.py
```diff
@@ -201,6 +554,48 @@ def delete_project_by_name(self, project_name):
...
+            computed_stats_keys = get_computed_stats_keys()
+            for curr_stats_key in computed_stats_keys:
+                curr_stat = v.pop(curr_stats_key)
+
+                # only NaN returns false when compared to itself;
+                # in case of NaN, set value of metric to 0
+                if curr_stat != curr_stat:
+                    curr_stat = 0
+                # alternately, if the metric is an integer, cast it to that;
+                # for some weird reason pandas is pulling in counts as floats
```
wasade, line 587: pandas does unexpected things sometimes. Would it make sense to cast integer columns up front (e.g., `df['foo'] = df['foo'].astype(int)`)?

microsetta_private_api/repo/admin_repo.py
```diff
@@ -201,6 +554,48 @@ def delete_project_by_name(self, project_name):
...
+        # turn the above dictionary into a list of Project objects, ordered
+        # by project id, while also pulling out the computed statistics for
+        # each project into their own sub-dictionary in the Project object.
+        result = []
+        for k, v in projects_dict.items():
+            stats_dict = {}
+            # pull computed statistics out of main project dictionary and
+            # into a sub-dictionary, cleaning them up in the process
+            computed_stats_keys = get_computed_stats_keys()
+            for curr_stats_key in computed_stats_keys:
```
wasade, line 579: It might be feasible to leverage the pandas API a little more here. Something like the following (not tested!!):

```python
def _normalize_types(v):
    if pd.isnull(v):
        return 0
    elif v == int(v):
        return int(v)
    else:
        return v

stats_df = projects_df[get_computed_stats_keys()]
stats_df.apply(_normalize_types, inplace=True)
for project_id in projects_df.index:
    # ...these _should_ still be in index order but using .loc to be safe
    project = project_df.loc[project_id].to_dict()
    stats = stats_df.loc[project_id].to_dict()
    project[COMPUTED_STATS_KEY] = stats
    result.append(Project(**project))  # should this be Project.from_dict?
```

microsetta_private_api/model/project.py
AmandaBirmingham: Someday ....

microsetta_private_api/model/project.py
AmandaBirmingham: Not particularly, except that it didn't seem to need the cls (but neither does from_db, afaict).  Will make them both class methods.

microsetta_private_api/model/project.py
AmandaBirmingham: Done

microsetta_private_api/model/project.py
AmandaBirmingham: changed

microsetta_private_api/db/patches/0069.sql
AmandaBirmingham: added

microsetta_private_api/repo/admin_repo.py
AmandaBirmingham: I didn't specify the field names originally because the list of fields in the table kept growing :-/. However, given that the list of allowed field names is explicitly specified as param names in the Project __init__, and that passing in additional unexpected params (that originate here) would break that, I guess this should be specified ... that is, assuming we continue to go the route of explicitly specifying the params in the Project object, which @dhakim87  dislikes :)

microsetta_private_api/model/project.py
```diff
@@ -0,0 +1,181 @@
...
+                raise ValueError("No project name provided")
+            self.project_name = db_project_name
+        elif project_name is not None:
+            if db_project_name is not None:
+                if project_name != db_project_name:
+                    raise ValueError("Conflicting project names provided")
+            self.project_name = project_name
+
+        if self.plating_start_date is not None:
+            if isinstance(self.plating_start_date, str):
```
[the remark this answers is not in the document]
AmandaBirmingham: Sometimes it actually is (as the name suggests) a date :) Whether the info comes into this constructor as a string or a date depends, if I recall correctly, on whether it got here from a form's text input field via `admin_impl.py`'s `create_project(body, token_info)` (in which case it is a string) or from the explicitly-date-type db field via `admin_repo.py`'s `get_projects()` (in which case it is already a date object).  I have added a comment to this effect.

microsetta_private_api/repo/admin_repo.py
```diff
@@ -16,11 +23,237 @@
...
+    ORDER BY project_id;
+"""
+
+# The below sql statements pull various computed counts about projects.
+# Each query must follow the convention that it returns a column named
+# "project_id" to join on, and that all other columns it returns are unique
+# to that query (not also returned by any other query).
+NUM_UNIQUE_SOURCES_SQL = """
+    SELECT project_barcode.project_id,
+    count(distinct ag_kit_barcodes.source_id) as {0}  --num_unique_sources
```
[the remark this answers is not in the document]
AmandaBirmingham: The issue is that I need each of these computed-column names to exist in the sql, and I also need it to exist in a list of names that I loop over and do after-market processing on (in `get_projects()`).  I didn't want to hardcode each of these names twice so I included them this way.

microsetta_private_api/model/project.py
```diff
@@ -0,0 +1,181 @@
...
+    def from_db(cls, row):
+        row_dict = dict(row)
+        return Project(**row_dict)
+
+    @staticmethod
+    def from_dict(values_dict):
+        return Project(**values_dict)
+
+    def to_api(self):
+        return vars(self)
```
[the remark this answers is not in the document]
wasade: I don't think that example corresponds to the use here as `vars` is called on an instance not the object:

```python
In [6]: class Person:
   ...:     name = "John"
   ...:     age = 36
   ...:     country = "norway"
   ...:

In [7]: x = Person()

In [8]: vars(x)
Out[8]: {}

In [10]: class Person:
    ...:     name = "John"
    ...:     age = 36
    ...:     country = "norway"
    ...:     def __init__(self):
    ...:         self.foobar = 10
    ...:

In [11]: x = Person()

In [12]: vars(x)
Out[12]: {'foobar': 10}
```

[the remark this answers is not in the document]
dhakim87: :facepalm:  Ignore me, I don't know python

microsetta_private_api/repo/admin_repo.py
```diff
@@ -16,11 +23,237 @@
...
+        -- get total number of samples (barcodes) in each kit, per proj
+        SELECT
+        project_barcode.project_id,
+        ag_kit_barcodes.ag_kit_id,
+        count(ag_kit_barcodes.barcode)
+        FROM ag.ag_kit_barcodes
+        LEFT JOIN barcodes.project_barcode
+        USING (barcode)
+        GROUP BY project_id, ag_kit_id
+        INTERSECT
```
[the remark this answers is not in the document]
AmandaBirmingham: @dhakim87  I tinkered around with it ... how about this?

```
    SELECT project_id,
    count(distinct ag_kit_id) as num_fully_returned_kits
    FROM (
	SELECT
        project_barcode.project_id,
        ag_kit_barcodes.ag_kit_id,
        count(distinct ag_kit_barcodes.barcode) as uniq_barcodes_in_kit,
	count(distinct barcode_scans.barcode) as uniq_received_barcodes_in_kit
        FROM ag.ag_kit_barcodes
        INNER JOIN barcodes.project_barcode
        USING (barcode)
	LEFT JOIN barcodes.barcode_scans
	USING (barcode)
        GROUP BY project_id, ag_kit_id
    ) as kit_lists
    WHERE uniq_barcodes_in_kit = uniq_received_barcodes_in_kit
    GROUP by project_id
    ORDER by project_id;
```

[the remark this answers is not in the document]
dhakim87: Looks pretty good!

You can remove the counts from the inner select's table with a HAVING clause rather than a WHERE, but I don't know if that actually makes any difference - I don't see any way to bring this down to a single select due to the two separate GROUP BYs, so I'd say good as is!

```
SELECT
project_barcode.project_id,
ag_kit_barcodes.ag_kit_id 
FROM ag.ag_kit_barcodes 
INNER JOIN barcodes.project_barcode 
USING (barcode) 
LEFT JOIN barcodes.barcode_scans                                                                                                                        USING (barcode)
GROUP BY project_id, ag_kit_id 
HAVING count(distinct ag_kit_barcodes.barcode)=count(distinct barcode_scans.barcode);
```

microsetta_private_api/model/project.py
```diff
@@ -0,0 +1,181 @@
...
+              NUM_SAMPLES_RECEIVED_KEY, NUM_UNIQUE_SOURCES_KEY,
+              NUM_PARTIALLY_RETURNED_KITS_KEY,
+              NUM_FULLY_RETURNED_KITS_KEY, NUM_KITS_W_PROBLEMS_KEY]
+
+    num_status_keys = get_num_status_keys()
+    result.extend(num_status_keys)
+    return result
+
+
+def get_num_status_keys():
```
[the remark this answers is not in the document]
AmandaBirmingham: Good point.  Changed name to `get_status_num_keys`.

microsetta_private_api/repo/admin_repo.py
AmandaBirmingham: Whoaaaa, I totally did not know about those!  I have rewritten the formatting throughout using these ;)

microsetta_private_api/model/project.py
```diff
@@ -0,0 +1,181 @@
...
+MASS_SPEC_CONTACT_NAME_KEY = 'mass_spec_contact_name'
+MASS_SPEC_CONTACT_EMAIL_KEY = 'mass_spec_contact_email'
+DO_OTHER_KEY = 'do_other'
+BRANDING_ASSOC_INSTRUCTIONS_KEY = 'branding_associated_instructions'
+BRANDING_STATUS_KEY = 'branding_status'
+SUBPROJECT_NAME_KEY = 'subproject_name'
+ALIAS_KEY = 'alias'
+SPONSOR_KEY = 'sponsor'
+COORDINATION_KEY = 'coordination'
+COMPUTED_STATS_KEY = 'computed_stats'
```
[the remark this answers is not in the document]
AmandaBirmingham: I agree that the current structure is unwieldy and will be error-prone if we have to make changes/additions to the field list.  I have created an issue for improving this (#264), since it will take some effort to determine which of the proposed fixes works as intended.

microsetta_private_api/model/project.py
```diff
@@ -0,0 +1,181 @@
...
+        self.mass_spec_contact_name = kwargs.get(MASS_SPEC_CONTACT_NAME_KEY)
+        self.mass_spec_contact_email = kwargs.get(MASS_SPEC_CONTACT_EMAIL_KEY)
+        self.do_other = kwargs.get(DO_OTHER_KEY)
+        self.branding_associated_instructions = kwargs.get(
+            BRANDING_ASSOC_INSTRUCTIONS_KEY)
+        self.branding_status = kwargs.get(BRANDING_STATUS_KEY)
+        self.subproject_name = kwargs.get(SUBPROJECT_NAME_KEY)
+        self.alias = kwargs.get(ALIAS_KEY)
+        self.sponsor = kwargs.get(SPONSOR_KEY)
+        self.coordination = kwargs.get(COORDINATION_KEY)
```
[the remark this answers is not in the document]
AmandaBirmingham: It isn't required that the Project object attribute name match the db field name (the key); it is just that currently they mostly do (with the notable exception of `self.project_name`) because I named them at the same time :)

It IS true that these string values exist again in the yaml definition of the "project" schema.  This is an issue we have throughout this codebase and that we haven't yet figured out how to mitigate ...

microsetta_private_api/repo/admin_repo.py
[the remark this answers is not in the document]
AmandaBirmingham: Thank you!

microsetta_private_api/repo/admin_repo.py
```diff
@@ -16,11 +23,237 @@
...
+    for curr_num_status_key in num_status_keys:
+        colnames.append("{0} bigint".format(curr_num_status_key))
+
+    cases_sql = " ".join(cases)
+    values_sql = ", ".join(values)
+    colnames_sql = ", ".join(colnames)
+
+    final_sql = "{}{}{}{}{}{}{}".format(first_chunk, cases_sql, second_chunk,
+                                        values_sql, third_chunk, colnames_sql,
+                                        fourth_chunk)
```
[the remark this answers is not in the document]
AmandaBirmingham: Improved (I think) with some renaming and use of f strings.

microsetta_private_api/repo/admin_repo.py
```diff
@@ -16,11 +23,237 @@
...
+            USING (barcode)
+            INNER JOIN (
+                SELECT barcode, max(scan_timestamp) AS scan_timestamp
+                FROM barcodes.barcode_scans
+                GROUP BY barcode
+            ) latest_scan
+            ON pb.barcode = latest_scan.barcode
+            GROUP BY project_id, sample_status
+            ORDER BY project_id,
+            CASE sample_status """
```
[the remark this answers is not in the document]
AmandaBirmingham: Thank you for catching this!  See related comment below.

microsetta_private_api/repo/admin_repo.py
```diff
@@ -16,11 +23,237 @@
...
+        FROM ag.ag_kit_barcodes
+        INNER JOIN barcodes.project_barcode
+        USING (barcode)
+        INNER JOIN barcodes.barcode_scans
+        USING (barcode)
+        INNER JOIN (
+            SELECT barcode, max(scan_timestamp) AS scan_timestamp
+            FROM barcodes.barcode_scans
+            GROUP BY barcode
+        ) latest_scan
```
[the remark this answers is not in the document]
AmandaBirmingham: Good point. Please see related comments below.

microsetta_private_api/repo/admin_repo.py
AmandaBirmingham: I have explicitly specified the field names.

microsetta_private_api/repo/admin_repo.py
```diff
@@ -16,11 +23,237 @@
...
+    SELECT project_id,
+    count(distinct barcode) as {0}  --num_samples_received
+    FROM project_barcode
+    INNER JOIN
+    barcode_scans
+    USING (barcode)
+    GROUP BY project_id
+    ORDER BY project_id; """
+
+NUM_FULLY_RECEIVED_KITS_SQL = """
```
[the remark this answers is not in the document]
AmandaBirmingham: Fair enough :) I've rewritten this query; see comments below.

microsetta_private_api/repo/admin_repo.py
AmandaBirmingham: I am hesitant to do this because of a specific issue with nans vs nones that I ran afoul of; see below.

AmandaBirmingham: Done.

microsetta_private_api/repo/admin_repo.py
AmandaBirmingham: Urrgh ... Finally got this working.  It turns out project_df.loc[project_id].to_dict() (i.e., Series.to_dict) *doesn't* convert numpy data types to python datatypes, whereas project_df.to_dict() (i.e., dataframe.to_dict) *does*, so took a bit of tinkering.  At least we now have a test to verify that the type conversion is done correctly.

AmandaBirmingham:
Code-review comments addressed, tests added; no longer WIP.  

@wasade @dhakim87 Below is the corrected query I have for getting the number of kits containing at least one received sample that currently (according to the latest scan_timestamp) isn't valid.  However, it takes 3-4 seconds to run on my dev machine, and my rudimentary efforts at query reformulation/optimization have not helped.  I'm hoping one of you folks will have a suggestion :)

```
        SELECT project_barcode.project_id,
        count(distinct ag_kit_barcodes.ag_kit_id) as num_kits_w_problem
        FROM ag.ag_kit_barcodes
        INNER JOIN barcodes.project_barcode
        USING (barcode)
        INNER JOIN barcodes.barcode_scans
        USING (barcode)
        INNER JOIN (
            SELECT barcode, max(scan_timestamp) AS scan_timestamp
            FROM barcodes.barcode_scans
            GROUP BY barcode
        ) latest_scan
        ON barcode_scans.barcode = latest_scan.barcode
        AND barcode_scans.scan_timestamp = latest_scan.scan_timestamp
        AND barcode_scans.sample_status <> 'sample-is-valid'
        GROUP BY project_barcode.project_id;
```

wasade:
Thanks, @AmandaBirmingham! 3-4 seconds is a long window of time. I just tested on staging and it's about 15s on the staging host. That's certainly a fair bit of time for the database, and we may trend toward HTTP timeouts if there are other long running queries. I unfortunately don't have any experience optimizing SQL queries though. Would it make sense for the nested select to be a view with indices?

AmandaBirmingham:
Agreed re lag time :( :(  Sadly I also know little about optimizing SQL (I already tried my very limited toolbox on this ...). @wasade , do you know anyone knowledgeable in the the extended lab team that we could call in for a consult, or should I wade into learning about views?

dhakim87:
The only part that looks to me like it should be slow is the inner select:

```
SELECT barcode, max(scan_timestamp) AS scan_timestamp
            FROM barcodes.barcode_scans
            GROUP BY barcode
```

If you profile just that statement, are you seeing that take the majority of the time?  If that's not the issue, we're just missing some necessary index and we can figure out which one from an EXPLAIN QUERY PLAN.  If that is the issue... well, SQL can't do that type of query very well - its underlying datastructure (per index) can be thought of as a sorted list per table, and you just can't do that operation using binary search.  

If the inner select is the issue, the simplest way out is duplication of the latest scan information in a separate table.  Annoying to maintain at the repo layer, but you can always instantly grab latest scan info per barcode.

AmandaBirmingham:
@dhakim87 From my manual testing, the "barcode_scans.sample_status <> 'sample-is-valid'" constraint is the one that really kicks it up into "toooo slooooow" territory.  If you are testing with an equal, it is fast, but the <> is slow.   I also tried testing for each of the different non-valid statuses with an OR, but that apparently was handled the same as the <> under the hood (when I did EXPLAIN) and was equally slow ...

I suppose we could go at it obliquely, by getting the number of barcodes in each kit that have been returned and also getting the number of barcodes in each kit that have sample_status = sample-is-valid in their latest scan ... if those two number are NOT the same, then that kit has at least one problem barcode ...

PR #258, opened 2020-08-25
Update sample_is_valid.jinja2

specsdom:
Participants have been receiving the incorrect email template when their sample is received by our lab and is marked as "sample-is-valid" upon being scanned in.

@dhakim87  @wasade


wasade:
Ah, that's weird, good catch. Can the plain text version be updated as well?

specsdom:
> Ah, that's weird, good catch. Can the plain text version be updated as well?

Plain text has the correct email template already!

wasade:
Okay cool

PR #259, opened 2020-08-25
Return account and source IDs with Sample

wasade:
`SampleRepo` and `Sample` now return the samples source ID and account ID. This is useful for diagnostics when examining a full kit as a there may be multiple sources or accounts.

### DIFF
--- a/microsetta_private_api/api/microsetta_private_api.yaml
+++ b/microsetta_private_api/api/microsetta_private_api.yaml
@@ -1558,8 +1558,10 @@ components:
           $ref: '#/components/schemas/sample_projects'
         source_id:
           $ref: '#/components/schemas/source_id'
+          nullable: true
         account_id:
           $ref: '#/components/schemas/account_id'
+          nullable: true
 
     # source section
     source_id:
--- a/microsetta_private_api/api/microsetta_private_api.yaml
+++ b/microsetta_private_api/api/microsetta_private_api.yaml
@@ -1402,6 +1402,11 @@ components:
       type: string
       readOnly: true
       example: "aaaaaaaa-bbbb-cccc-dddd-eeeeffffffff"
+    account_id_nullable:
+      type: string
+      readOnly: true
+      example: "aaaaaaaa-bbbb-cccc-dddd-eeeeffffffff"
+      nullable: true
     account_type: # e.g., regular user or admin--room to grow
       type: string
       readOnly: true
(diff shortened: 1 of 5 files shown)


microsetta_private_api/api/microsetta_private_api.yaml
```diff
@@ -1556,6 +1556,12 @@ components:
           $ref: '#/components/schemas/sample_notes'
         sample_projects:
           $ref: '#/components/schemas/sample_projects'
+        source_id:
+          $ref: '#/components/schemas/source_id'
+          nullable: true
+        account_id:
+          $ref: '#/components/schemas/account_id'
+          nullable: true
```
AmandaBirmingham, line 1564: So, I have pulled down the code and verified for myself that putting these nullables here DOES work (and that leaving them out causes a validation error), but honestly, according to the specs as I understand them, it SHOULDN'T work:

"Any sibling elements of a $ref are ignored. This is because $ref works by replacing itself and everything on its level with the definition it is pointing at."  (see https://swagger.io/docs/specification/using-ref/ ).  

I am feeling worried that connexion (which is doing our validation for us) may be interpreting this yaml in a way that is not valid for the OpenAPI 3 spec.

wasade: That is an *excellent* spot! To be honest, I wasn't sure if it would work specifying `nullable` here, and now that you mention it, I recall encountering that sibling aspect of `$ref` too. I completely agree this should be changed. 

I was hesitant to modify `#/components/schemas/account_id` (and `source_id`) to make them `nullable`. But, should we? Or maybe we introduce a parallel schema `#/components/schemas/account_id_nullable` and use a `oneOf` construct?

wasade:
Thanks, @AmandaBirmingham! I didn't realize that with `/kits` and will add that. I'm unsure how we should handle the openapi spec change but completely agree that we should not merge with the approach as is.

microsetta_private_api/api/tests/test_api.py
```diff
@@ -1260,6 +1266,31 @@ def test_survey_create_success_empty(self):
...
+                                   headers=self.dummy_auth)
+
+        # valid kit but all samples are assigned
+        self.assertEqual(get_resp.status_code, 404)
+
+        get_resp = self.client.get('/api/kits/?language_tag=en_US&'
+                                   'kit_name=%s' % EXISTING_KIT_NAME_2,
+                                   headers=self.dummy_auth)
+
+        # valid kit but all samples are assigned
```
AmandaBirmingham, line 1286: I believe this comment is a cut-and-paste leftover from the test above?  Seems like the code below here is testing the happy-path case where NOT all samples are assigned ...

microsetta_private_api/api/tests/test_api.py
```diff
@@ -1260,6 +1266,31 @@ def test_survey_create_success_empty(self):
 
 @pytest.mark.usefixtures("client")
 class SampleTests(ApiTests):
+    def test_get_unclaimed_samples(self):
```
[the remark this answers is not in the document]
wasade: :+1:

microsetta_private_api/api/tests/test_api.py
```diff
@@ -1260,6 +1266,31 @@ def test_survey_create_success_empty(self):
...
+        get_resp = self.client.get('/api/kits/?language_tag=en_US&'
+                                   'kit_name=%s' % MISSING_KIT_NAME,
+                                   headers=self.dummy_auth)
+        self.assertEqual(get_resp.status_code, 404)
+
+        get_resp = self.client.get('/api/kits/?language_tag=en_US&'
```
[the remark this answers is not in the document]
wasade: :+1:

microsetta_private_api/api/tests/test_api.py
wasade: Ya...

PR #260, opened 2020-08-26
Authentication support for non-browser clients

wasade:
This pull request enables programmatic interaction with the private API, providing a basis to build command line tools like cronjobs.

The authentication requires a user have (a) a valid user account within the microsetta-private-api database (b) a valid and active Authrocket user account with the operating realm (c) a valid Authrocket API key for the realm (d) read/write permission with the API key.

With the above, it is possible to programmatically authenticate against Authrocket to obtain a JWT that is compatible with the subsequent validation machinery of the private API. 

Important: testing here is entirely mocked out as to avoid placing an Authrocket API key into the repository.


wasade:
I'm not sure -- maybe we should defer to our meeting later today. But, my read here, which could very well be incorrect, is that these keys grant a lot of power but are not unlimited. I do not believe it can be used to login as _anyone_ as you still need the correct username as well as password. For example, in order to change an account password, you still need to specify their [existing password](https://authrocket.com/docs/api/users#method-update-password), and for [authenticating](https://authrocket.com/docs/api/users#method-authenticate) as a user you also need to specify their password.

I think the premise that you _could_ login as anyone _if_ you had their password is correct. Given that the password seems to still be required, I don't think this is inherently less secure than someone guessing emails / passwords on the Authrocket login page.

However, this key would allow an actor to _delete_ user accounts among some other actions. Leaking could certainly create chaos. 

What was hoping to do is use something like the [LoginRocket V1 API](https://v1.authrocket.com/docs/api/loginrocket#method-login) but it does not appear to be supported from what I can see for Authrocket. I don't think this would require any additional keys. I'll shoot them an email to see if something like this is feasible.

wasade:
I sent an inquiry out to Authrocket on this, thanks again @dhakim87!

microsetta_private_api/api/_account.py
```diff
@@ -123,17 +123,32 @@ def update_account(account_id, body, token_info):
...
-    try:
-        token_info = jwt.decode(token,
-                                AUTHROCKET_PUB_KEY,
-                                algorithms=["RS256"],
-                                verify=True,
-                                issuer="https://authrocket.com")
-    except InvalidTokenError as e:
-        raise(Unauthorized(INVALID_TOKEN_MSG, e))
+    token_info = None
+    for scheme in JWT_SCHEMES:
```
dhakim87, line 142: Possibly unnecessary to loop over schemes - we can always determine which jwt public key should be used depending on the route the request came in from.  Not a security flaw afaict, since we verify issuer in the token and account table, just unnecessary work being done.

wasade: I wasn't able to get that to work. While openapi allows for multiple schemas (see https://swagger.io/docs/specification/authentication/) to be expressed as a logical "OR", I could not get connexion to actually hit different or multiple operationIds. I wasn't sure if it was something I was doing wrong or if connexion was doing it wrong.

microsetta_private_api/api/microsetta_private_api.yaml
```diff
@@ -1827,14 +1827,13 @@ components:
...
       bearerFormat: JWT
-      x-bearerInfoFunc: microsetta_private_api.api.verify_authrocket
-
+      x-bearerInfoFunc: microsetta_private_api.api.verify_jwt
 
 # END COMPONENTS
 # This enables the authRocket securityScheme defined in components for all routes
 security:
-  - authRocket: []
+  - jwt: []
```
dhakim87, line 1839: Future option: I believe if you wanted to limit what the cron job was capable of, you could make it a separate security scheme and specify it on only those routes it was allowed to call.  Appears this could be done through openAPI rather than the account_type field as I was thinking about.  Potentially you wouldn't even need the cronjob to have an associated account in the account table with that type of solution.

wasade: Ish. Most (all?) endpoints also call [`_validate_account_access`](https://github.com/biocore/microsetta-private-api/blob/ad506b97b61bc6d809681521cc2435ec5143ddfa/microsetta_private_api/api/_account.py#L154`), so some modification there would be necessary

wasade:
Thanks, @dhakim87! Ya the looping was pragmatic not ideal. But, with such a small number of schemes, and not foreseeing a reason that will grow, I think this is probably reasonable enough.

PR #261, opened 2020-08-26
WIP: sending email via celery

wasade:
This PR adds in a broker-based async compute framework to offload long running tasks to separate processes. As a proof-of-concept, the send email functionality is now through a separate worker. The ability to run tasks like this is related to #260 in that it would be nice if (a) a cronjob could tickle the API to do something such as pull information from vioscreen and (b) to not block a web server process for something that may take many minutes or longer to run.

What's nice about celery is it's possible to just call the function like a regular Python function. So right now, the test implemented here is not relying on the async methods (as we're implicitly assuming celery does that stuff correctly).

Nothing here is set in stone. This is one way we can offload long running compute, and we can explore alternatives that is the consensus.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,10 @@ In the activated conda environment, initialize a test database:
 
 `python microsetta_private_api/LEGACY/build_db.py`
 
+Then we'll initiate a Celery worker:
+
+`celery -A celery_worker.celery worker --loglevel=info`
+
 Next, start the microservice using flask's built-in server by running, e.g., 
 
 `python ./microsetta_private_api/server.py`
--- a/README.md
+++ b/README.md
@@ -26,6 +26,11 @@ Install the JSON Web Tokens library with cryptography support:
 
 `pip install pyjwt[crypto]`
 
+Install Redis and Celery with Redis support for out-of-band compute:
+
+`conda install redis`
+`pip install celery[redis]`
+
 Then install the microsetta-private-api in editable mode:
 
 `pip install -e .`
--- a/celery_worker.py
+++ b/celery_worker.py
@@ -1,0 +1,3 @@
+from microsetta_private_api.server import build_app
+from microsetta_private_api.celery_utils import celery  # noqa
+app = build_app()
--- a/celery_worker.py
+++ b/celery_worker.py
@@ -1,3 +1,3 @@
-from microsetta_private_api.server import build_app
-from microsetta_private_api.celery_utils import celery  # noqa
-app = build_app()
+from microsetta_private_api.server import app
+from microsetta_private_api.celery_utils import celery, init_celery
+init_celery(celery, app.app)
--- a/microsetta_private_api/admin/admin_impl.py
+++ b/microsetta_private_api/admin/admin_impl.py
@@ -256,8 +256,7 @@ def send_email(body, token_info):
         template_args['resolution_url'] = resolution_url
         template_args['contact_name'] = contact_name
         celery_send_email.apply_async(args=[email, template_name,
-                                            template_args],
-                                      ignore_result=True)
+                                            template_args])
 
         # Add an event to the log that we sent this email successfully
         event = LogEvent(
--- a/microsetta_private_api/admin/admin_impl.py
+++ b/microsetta_private_api/admin/admin_impl.py
@@ -255,7 +255,7 @@ def send_email(body, token_info):
         template_args = dict(body['template_args'])
         template_args['resolution_url'] = resolution_url
         template_args['contact_name'] = contact_name
-        celery_send_email.apply_async(args=[email, template_name, 
+        celery_send_email.apply_async(args=[email, template_name,
                                             template_args],
                                       ignore_result=True)
 
--- a/microsetta_private_api/admin/admin_impl.py
+++ b/microsetta_private_api/admin/admin_impl.py
@@ -255,8 +255,9 @@ def send_email(body, token_info):
         template_args = dict(body['template_args'])
         template_args['resolution_url'] = resolution_url
         template_args['contact_name'] = contact_name
-        celery_send_email.delay(email, template_name, template_args,
-                                ignore_result=True)
+        celery_send_email.apply_async(args=[email, template_name, 
+                                            template_args],
+                                      ignore_result=True)
 
         # Add an event to the log that we sent this email successfully
         event = LogEvent(
--- a/microsetta_private_api/admin/admin_impl.py
+++ b/microsetta_private_api/admin/admin_impl.py
@@ -2,7 +2,6 @@ import uuid
 from flask import jsonify
 import datetime
 
-from microsetta_private_api.admin.email_templates import EmailMessage
 from microsetta_private_api.config_manager import SERVER_CONFIG
 from microsetta_private_api.model.log_event import LogEvent
 from microsetta_private_api.exceptions import RepoException
@@ -10,7 +9,8 @@ from microsetta_private_api.repo.account_repo import AccountRepo
 from microsetta_private_api.repo.event_log_repo import EventLogRepo
 from microsetta_private_api.repo.transaction import Transaction
 from microsetta_private_api.repo.admin_repo import AdminRepo
-from microsetta_private_api.util.email import SendEmail
+from microsetta_private_api.tasks import send_email as celery_send_email
+from microsetta_private_api.admin.email_templates import EmailMessage
 from microsetta_private_api.util.redirects import build_login_redirect
 from werkzeug.exceptions import Unauthorized, BadRequest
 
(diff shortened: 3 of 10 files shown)


celery_worker.py
```diff
@@ -0,0 +1,3 @@
+from microsetta_private_api.server import build_app
+from microsetta_private_api.celery_utils import celery  # noqa
+app = build_app()
```
dhakim87, line 3: What?  Why do workers running tasks from redis need copies of the Flask app?  Does this make a new Flask app?  or is it trying to grab a singleton?

microsetta_private_api/tasks.py
```diff
@@ -0,0 +1,9 @@
+from microsetta_private_api.celery_utils import celery
+from microsetta_private_api.util.email import SendEmail
+from microsetta_private_api.admin.email_templates import EmailMessage
+
+
+@celery.task()
```
dhakim87, line 6: This is super convenient, I like it.  This wraps tasks in some kind of Future type object where we can specify how to call the function I guess?

microsetta_private_api/server.py
```diff
@@ -68,6 +70,9 @@ def build_app():
         app.app.logger.handlers = gunicorn_logger.handlers
         app.app.logger.setLevel(gunicorn_logger.level)
 
+    init_celery(celery, app.app)
+    celery.autodiscover_tasks([__name__.split('.')[0]])
```
dhakim87, line 74: The __name__ black magic just grabs "microsetta_private_api" and it scans recursively from there for anything with a @celery.task?  This seems very clean.

celery_worker.py
wasade: The Flask pattern is to associate the application context with celery. See [here](https://flask.palletsprojects.com/en/1.1.x/patterns/celery/). But, I confess, I don't fully understand why. It is likely this could be refactored so it's not a full `build_app` call as I think only the `init_celery` call is needed.

microsetta_private_api/tasks.py
wasade: Ya exactly. So usually you would call `.delay()` and get a `Future` back. There are a few ways these can be [used](https://docs.celeryproject.org/en/stable/userguide/tasks.html) including with timeouts, scheduling, keeping results, ignoring results, callbacks, retries, on failure / success, etc.

microsetta_private_api/server.py
wasade: TBH, I was super annoyed with this :D Celery should automatically find modules named `tasks`, but the worker was not doing that, it wasn't obvious to me why, so I decided to force it. 

In this case, the value of `__name__` is `microsetta_private_api.server` as it refers to the name of the current python module. The split is performed so that we can extract the `microsetta_private_api` piece. And then the call to `celery.autodiscover_tasks` will look for, in this case, `microsetta_private_api.tasks`.

wasade:
Thanks, @dhakim87! So far it seems pleasant other than getting the workers to find the tasks. Returning data to users isn't bad as there is an async `ready`, a `get` which supports timeout among other [things](https://docs.celeryproject.org/en/stable/reference/celery.result.html#celery.result.AsyncResult.get), error checking, etc etc.  

You might want to take a look at the ways in which celery `tasks` can be linked as there is quite a bit of flexibility, including partial argument support, chaining (in a a few ways), and more. There is a pleasant use of operator overloading here, so linking successive tasks is basically `foo.s(blah) | bar.s(other) | baz.s(more_stuff)`, where the `.s` is shorthand for providing the function signature and `|` chains two tasks together. More info on that can be found [here](https://docs.celeryproject.org/en/stable/userguide/canvas.html).

README.md
```diff
@@ -26,6 +26,11 @@ Install the JSON Web Tokens library with cryptography support:
 
 `pip install pyjwt[crypto]`
 
+Install Redis and Celery with Redis support for out-of-band compute:
+
+`conda install redis`
```
dhakim87, line 31: Do we consider the redis installations for private-api and interface to be separate redis servers or the same redis server?  Same feels convenient, but we would need to be careful of overlapping key names I imagine.

microsetta_private_api/admin/admin_impl.py
```diff
@@ -249,13 +249,14 @@ def send_email(body, token_info):
...
+        template = EmailMessage[template_name]
+
         template_args = dict(body['template_args'])
         template_args['resolution_url'] = resolution_url
         template_args['contact_name'] = contact_name
-
-        # Send the email
-        SendEmail.send(email, template, template_args)
+        celery_send_email.delay(email, template_name, template_args,
+                                ignore_result=True)
```
dhakim87, line 259: I'm not 100% clear on the sending of ignore_result- I don't believe celery_send_email has a named parameter called ignore_result.  I think if ignore_result is an argument to celery, rather than an argument to the celery_send_email task, you have to instead call the much uglier 

`celery_send_email.apply_async((email, template_name, template_args), kwargs={}, ignore_result=True)`

That's my interpretation from https://docs.celeryproject.org/en/stable/userguide/calling.html  
Is that delay call truly equivalent, or is the ignore_result named parameter disappearing into the ether in the current implementation?

microsetta_private_api/celery_utils.py
```diff
@@ -0,0 +1,31 @@
...
+def init_celery(celery, app):
+    celery.conf.update(app.config)
+    TaskBase = celery.Task
+
+    class ContextTask(TaskBase):
+        def __call__(self, *args, **kwargs):
+            with app.app_context():
+                return TaskBase.__call__(self, *args, **kwargs)
+
+    celery.Task = ContextTask
```
dhakim87, line 21: This feels funky to me - I think that associating tasks with flask app contexts has implications that I don't really understand.

I believe the app context keeps track of the current request- that makes sense for synchronous operations, but for asynchronous operations, I'm confused as to what this means.  Does this keep TCP sockets open until all tasks started within a particular context are completed?  If not, what will the app context contain when it is queried from a particular async task long after the user has disconnected from our server?  

What breaks if we don't make this link?  What becomes possible if we do?  

Finally, are we specifically overwriting celery.Task because other modules know to look for the task base class at this location, or are we creating tasks ourselves?  If the second, I'd prefer we not overwrite celery.Task and instead use celery.ContextTask so that we can be explicit as to whether or not we want a flask context on our operations.

README.md
wasade: Doesn't matter, but it would be very surprising for the keynames used to overlap.

microsetta_private_api/admin/admin_impl.py
wasade: That is a very good spot. I'll change to `apply_async`. But, does suggest the args do not need to be wrapped in a tuple, and there should not be any need to explicitly provide an empty dict to `kwargs`.

wasade: https://docs.celeryproject.org/en/stable/userguide/calling.html#results-options

microsetta_private_api/celery_utils.py
wasade: We're following the guidance from Flask for use with Celery (https://flask.palletsprojects.com/en/1.1.x/patterns/celery/). I admit, I'm not entirely sure _why_ this is recommended, but it is the standard pattern. 

I don't see how Celery would know that `celery.task()` refers to `celery.ContextTask`, though we can probably figure it out. But, while I'm not excited about not fully grokking the interaction here, I'm much less excited about deviating from the recommended pattern without knowing why a deviation is necessary. 

Would you be alright proceeding with the recommended pattern until we discover a reason to change it?

README.md
dhakim87: May be worth prefixing the keys we use in some wrapper - "MICROSETTA_PRIVATE_" + keyname, if we're at all worried.  Not relevant to this commit though

microsetta_private_api/admin/admin_impl.py
dhakim87:
```
    def apply_async(self, args=None, kwargs=None, task_id=None, producer=None,
                    link=None, link_error=None, shadow=None, **options):
```

I think that's a typo in the docs.  Every other apply_async call on that page uses a tuple for args.  Unless there's even more black magic under the hood, apply_async(2,2) should set args to 2 and kwargs to 2, which is a very strange value to pass with **kwargs.  

But based on my vars() fiasco, I clearly don't know how python works under the hood :D

microsetta_private_api/celery_utils.py
dhakim87: Yep yep, no problem with it - won't impact us at all in our current usage.

README.md
wasade: I'd only be concerned if we had multiple celery apps using the same redis server. It looks like key prefixes are not fully supported yet in Celery (see https://github.com/celery/kombu/issues/853) but changing the queue name effectively solves this problem (see https://github.com/celery/kombu/issues/853#issuecomment-385252709). I'll set the queue name to 'microsetta_private_api' and that should solve this concern.

microsetta_private_api/admin/admin_impl.py
wasade: i just did `args=[foo, bar]`. It's explicit

microsetta_private_api/celery_utils.py
wasade: Great, thanks!!

wasade:
Just the queue item left i think. Will do in about 30 mins

PR #266, opened 2020-09-23
Designate active projects

AmandaBirmingham:
At Edgar's request, adds a field that designates each project as either active or not; default is active.  Adds an optional parameter for this value to endpoint that returns project list so that user can get project list limited to only active ones, only inactive ones, or not limited at all.  Filtering is done on the pandas dataframe in admin_repo rather than in sql so I didn't have to add a new optional parameter to every project statistics sql or else re-evaluate the joins.

This PR is necessary for the "Alter manage project page to allow filtering by project active status" PR in microsetta-admin.


microsetta_private_api/repo/admin_repo.py
```diff
@@ -579,13 +581,30 @@ def delete_project_by_name(self, project_name):
                         (project_name,))
             return cur.rowcount == 1
 
-    def get_projects(self):
-        """Return a list of Project objects, ordered by project id."""
+    def get_projects(self, is_active_val=None):
+        """Return a list of Project objects, ordered by project id.
+
```
wasade, line 586: Extra newline

wasade (changes requested):
One super minor comment

microsetta_private_api/repo/admin_repo.py
AmandaBirmingham: Fixed now :)

PR #267, opened 2020-09-26
WIP - Vioscreen Report PDF

dhakim87:
Adds support for talking to vioscreen using RPC through celery.  Have to worry about both the RPC channel only allowing jsonifiable data, handling async correctly, and dealing with content types and pdf embedding, but it all works as a prototype.  Note that the initial vioscreen login handshake takes about 35 seconds (as presumably does any token refresh), but once logged in the pdf generation time is near instant.  Must check if vioscreen meters our pdf generation and we're forced to cache results.

Uncovered a potential issue where survey ids in developer builds are referring to both valid vioscreen data and primary survey data, which is not supported by our model classes/repos.

### DIFF
--- a/microsetta_private_api/api/_survey.py
+++ b/microsetta_private_api/api/_survey.py
@@ -1,7 +1,5 @@
-import io
-
 import flask
-from flask import jsonify, make_response, send_file
+from flask import jsonify, make_response
 from werkzeug.exceptions import NotFound
 
 from microsetta_private_api.api._account import \
@@ -239,7 +237,9 @@ def top_food_report(account_id, source_id, survey_id, token_info):
         vioscreen_repo = VioscreenRepo(t)
 
         # Vioscreen username is our survey_id
-        status = vioscreen_repo.get_vioscreen_status(account_id, source_id, survey_id)
+        status = vioscreen_repo.get_vioscreen_status(account_id,
+                                                     source_id,
+                                                     survey_id)
         if status != 3:
             # Oops, we don't have results available for this one
             raise NotFound("No such survey recorded")
--- a/microsetta_private_api/api/_survey.py
+++ b/microsetta_private_api/api/_survey.py
@@ -1,5 +1,8 @@
+import io
+
 import flask
-from flask import jsonify
+from flask import jsonify, make_response, send_file
+from werkzeug.exceptions import NotFound
 
 from microsetta_private_api.api._account import \
     _validate_account_access
@@ -10,6 +13,7 @@ from microsetta_private_api.repo.survey_template_repo import SurveyTemplateRepo
 from microsetta_private_api.repo.transaction import Transaction
 from microsetta_private_api.repo.vioscreen_repo import VioscreenRepo
 from microsetta_private_api.util import vioscreen, vue_adapter
+from microsetta_private_api.util.vioscreen import VioscreenAdminAPI
 
 
 def read_survey_templates(account_id, source_id, language_tag, token_info):
@@ -226,3 +230,31 @@ def _submit_vioscreen_status(account_id, source_id, info_str):
                                     source_id,
                                     vio_info["username"])
     return response
+
+
+def top_food_report(account_id, source_id, survey_id, token_info):
+    _validate_account_access(token_info, account_id)
+
+    with Transaction() as t:
+        vioscreen_repo = VioscreenRepo(t)
+
+        # Vioscreen username is our survey_id
+        status = vioscreen_repo.get_vioscreen_status(account_id, source_id, survey_id)
+        if status != 3:
+            # Oops, we don't have results available for this one
+            raise NotFound("No such survey recorded")
+
+        vio = VioscreenAdminAPI()
+        sessions = vio.sessions(survey_id)
+        # Looks like vioscreen supports multiple sessions per user, do we care?
+        session_id = sessions[0]['sessionId']
+        report = vio.top_food_report(session_id)
+
+        response = make_response(report)
+        response.headers.set("Content-Type", "application/pdf")
+        # TODO: Do we want it to download a file or be embedded in the html?
+        # response.headers.set('Content-Disposition',
+        #                      'attachment',
+        #                      filename='top-food-report.pdf')
+
+        return response
--- a/microsetta_private_api/api/microsetta_private_api.yaml
+++ b/microsetta_private_api/api/microsetta_private_api.yaml
@@ -681,6 +681,32 @@ paths:
     # NB: NO "put"--you shouldn't update an answered survey, but instead should take a new one
     # NB: Currently no delete ... can dissociate a survey from a sample but not delete the survey itself
 
+  '/accounts/{account_id}/sources/{source_id}/surveys/{survey_id}/reports/topfoodreport':
+    # Attempts to generate the vioscreen top food report pdf from this vioscreen survey
+    get:
+      # TODO: Unsure if this should be get or post- it's post on vioscreen, but
+      # we can't easily embed the pdf if the link is a post.  May need to post
+      # and cache results so we can make it a get for embedding.
+      operationId: microsetta_private_api.api.top_food_report
+      tags:
+        - Survey
+      summary: Generate vioscreen top food report
+      description:  Generate vioscreen top food report
+      parameters:
+        - $ref: '#/components/parameters/account_id'
+        - $ref: '#/components/parameters/source_id'
+        - $ref: '#/components/parameters/survey_id'
+        - $ref: '#/components/parameters/language_tag'
+      responses:
+        '200':
+          description: Top Food Report pdf
+        '401':
+          $ref: '#/components/responses/401Unauthorized'
+        '403':
+          $ref: '#/components/responses/403Forbidden'
+        '404':
+          $ref: '#/components/responses/404NotFound'
+
   '/accounts/{account_id}/sources/{source_id}/samples':
     get:
       operationId: microsetta_private_api.api.read_sample_associations
--- a/microsetta_private_api/repo/survey_answers_repo.py
+++ b/microsetta_private_api/repo/survey_answers_repo.py
@@ -49,18 +49,19 @@ class SurveyAnswersRepo(BaseRepo):
             #  in dev builds)
             # TODO:  It's even worse than I feared.  Some of our primary
             #  surveys are ALSO vioscreen surveys with the same key. wtf.
-            # if len(rows) == 0:
-            vioscreen_repo = VioscreenRepo(self._transaction)
-            status = vioscreen_repo._get_vioscreen_status(survey_answers_id)
-            if status is not None:
-                return SurveyTemplateRepo.VIOSCREEN_ID
-            else:
-                return None
-                # TODO: Maybe this should throw an exception, but doing so
-                #  locks the end user out of the minimal implementation
-                #  if they submit an empty survey response.
-                # raise RepoException("No answers in survey: %s" %
-                #                     survey_answers_id)
+            if len(rows) == 0:
+                vioscreen_repo = VioscreenRepo(self._transaction)
+                status = vioscreen_repo._get_vioscreen_status(
+                    survey_answers_id)
+                if status is not None:
+                    return SurveyTemplateRepo.VIOSCREEN_ID
+                else:
+                    return None
+                    # TODO: Maybe this should throw an exception, but doing so
+                    #  locks the end user out of the minimal implementation
+                    #  if they submit an empty survey response.
+                    # raise RepoException("No answers in survey: %s" %
+                    #                     survey_answers_id)
 
             arbitrary_question_id = rows[0][1]
             cur.execute("SELECT surveys.survey_id FROM "
--- a/microsetta_private_api/repo/survey_answers_repo.py
+++ b/microsetta_private_api/repo/survey_answers_repo.py
@@ -44,18 +44,23 @@ class SurveyAnswersRepo(BaseRepo):
                         (survey_answers_id,))
             rows += cur.fetchall()
 
-            if len(rows) == 0:
-                vioscreen_repo = VioscreenRepo(self._transaction)
-                status = vioscreen_repo.get_vioscreen_status(survey_answers_id)
-                if status is not None:
-                    return SurveyTemplateRepo.VIOSCREEN_ID
-                else:
-                    return None
-                    # TODO: Maybe this should throw an exception, but doing so
-                    #  locks the end user out of the minimal implementation
-                    #  if they submit an empty survey response.
-                    # raise RepoException("No answers in survey: %s" %
-                    #                     survey_answers_id)
+            # TODO:  We probably can't merge the PR with this change, need
+            #  a policy to resolve these survey ids (unless they only happen
+            #  in dev builds)
+            # TODO:  It's even worse than I feared.  Some of our primary
+            #  surveys are ALSO vioscreen surveys with the same key. wtf.
+            # if len(rows) == 0:
+            vioscreen_repo = VioscreenRepo(self._transaction)
+            status = vioscreen_repo._get_vioscreen_status(survey_answers_id)
+            if status is not None:
+                return SurveyTemplateRepo.VIOSCREEN_ID
+            else:
+                return None
+                # TODO: Maybe this should throw an exception, but doing so
+                #  locks the end user out of the minimal implementation
+                #  if they submit an empty survey response.
+                # raise RepoException("No answers in survey: %s" %
+                #                     survey_answers_id)
 
             arbitrary_question_id = rows[0][1]
             cur.execute("SELECT surveys.survey_id FROM "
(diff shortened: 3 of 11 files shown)


microsetta_private_api/util/vioscreen.py
```diff
@@ -79,3 +86,214 @@ def decode_key(encoded):
...
+        method = self.session.get
+    elif method == "POST":
+        method = self.session.post
+    else:
+        raise Exception("Unknown Method")
+
+    url = urljoin(self.baseurl, url)
+
+    req = None
+    for try_twice in range(2):
```
gwarmstrong, line 162:
```suggestion
    for _ in range(2):
```

microsetta_private_api/util/vioscreen.py
```diff
@@ -79,3 +86,214 @@ def decode_key(encoded):
...
+        req = method(url, headers=self.headers, **kwargs)
+        if req.status_code != 200:
+            data = req.json()
+            code = data.get('Code')
+
+            if code == 1016:
```
gwarmstrong, line 168: Where do these codes come from?

microsetta_private_api/util/vioscreen.py
gwarmstrong: The logic of why you are doing this is a loop is a little confusing.

What about moving things to a function that returns `(data, needs_retry)`

Then something like:
```
data, needs_retry = get_pdf(...)
if needs_retry:
    self.update_headers()
    data, needs_retry = get_pdf(...)
```

microsetta_private_api/util/vioscreen.py
```diff
@@ -79,3 +86,214 @@ def decode_key(encoded):
...
+    def _get_session_data(self, id_, name):
+        url = 'sessions/%s/%s' % (id_, name)
+        result = self.get(url)
+        if 'error' in result:
+            return {}
+        else:
+            return result
+
+    def foodcomponents(self, id_):
+        return self._get_session_data(id_, 'foodcomponents').get('data')
```
gwarmstrong, line 232: Feels weird that you cannot get all this info with a single RPC. Is there not a way to query all of this info at once?

gwarmstrong (commented):
Looks good. A few comments.

ALSO: looks like you have failing tests, can you address those?

microsetta_private_api/api/_survey.py
```diff
@@ -226,3 +228,33 @@ def _submit_vioscreen_status(account_id, source_id, info_str):
...
+        status = vioscreen_repo.get_vioscreen_status(account_id,
+                                                     source_id,
+                                                     survey_id)
+        if status != 3:
+            # Oops, we don't have results available for this one
+            raise NotFound("No such survey recorded")
+
+        vio = VioscreenAdminAPI()
+        sessions = vio.sessions(survey_id)
+        # Looks like vioscreen supports multiple sessions per user, do we care?
```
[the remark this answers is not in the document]
wasade: I don't believe we're relying on multiple sessions

microsetta_private_api/api/_survey.py
```diff
@@ -226,3 +228,33 @@ def _submit_vioscreen_status(account_id, source_id, info_str):
...
+    with Transaction() as t:
+        vioscreen_repo = VioscreenRepo(t)
+
+        # Vioscreen username is our survey_id
+        status = vioscreen_repo.get_vioscreen_status(account_id,
+                                                     source_id,
+                                                     survey_id)
+        if status != 3:
+            # Oops, we don't have results available for this one
+            raise NotFound("No such survey recorded")
```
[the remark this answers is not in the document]
dhakim87: The NotFound will bubble up to Flask's error handlers and throw a 404 for the requested route.  It should be clear from the context of the url what the account id, source id and survey id are if a user hits the page (and will also be ignored at the moment by the microsetta-interface project since no user actually hits these routes directly).  Unless/until we move to a system with application layer error codes that can be parsed and handled in code, I don't see a reason to extend the 404 beyond a string I can grep for if it appears in an error report.  

Would punt for now.

microsetta_private_api/api/microsetta_private_api.yaml
```diff
@@ -681,6 +681,32 @@ paths:
...
+      summary: Generate vioscreen top food report
+      description:  Generate vioscreen top food report
+      parameters:
+        - $ref: '#/components/parameters/account_id'
+        - $ref: '#/components/parameters/source_id'
+        - $ref: '#/components/parameters/survey_id'
+        - $ref: '#/components/parameters/language_tag'
+      responses:
+        '200':
+          description: Top Food Report pdf
```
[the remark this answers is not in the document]
dhakim87: Definitely.  I spent 20 minutes and couldn't find the magic sauce to make yaml stop complaining.  I'll give this a shot.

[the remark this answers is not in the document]
dhakim87: Yep, nope.  Connexion explodes trying to verify the pdf response with these settings on output.  I'm sure there is a proper schema that can be assigned, but I have no idea how to do it.

microsetta_private_api/api/_survey.py
```diff
@@ -226,3 +228,33 @@ def _submit_vioscreen_status(account_id, source_id, info_str):
...
+
+        vio = VioscreenAdminAPI()
+        sessions = vio.sessions(survey_id)
+        # Looks like vioscreen supports multiple sessions per user, do we care?
+        session_id = sessions[0]['sessionId']
+        report = vio.top_food_report(session_id)
+
+        response = make_response(report)
+        response.headers.set("Content-Type", "application/pdf")
+        # TODO: Do we want it to download a file or be embedded in the html?
```
wasade, line 255: Not sure of the pros/cons here

microsetta_private_api/repo/survey_answers_repo.py
```diff
@@ -44,18 +44,23 @@ def find_survey_template_id(self, survey_answers_id):
...
-                if status is not None:
-                    return SurveyTemplateRepo.VIOSCREEN_ID
-                else:
-                    return None
-                    # TODO: Maybe this should throw an exception, but doing so
-                    #  locks the end user out of the minimal implementation
-                    #  if they submit an empty survey response.
-                    # raise RepoException("No answers in survey: %s" %
-                    #                     survey_answers_id)
+            # TODO:  We probably can't merge the PR with this change, need
```
wasade, line 47: should we put in a patch in a separate PR to resolve this?

microsetta_private_api/repo/survey_answers_repo.py
```diff
@@ -44,18 +44,23 @@ def find_survey_template_id(self, survey_answers_id):
...
+            # if len(rows) == 0:
+            vioscreen_repo = VioscreenRepo(self._transaction)
+            status = vioscreen_repo._get_vioscreen_status(survey_answers_id)
+            if status is not None:
+                return SurveyTemplateRepo.VIOSCREEN_ID
+            else:
+                return None
+                # TODO: Maybe this should throw an exception, but doing so
+                #  locks the end user out of the minimal implementation
+                #  if they submit an empty survey response.
```
wasade, line 61: Do we have a notion of an empty survey response here?

microsetta_private_api/util/vioscreen.py
```diff
@@ -79,3 +86,221 @@ def decode_key(encoded):
...
+class VioscreenAdminAPITask(Task):
+    def __init__(self):
+        print("Initializing Vioscreen Communications Task")
+        self.baseurl = urljoin('https://api.viocare.com',
+                               SERVER_CONFIG["vioscreen_regcode"]) + '/'
+        self.user = SERVER_CONFIG["vioscreen_admin_username"]
+        self.password = SERVER_CONFIG["vioscreen_admin_password"]
+        self.session = requests.Session()
+        self.headers = None
+        # TODO:  Rage.  The flask side instantiates this task upon doing RPC
```
wasade, line 112: I *think* it's getting instantiated on the `@celery.task()` decorator when this file is imported

wasade: One alternative that would (a) avoid the instance getting triggered by flask on module import and (b) be process specific is below:

```python
VioscreenAdminAPITask_singleton = None

@celery.task(
    ignore_result=False,  # We need the results back
    default_retry_delay=5  # Default policy: wait 5 seconds on failure
)
def make_vioscreen_request(method, url, **kwargs):
    global VioscreenAdminAPITask_singleton
    if VioscreenAdminAPITask_singleton is None:
        VioscreenAdminAPITask_singleton = VioscreenAdminAPITask()
    ... do stuff ...
```

microsetta_private_api/util/vioscreen.py
wasade: trial and error

microsetta_private_api/util/vioscreen.py
wasade: Not with the vioscreen API, at least that im aware of

microsetta_private_api/util/vioscreen.py
```diff
@@ -79,3 +86,221 @@ def decode_key(encoded):
...
+    def dietaryscore(self, id_):
+        return self._get_session_data(id_, 'dietaryscore').get('dietaryScore')
+
+    def users(self):
+        return self.get('users')['users']
+
+    def sessions(self, vioscreen_user):
+        detail = self.get('users/%s/sessions' % vioscreen_user)
+
+        # TODO: Can we push waiting for the async result out to flask to
```
wasade, line 263: I think we should just have the client poll. I'm much more concerned about blocked webserver processes than the load from polling

microsetta_private_api/util/vioscreen.py
dhakim87: Agree...  I was basically using the continue as a one time GOTO which is always super clunky.  I've moved out the request handling to a separate inner function and split out the update_headers as you indicate, which does clean up the logic a little bit.  I was also thinking I could mess with the retry delay based on which error code came back from vioscreen, so that auth errors retry without delay, but if this mechanism works well enough, I don't see a reason to atm.

microsetta_private_api/util/vioscreen.py
```diff
@@ -79,3 +86,221 @@ def decode_key(encoded):
...
+            data = f(session_id)
+            if data is None:
+                errors.append("FFQ appears incomplete or not taken")
+                break
+
+            results[name] = data
+
+        return errors, results
+
+    def top_food_report(self, session_id):
```
wasade, line 294: should we store the result in the database so we only fetch from vioscreen once?

wasade:
Thanks, @dhakim87!! I think this looks super cool. I completely get that it is not ideal on the async side of things, but it doesn't look that bad in my opinion if we tolerate a little bit of polling

microsetta_private_api/util/vioscreen.py
dhakim87: Vioscreen application layer error codes.  I don't know that we have a full list of them or if these were just the ones we encountered doing initial development against their apis.

microsetta_private_api/util/vioscreen.py
dhakim87: Our functions just match the swagger api of the vioscreen website (https://api.viocare.com/ - can send creds or show you their api out of band).  I don't see any reason to diverge from their api here, and I don't know that they have a bulk download url implemented anyway.

microsetta_private_api/api/_survey.py
dhakim87: If we want to show the user their pdf in the browser, the current implementation is correct.  If we want the user to download the pdf directly to their computer, the second option would be more correct (though it hardly matters since this gets routed through microsetta-interface anyway, the response headers set in that project are the ones that actually determine what happens)

microsetta_private_api/repo/survey_answers_repo.py
dhakim87: Yes please!  I consider this the main reason to hold back this PR: none of our routes are capable of saying that a survey is both a primary survey and a vioscreen survey, but that occurs in the dev db.  If that's also happening on staging/production, we either need to change our model to add a legacy combined survey type, or we need a schema change to split into two identifiers.

microsetta_private_api/repo/survey_answers_repo.py
dhakim87: iirc, it's difficult or impossible to determine what survey was taken if the user sends back an empty response in our current database models.  I shifted code around to make surveys that are both primary and vioscreen return as vioscreen rather than primary for testing purposes, and probably mucked up the empty survey handling in the process.  But I want to remove all of that and have us externally resolve the surveys that are both primary and vioscreen because I don't know what to do with the logic here.

microsetta_private_api/util/vioscreen.py
dhakim87: afaict, flask instantiates the VioscreenAdminAPITask when make_vioscreen_request is called for the first time.  It should be totally unnecessary to do that, since it's unclear to me how that object would be synchronized with the one in the celery process, but that appears to be the library's behavior.  I don't really care all too much about stopping it from having that behavior- I just can't let flask's workers run off and make requests to vioscreen, so I had to remove the update_headers call from the constructor.  Frankly I think making web requests within the constructor has the potential for badness anyway.  

I think the real issue to address here is on what events should we be refreshing vioscreen credentials on the celery side.  Right now we do it whenever someone makes a request and our auth isn't set or fails, but we could theoretically have a timer that triggers a refresh a few minutes prior to our tokens expiring.  That would give a much better experience to the first poor user who makes vioscreen requests after the token expires, since they wouldn't be waiting for the authentication to go through

microsetta_private_api/util/vioscreen.py
dhakim87: I think polling feels like a trap.  It trades cpu, network usage, and increased complexity on the javascript, microsetta-interface, and private-api all because we're trying to mix two libraries with incompatible approaches.  Suddenly embed tags and pdf downloads don't behave in a standard way and the client browser must have javascript enabled.  It adds extra complexity in the client pages and extra complexity in either redis or postgres to store which account made which celery request so end users can wait on them.  

The celery api's are set up for async but flask doesn't have an event loop to make use of it.  I'd prefer we either bite the bullet and go full async, or accept flask's limitations and block.  It's provable that blocking here can't cause deadlocks since celery can't start new flask requests.  So the only issue we run into is if we exhaust the pool of flask workers.  But we have control over how many flask workers are started, and can even configure it to start workers per request if we're worried.  It's a waste of memory allocating all those empty stacks for new threads, but it seems to be the way flask is designed.  

If these requests took minutes or hours or days I could totally understand having the user come back later, but for a pdf download that takes half a second (assuming we manage the vioscreen update_headers calls in a smart way), I just don't see a reason for all that.

microsetta_private_api/util/vioscreen.py
dhakim87: I'm opposed to polling approaches, but I recognize it would be necessary to cache responses if we were to move to one.  The good part is that we lower the number of requests we make to vioscreen.  The bad part is that I have no idea if vioscreen updates what comes back in the pdf over time, and have no idea what policy should exist to determine when to pull new data from vioscreen and when to use cached data from the database.  I don't believe they have any kind of ETag to check, but I admit I wasn't looking for one.  

If we plan to go back to them and ask them for a way to hide all the N/A and Age information at the top of the pdf reports (that appears since we don't pass them the information they would need to generate this information), caching results locally will complicate updating to the most recent version of their api.

microsetta_private_api/api/_survey.py
wasade: Okay, thanks. The download sound more correct from -private-api and if needed the headers are modified in -interface

microsetta_private_api/repo/survey_answers_repo.py
wasade: Can you send a query that shows this? I'll investigate a patch

microsetta_private_api/util/vioscreen.py
wasade: That sounds like a good solution. It looks like celery supports periodic tasks natively

microsetta_private_api/util/vioscreen.py
wasade: Just waht was encountered

microsetta_private_api/util/vioscreen.py
wasade: Okay. Let's block and over allocate worker processes.

microsetta_private_api/util/vioscreen.py
wasade: Okay, let's not cache right now and revisit if needed

microsetta_private_api/repo/survey_answers_repo.py
dhakim87: select * from ag_login_surveys left join survey_answers using (survey_id) where vioscreen_status=3 limit 10;

wasade:
Note failures look real here

PR #270, opened 2020-10-16
Split Primary/Vioscreen Surveys Into Separate IDs

dhakim87:
Seems that our database has primary and vioscreen surveys sharing IDs in a number of legacy accounts.  This is frustrating since we need the ability to say what type a particular survey ID is, and saying both isn't part of our model.  This PR is a schema change to fork the IDs- primary surveys are assigned a new UUID and tables that reference the survey ID are updated or forked as necessary.

### DIFF
--- a/microsetta_private_api/db/migration_support.py
+++ b/microsetta_private_api/db/migration_support.py
@@ -373,7 +373,7 @@ class MigrationSupport:
                     linked_barcode = row[0]
                     TRN.add("INSERT INTO source_barcodes_surveys "
                             "(barcode, survey_id) "
-                            "VALUES(%s, %s)", linked_barcode, new_survey_id)
+                            "VALUES(%s, %s)", (linked_barcode, new_survey_id))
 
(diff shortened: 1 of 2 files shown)


wasade (commented):
this seems good... would it make sense to retain a record of old <-> new survey IDs just in case? One possibly stupid way to do it would be a `create table legacy_mapping_patch_70 (old_id varchar, new_id varchar);`?

dhakim87:
I don't particularly need the old->new mapping afaik, if we're worried about data loss we should have snapshots from before the schema change.  I did consider assigning the new IDs as some determined function of the old IDs, we could do that instead, but then we wouldn't have true uuidv4s anymore.  The mapping is also (mostly) implicitly defined by looking at the ag_login_survey table -- anywhere a primary and vioscreen survey were taken for the same account and source either those two used to be the same ID or the surveys weren't taken until after the ID split.

wasade:
True -- very good point regarding the rolling backups.

wasade (approved):

PR #271, opened 2020-10-23
Update missing_sample_info.plain

specsdom:
This is a little rough around the edges as I'm hastily finishing up everything on my last day, but changing the wording of this notification email.  Essentially the issue is that participants get confused when they try to update their sample info by accessing their account, but can't since our system has locked the sample for editing, so it'd be nice to reduce that confusion by telling them what exactly they need to do (i.e. emailing us back with the following information).


wasade:
Thanks! Closing in favor if #273 which includes this commit

PR #272, opened 2020-10-23
Update no_associated_source.plain

specsdom:
Likewise here-- updating the language for these emails to reduce the confusion of participants, as their samples would get locked for editing and would need to be added & updated by an admin on our end.


wasade:
Thanks! Closing in favor of #273

PR #274, opened 2020-10-27
Migrate metadata pulldown from -admin to -private-api

wasade:
This monster of a PR is almost entirely previously merged/tested code from the microsetta-admin repository. The reason for the migration is to facilitate automatic movement of sample metadata from the private-api to Qiita, by way of either (a) the celery periodic mechanisms part of this repository or (b) a cronjob using the non-AuthRocket JWT codepath; I'm still evaluating some pros/cons here. Further motivation to migrate this from -admin to -private-api is that a request for ~100 samples would result in a http timeout, as the time required to extract and transform the data is high particularly when -admin has to make many remote API calls to -private-api.

Assuming the reasoning for this migration is sound, I think the extent of review needed is small. Specifically, because it was previously checked out, I think only a review of changes in the following files is necessary:

- `admin_impl.py`
- `microsetta_private_api.yaml` (note: the diff here is *weird* and seems slightly out of alignment with the preceding endpoint) 

Everything else is essentially the same as from -admin other than removal of the API request calls in favor of interaction via repos.

I'll open a separate PR, if appropriate, for handling the -> Qiita connection


microsetta_private_api/admin/admin_impl.py
```diff
@@ -86,6 +88,20 @@ def sample_pulldown_multiple_survey(token_info,
     return jsonify(sample_pulldown), 200
 
 
+def qiita_compatible_metadata(token_info, include_private, body):
```
wasade, line 91: Should add a exercise of this into `test_admin_api.py`

microsetta_private_api/api/microsetta_private_api.yaml
```diff
@@ -1418,6 +1451,12 @@ components:
       description: true if computed statistics about projects should be included in projects list (takes longer)
       schema:
         type: boolean
+    include_private:
+      name: include_private
+      in: query
+      description: include private information in a metadata pulldown; may be true or false. Optional.
+      schema:
+        $ref: '#/components/schemas/is_active'
```
wasade, line 1459: copy/pasta, should be `include_private`

microsetta_private_api/api/microsetta_private_api.yaml
```diff
@@ -1143,7 +1143,40 @@ paths:
         - $ref: '#/components/parameters/sample_barcode'
       responses:
         '200':
-          description: QIIME-compatible sample metadata
+          description: Summarized sample metadata
```
dhakim87, line 1146: Thoughts on changing the route to match the description or vice versa?

microsetta_private_api/repo/metadata_repo/_constants.py
```diff
@@ -0,0 +1,374 @@
+MISSING_VALUE = 'not provided'
+
+# standard fields that are set based on sampling site
+ENV_LOOKUP = {
```
dhakim87, line 4: I might prefer that this file become a .json at some point.  I think it's an adapter layer that is probably never going to change, but pure json files are easier to change with deployment scripts or schema changes if ever we run into scenarios where the adapter needs to change.  

I could also imagine an argument that this is copy pasted from somewhere else, and therefore indicates the need for a common library.  

I probably wouldn't do anything about this right now, but wait for the first time we actually run into problems with the file.

microsetta_private_api/api/microsetta_private_api.yaml
wasade: I made this change as it was incorrect w.r.t what that endpoint produces

microsetta_private_api/repo/metadata_repo/_constants.py
wasade: Good point :) It *is* copied from somewhere else right now, however I plan on removing it from -admin

dhakim87 (approved):
Seems good.  I don't particularly like the style of MISSING_VALUE fields over just using None with a view that changes it when read but I'm guessing that code is really just along for the ride.  Similarly, a lot of _constants.py appears to define an interface that must be shared with some other code somewhere, and it is always painful to see code like that duplicated.  As far as the code that's the target of this review though, everything looks standard.

microsetta_private_api/admin/admin_impl.py
```diff
@@ -86,6 +88,24 @@ def sample_pulldown_multiple_survey(token_info,
...
+    validate_admin_access(token_info)
+
+    samples = body.get('sample_barcodes')
+    if samples is None:
+        return jsonify(code=404, message='No samples provided'), 404
+
```
wasade, line 97: wrap in transaction

wasade: scratch that, just add comment taht this code is wrapping other repos

wasade: reduce from n+1 to 1 transaction

PR #275, opened 2020-11-05
Support for populating and reading db table of Daklapack "articles"

AmandaBirmingham:
A bite-sized chunk of the new daklapack order functionality :)


microsetta_private_api/db/patches/0071.sql
```diff
@@ -0,0 +1,47 @@
+-- add and fill daklapack_article table
+
+-- create table barcodes.daklapack_article
+CREATE TABLE barcodes.daklapack_article
+(
+    dak_article_id uuid DEFAULT uuid_generate_v4() NOT NULL,
+    dak_article_code INTEGER NOT NULL,
```
dhakim87, line 7: Is article code also unique?  Should it have another unique index on it?

dhakim87: Should be indexed regardless as it appears in the ORDER BY clause

dhakim87 (approved):
Seems good - maybe add a (possibly unique) index to article code for the order by clause.

microsetta_private_api/db/patches/0071.sql
AmandaBirmingham: Excellent--done!

PR #276, opened 2020-11-12
Daklapack orders, step 1

AmandaBirmingham:
functionality to take daklapack order specifications from an api call, generate an order object (including correct json), and write to new db tables that record orders. Note that functionality to actually submit order to daklapack api is not in here (yet) :)


microsetta_private_api/db/patches/0072.sql
```diff
@@ -0,0 +1,25 @@
+-- create tables to record daklapack orders
+
+CREATE TABLE barcodes.daklapack_order
+(
+    dak_order_id uuid DEFAULT uuid_generate_v4() NOT NULL,
+    submitter_acct_id uuid NOT NULL,
+    description varchar,
+    fulfillment_hold_msg varchar,
+    order_json varchar NOT NULL,
```
wasade, line 9: Should this be of either `JSON` or `JSONB` type?

wasade: ...a possibly relevant document https://www.compose.com/articles/faster-operations-with-the-jsonb-data-type-in-postgresql/

microsetta_private_api/model/tests/test_daklapack_order.py
```diff
@@ -0,0 +1,151 @@
...
+        self.assertEqual(DUMMY_DAK_ORDER_ID, real_out.id)
+
+        real_addresses = real_out.order_structure['articles'][0]['addresses']
+        for real_address in real_addresses:
+            # pop out creation date and make sure it is approximately now :)
+            curr_create_str = real_address.pop("creationDate")
+            curr_create_time = parser.isoparse(curr_create_str)
+            self.assertAlmostEqual(datetime.now(timezone.utc),
+                                   curr_create_time,
+                                   delta=timedelta(minutes=2))
```
wasade, line 117: Did not know about this -- that is super cool!

wasade (commented):
Mostly curious about whether a postgres JSON or JSONB type makes sense here

microsetta_private_api/admin/admin_impl.py
```diff
@@ -297,3 +299,34 @@ def get_daklapack_articles(token_info):
...
+
+    body[ORDER_ID_KEY] = str(uuid.uuid4())
+
+    with Transaction() as t:
+        account_repo = AccountRepo(t)
+        body[SUBMITTER_ACCT_KEY] = account_repo.find_linked_account(
+            token_info['iss'], token_info['sub'])
+
+        try:
+            daklapack_order = DaklapackOrder.from_api(**body)
```
dhakim87, line 315: This is both really cool and kind of scary.  Should we be making a defensive copy here that drops any additional fields in body or is body already tightly locked down by the yaml?  Just feels weird to let the client send unstructured data through us to daklapack.  But they are guaranteed to have admin... soo.....

dhakim87: Ooops.  Nevermind - this from_api is making the defensive copy.  

I guess what felt off is that the body parameter was being updated in place.  In C# at least, objects with the lifetime of the request were used not just in within the handler function, but also in other stages of the pipeline, meaning I expect other parts of the Flask pipeline will suddenly be able to see the order ID key, and the submitter acct key.  This could be changed by making a private copy of body local to this function, but it is extremely minor.

microsetta_private_api/model/daklapack_order.py
```diff
@@ -0,0 +1,122 @@
...
+                 order_structure, description, fulfillment_hold_msg,
+                 creation_timestamp=None, last_polling_timestamp=None,
+                 last_polling_status=None):
+
+        # NB: these are the only fields that exists in BOTH the db record
+        # AND, separately, in the json
+        self.id = daklapack_order_id
+        self.creation_timestamp = creation_timestamp
+        if self.creation_timestamp is None:
+            self.creation_timestamp = datetime.now(timezone.utc)
```
dhakim87, line 28: 👍

microsetta_private_api/admin/admin_impl.py
wasade: :+1:, I think restated, this would be adding `body = body.copy()` immediately prior to `body[ORDER_ID_KEY] = str(uuid.uuid4())`

microsetta_private_api/model/daklapack_order.py
```diff
@@ -0,0 +1,122 @@
...
+        fedex_refs_dicts_list = []
+        fedex_keys = {FEDEX_REF_1_KEY: "Reference 1",
+                      FEDEX_REF_2_KEY: "Reference 2",
+                      FEDEX_REF_3_KEY: "Reference 3"}
+        for curr_key, curr_val in fedex_keys.items():
+            curr_input = kwargs.get(curr_key)
+            if curr_input:
+                fedex_refs_dicts_list.append({
+                    "key": curr_val,
+                    "value": curr_input})
```
dhakim87, line 86: Can you walk me through this?  

My read is:  if "fedex_ref_1" is assigned in kwargs, fedex_refs_dicts_list will append the object 
`
{"key": "Reference 1", "value": "fedex_ref_1"}
`
Same for refs 2 and 3, but it never actually looks at the data in kwargs...?

microsetta_private_api/model/daklapack_order.py
```diff
@@ -0,0 +1,122 @@
...
+
+        for curr_addr_dict in address_dicts_list:
+            # creation date is now
+            curr_timestamp = datetime.now(timezone.utc)
```
dhakim87, line 90: Move datetime.now outside of loop?  Appears that was intended based on definition of curr_timestamp at beginning of function.

microsetta_private_api/model/daklapack_order.py
```diff
@@ -0,0 +1,122 @@
...
+            "shippingType": "FEDEX_2_DAY",
+            "shippingProviderMetadata": fedex_refs_dicts_list
+        }
+
+        return cls(order_id, submitter_acct, project_ids_list,
+                   order_structure, description, fulfillment_hold_msg,
+                   creation_timestamp=curr_timestamp)
+
+    def set_last_polling_info(self, last_polling_status,
+                              last_polling_timestamp=None):
```
dhakim87, line 118: Who is going to call this function?  Should this timestamp match creation time when the object is instantiated?

microsetta_private_api/db/patches/0072.sql
```diff
@@ -0,0 +1,25 @@
...
+
+CREATE TABLE barcodes.daklapack_order_to_project
+(
+    dak_order_to_project_id uuid DEFAULT uuid_generate_v4() NOT NULL,
+    dak_order_id uuid NOT NULL,
+    project_id bigint NOT NULL,
+    CONSTRAINT dak_order_to_project_pkey PRIMARY KEY (dak_order_to_project_id),
+    CONSTRAINT fk_dak_order_id FOREIGN KEY ( dak_order_id ) REFERENCES barcodes.daklapack_order( dak_order_id ),
+    CONSTRAINT fk_project_id FOREIGN KEY ( project_id ) REFERENCES barcodes.project( project_id )
+);
```
dhakim87, line 25: Is this many to many, many to one, or one to one?  You might be able to enforce a more rigid relationship with unique indexes on (dak_order_id, project_id) or (project_id, dak_order_id).  Having indexes, unique or not, would speed up lookups that join through the mapping table.

microsetta_private_api/db/patches/0072.sql
```diff
@@ -0,0 +1,25 @@
...
+    creation_timestamp timestamptz,
+    last_polling_timestamp timestamptz,
+    last_polling_status varchar,
+    CONSTRAINT dak_order_pkey PRIMARY KEY (dak_order_id),
+    CONSTRAINT fk_acct_id FOREIGN KEY ( submitter_acct_id ) REFERENCES ag.account( id )
+);
+
+CREATE TABLE barcodes.daklapack_order_to_project
+(
+    dak_order_to_project_id uuid DEFAULT uuid_generate_v4() NOT NULL,
```
dhakim87, line 19: Do we ever intend to use the uuid of the row of the mapping table?  A compound primary key of order_id and project_id might be sufficient.

dhakim87 (approved):
Looks good, I left only trivial/minor comments

microsetta_private_api/admin/admin_impl.py
AmandaBirmingham: >feels weird to let the client send unstructured data through us to daklapack

We aren't really doing that here because the body arguments are being sent to the `from_api` function, which doesn't itself write to the db.  It plucks out only the body json elements with specific, pre-defined keys to be used to fill the new order object, for example like this:

```
        # fields that don't go into json (only db):
        # required:
        submitter_acct = kwargs[SUBMITTER_ACCT_KEY]
        project_ids_list = kwargs[PROJECT_IDS_LIST_KEY]
        # optional:
        description = kwargs.get(DESCRIPTION_KEY)
        fulfillment_hold_msg = kwargs.get(FULFILLMENT_HOLD_MSG)
```

Any other keys/values in the json that comes in via the api are just silently thrown away.

microsetta_private_api/model/daklapack_order.py
AmandaBirmingham: Yes, that is right :)

microsetta_private_api/model/daklapack_order.py
AmandaBirmingham: You are absolutely right--thank you very much for catching that!

microsetta_private_api/model/daklapack_order.py
AmandaBirmingham: This is MOSTLY going to be called long (in computer time--like hours or days) after the order is created/submitted, when we reach out to Daklapack and ask them "hey, what is the status of order X?"  The timestamp should be the time at which we sent out that query, so the timestamp here will basically never match the timestamp of when the order was created (and we don't want to fill this field UNTIL we have actually queried Daklapack about the status, so we don't fill it at instantiation).

microsetta_private_api/db/patches/0072.sql
AmandaBirmingham: This is a many-to-many relationship: one order can include items for multiple projects, and multiple orders can include items for a single project.

microsetta_private_api/model/daklapack_order.py
AmandaBirmingham: :)

microsetta_private_api/model/tests/test_daklapack_order.py
AmandaBirmingham: I did not know about it until I had to write this code :D

microsetta_private_api/admin/admin_impl.py
AmandaBirmingham: Sorry for the earlier response--I didn't refresh and see the new comments.  Adding `body = body.copy()` :)

microsetta_private_api/db/patches/0072.sql
AmandaBirmingham: I had no idea either of those data types existed :)  Have changed it to JSONB.

microsetta_private_api/db/patches/0072.sql
AmandaBirmingham: A very good point!  Have removed the uuid and replaced with compound primary key as suggested.

PR #277, opened 2020-11-13
Account Summaries Backend

dhakim87:
Supports displaying summaries of samples in accounts by email address for microsetta-admin

### DIFF
--- a/microsetta_private_api/admin/admin_impl.py
+++ b/microsetta_private_api/admin/admin_impl.py
@@ -1,4 +1,6 @@
 import uuid
+from collections import defaultdict
+
 from flask import jsonify, Response
 
 from microsetta_private_api.config_manager import SERVER_CONFIG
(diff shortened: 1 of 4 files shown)


microsetta_private_api/admin/admin_impl.py
```diff
@@ -297,3 +302,72 @@ def get_daklapack_articles(token_info):
         admin_repo = AdminRepo(t)
         dak_article_dicts = admin_repo.get_daklapack_articles()
         return jsonify(dak_article_dicts), 200
+
+
+def query_email_stats(body, token_info):
+    validate_admin_access(token_info)
+
+    email_list = body.get("emails")
```
wasade, line 310: I think this should default to `[]` rather than `None` as it is iterated over on line 320

microsetta_private_api/repo/sample_repo.py
```diff
@@ -228,3 +228,18 @@ def dissociate_sample(self, account_id, source_id, sample_id,
         # And detach the sample from the source
         self._update_sample_association(sample_id, None,
                                         override_locked=override_locked)
+
+    def get_sample_status(self, sample_barcode, scan_timestamp):
+        with self._transaction.cursor() as cur:
+            cur.execute(
+                "SELECT "
```
wasade, line 235: A barcode can have many scans, should the select be ordered so the latest is picked?

wasade (changes requested):
Would it be possible to add in a test of one or two "email addresses" from the test database, with and without a project restriction?

microsetta_private_api/repo/sample_repo.py
dhakim87: The WHERE searches with the scan_timestamp of the latest, our model can't determine which one is the latest if there are in fact two scans at the same time.  So I just break the tie with a limit 1.

wasade: Okay, thanks!

PR #278, opened 2020-11-19
Support for submitting order to daklapack api and sending email to daklapack if order fulfillment needs to be held

AmandaBirmingham:
Ready for real review

Original:
WIP: I *believe* all this stuff actually works (when one fills real values into the config! :) but I am struggling with how to unit-test ...

### DIFF
--- a/microsetta_private_api/admin/daklapack_communication.py
+++ b/microsetta_private_api/admin/daklapack_communication.py
@@ -20,7 +20,6 @@ def _get_daklapack_oauth2_session():
     return dak_session
 
 
-# TODO: AB: unit-test daklapack_communication.post_daklapack_order
 def post_daklapack_order(payload):
     oauth_session = _get_daklapack_oauth2_session()
     # the json parameter sets the content-type in the headers
@@ -35,7 +34,6 @@ def post_daklapack_order(payload):
     return result
 
 
-# TODO: AB: unit-test daklapack_communication.send_daklapack_hold_email
 def send_daklapack_hold_email(daklapack_order):
     try:
         template_args = {"order_id": daklapack_order.id,
--- a/microsetta_private_api/admin/daklapack_communication.py
+++ b/microsetta_private_api/admin/daklapack_communication.py
@@ -1,0 +1,55 @@
+from oauthlib.oauth2 import BackendApplicationClient
+from requests_oauthlib import OAuth2Session
+from microsetta_private_api.config_manager import SERVER_CONFIG
+from microsetta_private_api.tasks import send_basic_email as celery_send_email
+
+ORDER_HOLD_TEMPLATE_PATH = "email/daklapack_fulfillment_hold_request"
+
+
+def _get_daklapack_oauth2_session():
+    # run the "Resource Owner Client Credentials Grant Type" oauth2 workflow
+    client_id = SERVER_CONFIG["daklapack_client_id"]
+    auth_client = BackendApplicationClient(client_id=client_id)
+    auth_session = OAuth2Session(client=auth_client)
+    token = auth_session.fetch_token(
+        token_url=SERVER_CONFIG["daklapack_auth_url"],
+        client_id=client_id,
+        client_secret=SERVER_CONFIG["daklapack_client_secret"])
+
+    dak_session = OAuth2Session(client_id, token=token)
+    return dak_session
+
+
+# TODO: AB: unit-test daklapack_communication.post_daklapack_order
+def post_daklapack_order(payload):
+    oauth_session = _get_daklapack_oauth2_session()
+    # the json parameter sets the content-type in the headers
+    # to application/json, whereas if used data parameter, would have to set
+    # content-type manually
+    dak_order_post_url = f"{SERVER_CONFIG['daklapack_api_base_url']}" \
+                         f"/api/Orders"
+    result = oauth_session.post(
+        dak_order_post_url, json=payload,
+        headers={SERVER_CONFIG["daklapack_subscription_key_name"]:
+                     SERVER_CONFIG["daklapack_subscription_key_val"]})
+    return result
+
+
+# TODO: AB: unit-test daklapack_communication.send_daklapack_hold_email
+def send_daklapack_hold_email(daklapack_order):
+    try:
+        template_args = {"order_id": daklapack_order.id,
+                         "fulfillment_hold_msg":
+                             daklapack_order.fulfillment_hold_msg}
+        email_subject = f"Hold fulfillment of order {daklapack_order.id}"
+        dak_service_email = SERVER_CONFIG["daklapack_service_email"]
+        celery_send_email.apply_async(
+            args=[dak_service_email, email_subject,
+                  ORDER_HOLD_TEMPLATE_PATH,
+                  list(template_args.keys()), template_args,
+                  "EMAIL", "DAK_ORDER_HOLD"])
+        email_success = True
+    except Exception:  # noqa
+        email_success = False
+
+    return email_success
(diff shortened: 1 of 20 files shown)


wasade:
Just FYI, it looks like the build is failing with the requirements file (see https://travis-ci.com/github/biocore/microsetta-private-api/jobs/443230190#L610)

microsetta_private_api/admin/daklapack_communication.py
```diff
@@ -0,0 +1,53 @@
...
+
+ORDER_HOLD_TEMPLATE_PATH = "email/daklapack_fulfillment_hold_request"
+
+
+def _get_daklapack_oauth2_session():
+    # run the "Resource Owner Client Credentials Grant Type" oauth2 workflow
+    client_id = SERVER_CONFIG["daklapack_client_id"]
+    auth_client = BackendApplicationClient(client_id=client_id)
+    auth_session = OAuth2Session(client=auth_client)
+    token = auth_session.fetch_token(
```
dhakim87, line 14: Does this cache and use refresh tokens or run the entire flow every time?  Not that it particularly matters at the rate we're placing daklapack orders

microsetta_private_api/admin/daklapack_communication.py
```diff
@@ -0,0 +1,53 @@
...
+                             daklapack_order.fulfillment_hold_msg}
+        email_subject = f"Hold fulfillment of order {daklapack_order.id}"
+        dak_service_email = SERVER_CONFIG["daklapack_service_email"]
+        celery_send_email.apply_async(
+            args=[dak_service_email, email_subject,
+                  ORDER_HOLD_TEMPLATE_PATH,
+                  list(template_args.keys()), template_args,
+                  "EMAIL", "DAK_ORDER_HOLD"])
+        email_success = True
+    except Exception:  # noqa
```
dhakim87, line 50: Maybe grab the exception and return it so we have some hope of debugging this later?

microsetta_private_api/admin/email_templates.py
```diff
@@ -57,3 +57,15 @@ def __init__(self, subject, html, plain, required, event_type, event_sub):
         self.plain = EmailTemplate(plain, required)
         self.event_type = event_type
         self.event_subtype = event_sub
+
+
+class BasicEmailMessage:
```
dhakim87, line 62: This is really basic.  Maybe make it extend EmailMessage and call that classes constructor on one line?

dhakim87: Or add required email templates to the enum??  Hmm... forgot that it was an enum

microsetta_private_api/templates/email/daklapack_fulfillment_hold_request.jinja2
```diff
@@ -0,0 +1,23 @@
...
+Dear Daklapack,
+</p>
+<p>
+Please put a hold on the fulfillment of Microsetta order {{ order_id }}.
+</p>
+<p>
+{{ fulfillment_hold_msg }}
+</p>
+<p>
+If you have questions or difficulties, please contact Edgar Diaz Pardo at ediazpardo@health.ucsd.edu .
```
dhakim87, line 16: Might be a little bit wrong to hardcode the point of contact *shrug*

microsetta_private_api/admin/daklapack_communication.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def send_daklapack_hold_email(daklapack_order):
+    try:
+        template_args = {"order_id": daklapack_order.id,
+                         "fulfillment_hold_msg":
+                             daklapack_order.fulfillment_hold_msg}
+        email_subject = f"Hold fulfillment of order {daklapack_order.id}"
+        dak_service_email = SERVER_CONFIG["daklapack_service_email"]
+        celery_send_email.apply_async(
```
dhakim87, line 44: Can this actually throw for failure?  I imagine this will throw if celery itself is down, but I suspect the fact it says apply_async means it won't return a result prior to leaving this function with email_success set to true.

dhakim87 (approved):
Sending an email remains a pretty poor automated communications channel, I don't think we can actually guarantee receipt of messages over that channel.  But there's not really anything to be done about that until Daklapack writes a new api, or I have a better understanding of why we need to place an order and then put a hold on it, rather than place an order when we actually want it to go out.

microsetta_private_api/admin/daklapack_communication.py
wasade: If it raises, `email_success` will not be set to `True`

microsetta_private_api/admin/daklapack_communication.py
wasade: This would be very exceptional. I'm not sure if the complexity of fully handling the re-raise is worth it? If we get an email unsuccessful indication, we will still have timing information to assess mail logs

wasade (approved):

wasade:
@AmandaBirmingham, possible to do a `git pull --rebase upstream master`? That should bring in the github actions workflow for continuous integration

microsetta_private_api/admin/daklapack_communication.py
AmandaBirmingham: Run the entire workflow every time.  The library in question, requests_oauthlib, has support for refresh tokens but I didn't investigate it.  If it is needed, someone holler please :)

microsetta_private_api/admin/email_templates.py
AmandaBirmingham: >Maybe make it extend EmailMessage and call that classes constructor on one line?

Yeah, I tried that and smack-dab into the Enum :D  You are right that I certainly *could* add to the Enum, but I felt kind of queasy about every new email we send needing to be defined in one big enum .. right now the EmailMessage Enum is really specifically a set of customer support emails, which is a nice clean scope ... However, if we do want to make it a master email handler, I will do so.

microsetta_private_api/templates/email/daklapack_fulfillment_hold_request.jinja2
AmandaBirmingham: Hmm, fair point ... I was thinking that templates are easy to change, but I guess it has to be changed in two places (two templates). @wasade , is this something that is worth moving to the config?

microsetta_private_api/admin/daklapack_communication.py
AmandaBirmingham: Hmm, I am definitely no expert on asynchronous processes.  If this code works the way I hazard a guess it does, then we won't get errors if something happens during the actual email sending from the server ... that will be asynchronous and out of our sight (sadly, because that's definitely something we'd want the user to know ...).  BUT I know from my own development efforts that I definitely CAN end up raising an error in real time here if I pass apply_async inappropriate arguments :D

A good argument could be made that that is a development-level kind of error, not a production-level kind of error, and therefore the only kind of exceptions we could catch here are kinds that are pretty much irrelevant to the user because we should have worked them out in testing (which I hope I have) .... If that is the case, then probably shouldn't bother to catch such errors but just let them be handled as general runtime snafus; I think we would then take out this handling on the grounds that all it does is give the user a false sense of security about their email being sent when actually we don't know it has been.

microsetta_private_api/templates/email/daklapack_fulfillment_hold_request.jinja2
wasade: :+1:, maybe just as a `microsetta_point_of_contact`? However, I vote we do it as a separate PR as we should also centralize the microsetta help account email address as well, and possibly some other contact bits

PR #279, opened 2020-11-25
Route to retrieve preparation data over api

dhakim87:
Adds functionality to retrieve preparation data

### DIFF
--- a/microsetta_private_api/admin/admin_impl.py
+++ b/microsetta_private_api/admin/admin_impl.py
@@ -10,6 +10,7 @@ from microsetta_private_api.model.daklapack_order import DaklapackOrder, \
     ORDER_ID_KEY, SUBMITTER_ACCT_KEY
 from microsetta_private_api.exceptions import RepoException
 from microsetta_private_api.repo.account_repo import AccountRepo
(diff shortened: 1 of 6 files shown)


wasade (changes requested):
I think this is perfect, other than it would be helpful to not require admin. It actually doesn't require authorization either as this is public detail. Sorry for not clarifying that item earlier

dhakim87:
> I think this is perfect, other than it would be helpful to not require admin. It actually doesn't require authorization either as this is public detail. Sorry for not clarifying that item earlier

Possibilities I've considered: 
1.  Public like microsetta-public-api where you don't need an account?  
2.  Public like the read_kit route that gives unused samples from a kit?  As in, you can query this for any barcode so long as you have an account in our system?  
3.  Private to admin and/or person who has claimed this sample and has it associated with their account?
4.  Restricted to admin privileges?  

Right now on 4, I'll switch it to 2?

wasade:
Super helpful, thank you -- l think we should do (2) so we're consistent with private-api where there is a level of auth involved.

wasade (approved):

PR #283, opened 2020-12-10
First pass, GH actions

wasade:


.github/workflows/python-package-conda.yml
```diff
@@ -0,0 +1,85 @@
+name: microsetta-private-api CI
+    
+on: 
+  push:
+    branches: [ master ]
+  pull_request:
+
```
dhakim87, line 7: Does this need a branches: [ master ] as well?  Or any PR to any branch will build and run tests?

.github/workflows/python-package-conda.yml
```diff
@@ -0,0 +1,85 @@
...
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          # based on https://github.com/actions/example-services/blob/master/.github/workflows/postgres-service.yml#L44-L72
+          - 5432/tcp
+
+    steps:
+      # Downloads a copy of the code in your repository before running CI tests
+      - name: Check out repository code
+        uses: actions/checkout@v2
```
dhakim87, line 37: What is the @v2 here?

microsetta_private_api/config_manager.py
```diff
@@ -15,7 +15,7 @@
 class DBConfig(object):
     def __init__(self):
         self.user = 'postgres'
-        self.password = ''
+        self.password = 'postgres'
```
dhakim87, line 18: Intentional change to default localhost postgres server password?

dhakim87 (commented):
I admit I don't really know a ton about GH actions.  Everything seems legit- I've sprinkled a few questions throughout.

.github/workflows/python-package-conda.yml
wasade: Not fully sure, just using templates right now

.github/workflows/python-package-conda.yml
wasade: I believe this refers back to a version of the github action definition from the github actions repository itself. But, not very sure, just pulling from templates

microsetta_private_api/config_manager.py
wasade: I don't know what the default is for the build of postgres used

wasade:
Thanks!!! And ya, I don't know a ton either... but it seems to run :)

PR #290, opened 2021-01-20
Only valid samples are locked

### LINKED ISSUE #287: Revisiting sample locking
Samples are locked on receipt. This is a problem for participants that send in samples with problems as it requires admins to update the the state. Is there a way to lock samples if and only if the are “valid”? This would allow participants to update their sample if there is an issue

dhakim87:
Updated sample constructor to include scan status, though it doesn't pass through api.  sample.is_locked now requires scan status of sample-is-valid

Fix #287

### DIFF
--- a/microsetta_private_api/api/tests/test_api.py
+++ b/microsetta_private_api/api/tests/test_api.py
@@ -502,7 +502,8 @@ def create_dummy_sample_objects(filled=False):
                     None,
                     info_dict['source_id'],
                     info_dict['account_id'],
-                    info_dict["sample_projects"])
+                    info_dict["sample_projects"],
+                    None)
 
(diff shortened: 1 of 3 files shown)


dhakim87:
We need to split the is_locked property:

Sample editing should be locked once sample status is sample-is-valid.
However, sample removal from a source should be locked the moment the sample is first scanned.  

Admin should still be able to override any lock flags.

wasade:
Sorry for delay, would it be possible to include an integration test on this? If the locking mechanics break in the future, it would be great to catch it at CI rather than staging or prod

dhakim87:
With split of is_locked, now requires a matching PR on the interface side.

wasade (approved):

PR #291, opened 2021-01-23
Deleting source now deletes associated surveys.

### LINKED ISSUE #288: Removal of duplicate or empty sources
The help team has had requests to delete empty or duplicate sources

dhakim87:
Fix #288 

See https://github.com/biocore/microsetta-interface/pull/71 for interface side.

### DIFF
--- a/microsetta_private_api/api/tests/test_integration.py
+++ b/microsetta_private_api/api/tests/test_integration.py
@@ -732,6 +732,76 @@ class IntegrationTests(TestCase):
                            headers=MOCK_HEADERS
                            )
 
+    def test_delete_source(self):
+        """
+            Create a source, add a survey, delete the source
+        """
+        """To add a human source, we need to get consent"""
+        resp = self.client.get(
+            '/api/accounts/%s/consent?language_tag=en-US&consent_post_url=%s' %
+            (ACCT_ID, DUMMY_CONSENT_POST_URL),
+            headers=MOCK_HEADERS
+        )
+        check_response(resp)
+
+        # TODO: This should probably fail as it doesn't perfectly match one of
+        #  the four variants of consent that can be passed in.  Split it up?
+        resp = self.client.post(
+            '/api/accounts/%s/consent?language_tag=en-US' %
+            (ACCT_ID,),
+            content_type='application/json',
+            data=json.dumps(
+                {"age_range": "18-plus",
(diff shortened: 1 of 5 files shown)


wasade:
Same here, could this be supported by a test?

microsetta_private_api/api/tests/test_integration.py
```diff
@@ -732,6 +732,76 @@ def test_create_human_source(self):
...
+        """To add a human source, we need to get consent"""
+        resp = self.client.get(
+            '/api/accounts/%s/consent?language_tag=en-US&consent_post_url=%s' %
+            (ACCT_ID, DUMMY_CONSENT_POST_URL),
+            headers=MOCK_HEADERS
+        )
+        check_response(resp)
+
+        # TODO: This should probably fail as it doesn't perfectly match one of
+        #  the four variants of consent that can be passed in.  Split it up?
```
wasade, line 748: Not sure I follow w/ split it up? 

Agree that if it doesn't match, we probably want to not accept... but that would be out of scope for this PR in my opinion

microsetta_private_api/api/tests/test_integration.py
```diff
@@ -732,6 +732,76 @@ def test_create_human_source(self):
...
+                {
+                    'survey_template_id': chosen_survey,
+                    'survey_text': model
+                }),
+            headers=MOCK_HEADERS
+        )
+        check_response(resp, 201)
+
+        # Delete the newly created source.
+        resp = self.client.delete(
```
wasade, line 799: Can this succeed if a sample is associated? I think it would be valuable to exercise that case as well, as a future unexpected change in behavior would lead to a lot of not fun stuff

microsetta_private_api/api/tests/test_integration.py
dhakim87: Copy pasted from another test case, I think out of scope.

microsetta_private_api/api/tests/test_integration.py
dhakim87: Updated to handle this case- It would throw psycopg2 foreign key violations as expected, but the server would return a 500.  Now the server wraps these in repo exceptions for a 422 and the test case is updated to ensure this functionality stays.

wasade: Thanks!

PR #293, opened 2021-01-28
Vioscreen Hotfix - Registry Table

dhakim87:
New table tracks and reuses vioscreen survey ids.  This is potentially a first step towards a sane schema for storing user surveys, and should allow us to reconstruct what sample led to each vioscreen survey manually if need be.  

This is probably not good enough to allow users to return to surveys they have already started, and needs to be tested to see behavior in such a case.  It also does not get updated when vioscreen surveys are attached to multiple samples within a source, does not know what to do when migrating existing vioscreen surveys that are associated with a source but not any samples, probably explodes if you delete a source with an associated vioscreen survey and so forth.  

I don't see a way to make the vioscreen round trip 100% rock solid- there is always a chance of network/power outage/user quitting browser that prevents us from hearing the finished survey.  The best we can do is try to keep track of the status and potentially allow users to return to incomplete surveys, which may be possible now that we hold onto the survey ID.

### DIFF
--- a/microsetta_private_api/api/_survey.py
+++ b/microsetta_private_api/api/_survey.py
@@ -44,7 +44,8 @@ def read_survey_templates(account_id, source_id, language_tag, token_info):
 
 
 def read_survey_template(account_id, source_id, survey_template_id,
-                         language_tag, token_info, survey_redirect_url=None):
+                         language_tag, token_info, survey_redirect_url=None,
+                         vioscreen_ext_sample_id=None):
     _validate_account_access(token_info, account_id)
 
     # TODO: can we get rid of source_id?  I don't have anything useful to do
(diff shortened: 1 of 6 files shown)


microsetta_private_api/api/_survey.py
```diff
@@ -58,14 +59,20 @@ def read_survey_template(account_id, source_id, survey_template_id,
...
+            if vioscreen_ext_sample_id:
+                # User is about to start a vioscreen survey for this sample
+                # record this in the database.
+                db_vioscreen_id = survey_template_repo.create_vioscreen_id(
+                    account_id, source_id, vioscreen_ext_sample_id
+                )
             url = vioscreen.gen_survey_url(
-                language_tag, survey_redirect_url
+                db_vioscreen_id, language_tag, survey_redirect_url
             )
```
AmandaBirmingham, line 70: Is the github display formatting misleading me here?  From the indents I can see, it looks like  `db_vioscreen_id` is set only inside `if vioscreen_ext_sample_id`, but that the call to `vioscreen.gen_survey_url()`, which takes `db_vioscreen_id`, is done outside that `if`?

dhakim87: That's true, but I don't think it's any more broken than the rest of the way things work: vioscreen_ext_sample_id is always set when survey_template_id is SurveyTemplateRepo.VIOSCREEN_ID, because I need the sample ID to actually make this all work.

dhakim87: It will now raise a value error if this is not the case.

PR #298, opened 2021-02-24
BUG fix: vioscreen URL had invalid DOB, causing 502 error on vioscreen

razrichter:
Added query to get birth year and gender by account_id and to assign proper values for building the vioscreen query.
All dates are June 30 of birth year, and defaults are 1/1/1970, Female

### DIFF
--- a/microsetta_private_api/api/_survey.py
+++ b/microsetta_private_api/api/_survey.py
@@ -68,8 +68,9 @@ def read_survey_template(account_id, source_id, survey_template_id,
             else:
                 raise ValueError("Vioscreen Template requires "
                                  "vioscreen_ext_sample_id parameter.")
+            (birth_year, gender) = survey_template_repo.fetch_user_birth_year_gender(account_id)
             url = vioscreen.gen_survey_url(
-                db_vioscreen_id, language_tag, survey_redirect_url
+                db_vioscreen_id, language_tag, survey_redirect_url, birth_year=birth_year, gender=gender
             )
             # TODO FIXME HACK: This field's contents are not specified!
             info.survey_template_text = {
--- a/microsetta_private_api/api/_survey.py
+++ b/microsetta_private_api/api/_survey.py
@@ -48,10 +48,6 @@ def read_survey_template(account_id, source_id, survey_template_id,
                          vioscreen_ext_sample_id=None):
     _validate_account_access(token_info, account_id)
 
-    # TODO: can we get rid of source_id?  I don't have anything useful to do
-    #  with it...  I guess I could check if the source is a dog before giving
-    #  out a pet information survey?
-
     with Transaction() as t:
         survey_template_repo = SurveyTemplateRepo(t)
         info = survey_template_repo.get_survey_template_link_info(
@@ -70,7 +66,7 @@ def read_survey_template(account_id, source_id, survey_template_id,
                                  "vioscreen_ext_sample_id parameter.")
             (birth_year, gender) = \
                 survey_template_repo.fetch_user_birth_year_gender(
-                account_id)
+                account_id, source_id)
             url = vioscreen.gen_survey_url(
                 db_vioscreen_id,
                 language_tag,
--- a/microsetta_private_api/api/_survey.py
+++ b/microsetta_private_api/api/_survey.py
@@ -68,9 +68,15 @@ def read_survey_template(account_id, source_id, survey_template_id,
             else:
                 raise ValueError("Vioscreen Template requires "
                                  "vioscreen_ext_sample_id parameter.")
-            (birth_year, gender) = survey_template_repo.fetch_user_birth_year_gender(account_id)
+            (birth_year, gender) = \
+                survey_template_repo.fetch_user_birth_year_gender(
+                account_id)
             url = vioscreen.gen_survey_url(
-                db_vioscreen_id, language_tag, survey_redirect_url, birth_year=birth_year, gender=gender
+                db_vioscreen_id,
+                language_tag,
+                survey_redirect_url,
+                birth_year=birth_year,
+                gender=gender
             )
             # TODO FIXME HACK: This field's contents are not specified!
             info.survey_template_text = {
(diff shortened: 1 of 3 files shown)


microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -230,3 +230,24 @@ def create_vioscreen_id(self, account_id, source_id,
             else:
                 vioscreen_id = rows[0][0]
         return vioscreen_id
+
+    def fetch_user_birth_year_gender(self, account_id):
```
wasade, line 234: An account may have multiple sources (e.g., a parent, a child, etc). Is it possible for this query to operate on `account_id` and `source_id`?

microsetta_private_api/util/vioscreen.py
```diff
@@ -15,21 +15,33 @@
...
+def gen_survey_url(user_id, language_tag, survey_redirect_url, birth_year=None, gender=None):
     if not survey_redirect_url:
         raise BadRequest("Food Frequency Questionnaire Requires "
                          "survey_redirect_url")
 
+    gender_map = {'male': 1, 'female': 2}
+    if gender is None:
+        gender_id = 2 # default to female
+    else:
+        gender_id = gender_map.get(gender.lower(), 2)  # again, default female
```
wasade, line 27: is `lower` needed? it should be controlled case in the database. Without it, it would remove the need for a conditional here against `None`

wasade (changes requested):
Thanks @razrichter! A few small requests. I suspect it's failing right now due to flake8 errors on line lengths

microsetta_private_api/util/vioscreen.py
razrichter: Good to know! I saw that they were limited in the test db, but wasn't sure about live data. 
Done.

wasade: From production:

```sql
ag=> select distinct response from ag.ag_login_surveys join ag.survey_answers using (survey_id) where survey_question_id=107;
  response
-------------
 Female
 Male
 Other
 Unspecified
(4 rows)
```

razrichter: Changed as suggested.

microsetta_private_api/repo/survey_template_repo.py
razrichter: Could you elaborate? Are the sources such that they'd be from different people, with different sexes and birth years? A comment in read_survey_template:51-53 questions the need for source_id in the function, which is why I left it out from the query.

wasade: A source can be a human, animal or environmental, and an account can have multiple sources. If there are multiple humans, there is no assurance they are the same person. The primary surveys and samples are associated to the source directly. Does that make sense?

That TODO comment should probably be removed as it seems misleading. Want to do that in this PR?

razrichter: Got it. There may be more default/null values from the query with the added restriction (in test DB, ~2K ag_login_survey records don't have a birth year question), but considering the prior state was that all are faked, it's not likely to be a big deal.

wasade: Ya, this is going from definitely garbage to generally not garbage. It's an infinite improvement!! :)

razrichter:
I'm not sure why datetime validation tests should have failed after this change. Is there a clean way to re-run the commit hooks?

microsetta_private_api/api/_survey.py
```diff
@@ -48,10 +48,6 @@ def read_survey_template(account_id, source_id, survey_template_id,
                          vioscreen_ext_sample_id=None):
     _validate_account_access(token_info, account_id)
 
-    # TODO: can we get rid of source_id?  I don't have anything useful to do
-    #  with it...  I guess I could check if the source is a dog before giving
-    #  out a pet information survey?
-
```
dhakim87, line 54: Hooray!  Someone finally used source_id for something 👍

microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -230,3 +230,30 @@ def create_vioscreen_id(self, account_id, source_id,
             else:
                 vioscreen_id = rows[0][0]
         return vioscreen_id
+
+    def fetch_user_birth_year_gender(self, account_id, source_id):
+        """Given an account ID,
```
dhakim87, line 235: and a source_id

microsetta_private_api/util/vioscreen.py
```diff
@@ -48,12 +63,15 @@ def pkcs7_unpad_message(in_message):
     return in_message
 
 
-def encrypt_key(survey_id, language_tag, survey_redirect_url):
+def encrypt_key(survey_id,
+                language_tag,
+                survey_redirect_url,
+                gender_id=2,
+                dob="01011970"
+                ):
```
dhakim87, line 71: Looks like gender_id and dob now have matching default values in both encrypt_key and gen_survey_url.  If gen_survey_url is the only use of encrypt_key (I think it is?), can probably delete one or the other.

razrichter: Makes sense. Since there's no reason for encrypt_key to ever get called from more than one place, that's the one to drop the defaults.

razrichter:
I'm not sure why the tests of the swagger API are failing. I didn't touch them, and `make test` succeeds with deprecation warnings on my machine (the universal excuse...)

dhakim87:
I actually pulled this up to check if my recent PR was failing for the same reasons.  It appears to be the same errors.  I wouldn't worry about it, I don't think it relates to this change, something must have snuck in elsewhere.

wasade:
Unclear what's causing the failure. A quick review of the commit history suggests the entity flagged as problematic was last changed 7 months ago. Agree that is should not block this PR. Smells like maybe a version issue connexion but that's just a guess right now

microsetta_private_api/repo/survey_template_repo.py
wasade: @razrichter, I think this is the last comment then should be :+1:

PR #299, opened 2021-02-24
Expose Vioscreen Status

dhakim87:
Sends the vioscreen_status across the api in a new survey_status field that allows the interface to decide whether to show the report pdf or the survey link.


wasade:
Just to verify, this passes locally? The unrelated failures may obscure errors so just want to check...

PR #300, opened 2021-02-25
DO NOT REALLY MERGE: trivial change to trigger build

AmandaBirmingham:


wasade:
Closing in favor of #303. The `required` field here may be needed to resolve #305, however let's delay as the `openapi-schema-validator` seems to be under a lot of change right now, and we should wait for it to stabilize again

PR #303, opened 2021-03-01
KICK: noop to trigger CI

wasade:
Testing CI. It is *possible* the failure is fixed as `openapi-schema-validator` released a new version over the weekend with a commit related to datetime.


wasade:
Created an issue to address the version pinning (see #305), merging for now to restore CI

PR #310, opened 2021-03-05
More metadata tweaks

wasade:
...based on comments from Gail. 

Of note, `anonymized_name` doesn't fit will in the transform logic as it's based on the index rather than a column, and the existing transforms are based on columns.


wasade:
...I dont think this needs review TBH, merging

PR #311, opened 2021-04-09
TMI sample status report v1

wasade:
This is the first piece of a barcode oriented summary report using a specification defined by CMI.

Evaluation of the FFQs are coming in a later pull request. 

If this seems reasonable, the next phase will be support for interaction with this via microsetta-interface. 

cc @sejsong


microsetta_private_api/repo/admin_repo.py
```diff
@@ -414,6 +414,58 @@ def _rows_to_dicts_list(rows):
 
             return diagnostic
 
+    def get_project_barcodes(self, project_id_or_name):
```
dhakim87, line 417: I think this may have been simpler as two functions- get_barcodes_by_project_name, get_barcodes_by_project_id...  If it was just a WHERE clause being swapped, I'd get keeping them the same, but these feel like two separate queries to me (4 really, with the test for existence).

dhakim87: Looks like it can only search by project_id anywhere based on the definition of project in the yaml, can we split out or delete the search by project name?

microsetta_private_api/admin/admin_impl.py
```diff
@@ -377,6 +379,83 @@ def query_email_stats(body, token_info):
...
+    summaries = []
+    with Transaction() as t:
+        admin_repo = AdminRepo(t)
+        sample_repo = SampleRepo(t)
+        survey_template_repo = SurveyTemplateRepo(t)
+
+        try:
+            project_barcodes = admin_repo.get_project_barcodes(project)
+        except ValueError as e:
+            raise RepoException(e)
```
dhakim87, line 397: Strikes me as odd to wrap the ValueError as RepoException - Value Error is thrown for project id not found, which strikes me as a 404, RepoException will be converted to a 422.

microsetta_private_api/admin/admin_impl.py
```diff
@@ -377,6 +379,83 @@ def query_email_stats(body, token_info):
...
+
+        if barcodes is None:
+            barcodes = project_barcodes
+        else:
+            barcodes = [b for b in barcodes if b in set(project_barcodes)]
```
dhakim87, line 402: Should we send back anything for barcodes that aren't associated with the project?  Like {"00123": summary123, "xxxxx":"Not Found"}?

microsetta_private_api/admin/admin_impl.py
```diff
@@ -377,6 +379,83 @@ def query_email_stats(body, token_info):
...
+            account_email = None if account is None else account.email
+            source_email = None
+            source_type = None if source is None else source.source_type
+            vio_id = None
+
+            if source is not None and source_type is Source.SOURCE_TYPE_HUMAN:
+                source_email = source.email
+
+                # NOTE: create_vioscreen_id is idempotent
+                vio_id = survey_template_repo.create_vioscreen_id(account.id,
```
dhakim87, line 419: I'm pretty sure create_vioscreen_id has to write the ID that it generates to the database, that's how it becomes idempotent.  But without committing the transaction at the end, the written ID will be rolled back and forgotten.  Thus every time you call this function on a barcode with no associated vioscreen id, we generate a different random vioscreen id and return it.  

Can be fixed either by committing the transaction, or adding a get_vioscreen_id_if_exists kind of method.  I think the second is more correct - no reason to report a vioscreen id if this user has never taken one.

microsetta_private_api/admin/admin_impl.py
```diff
@@ -377,6 +379,83 @@ def query_email_stats(body, token_info):
...
+                "source-type": source_type,
+                "site-sampled": sample.site,
+                "source-email": source_email,
+                "account-email": account_email,
+                "vioscreen_username": vio_id,
+
+                # NOTE: we *do not* have accurate visibility on these statuses
+                # yet so they are assumed false. This support is coming in
+                # a separate pull request
+                "ffq-taken": False,
```
dhakim87, line 440: Probably needs to be updated at the same time as the vio_id

microsetta_private_api/admin/admin_impl.py
wasade: ohhhhhh that is a very good catch...

microsetta_private_api/admin/admin_impl.py
wasade: :+1:

microsetta_private_api/admin/tests/test_admin_api.py
```diff
@@ -975,3 +975,73 @@ def test_post_daklapack_orders_failure_dak_api(self):
...
+            headers=MOCK_HEADERS
+        )
+
+        # an empty string project should be unkown
+        self.assertEqual(422, response.status_code)
+
+    def test_query_barcode_stats_no_project_barcodes(self):
+        # project isn't nullable so this is qualitatively the same as
+        # test_query_barcode_stats_no_project_no_barcodes
+        pass
```
dhakim87, line 995: I vote delete this test 😄

microsetta_private_api/repo/admin_repo.py
wasade: I'll delete search by project name, and if the need comes again, we can re-add

microsetta_private_api/admin/admin_impl.py
wasade: Okay cool, sounds good. Didn't realize the `ValueError` would yield 404

microsetta_private_api/admin/tests/test_admin_api.py
```diff
@@ -975,3 +975,73 @@ def test_post_daklapack_orders_failure_dak_api(self):
...
+        input_json = json.dumps({'project': 7, 'sample_barcodes': None})
+
+        response = self.client.post(
+            "api/admin/account_barcode_summary",
+            content_type='application/json',
+            data=input_json,
+            headers=MOCK_HEADERS
+        )
+
+        # an empty string project should be unkown
```
dhakim87, line 1007: Copy pasted comment

microsetta_private_api/admin/admin_impl.py
wasade: I believe so, yes. Good spot

microsetta_private_api/admin/tests/test_admin_api.py
wasade: What, you don't love seeing extra dots on the screen with pytest executions...? :) Okay, sounds good, will remove. Had added first when enumerating conditions then realized after it wasn't needed

microsetta_private_api/admin/tests/test_admin_api.py
wasade: 🍝

dhakim87 (commented):
Only non trivial issue looks to be in generating the vio_id when the user hasn't taken that survey, and I'm not even 100% sure that's an issue depending on how you're using this.  Also looks like you were already aware of it.  

Trivial stuff is mostly in the comments, handling the int/str in one function feels unnecessary and the some of the tests will be fragile if we ever want to change our baseline database.

microsetta_private_api/admin/admin_impl.py
dhakim87: I don't think that it does, honestly it probably 500's out.  There's a Werkzeug.NotFound you can throw from within get_project_barcodes that will bubble up and return a 404.

microsetta_private_api/admin/admin_impl.py
wasade: having this throw a 404 instead

dhakim87 (approved):
Looks good to me, saw changes to address the vioscreen creation and warn on missing barcodes, which were the only issues I remember.

PR #312, opened 2021-04-09
Activation Codes

dhakim87:
Includes urls to generate activation code, CRUD operations for persisting them, updates account creation to allow for them.  

Probably shouldn't be merged until we have a way to distribute these to end users?  
Relies on PR in interface for admin pages and changes to account creation.

### DIFF
--- a/microsetta_private_api/api/_account.py
+++ b/microsetta_private_api/api/_account.py
@@ -88,8 +88,9 @@ def register_account(body, token_info):
         acct_repo = AccountRepo(t)
         acct_repo.create_account(account_obj)
         new_acct = acct_repo.get_account(new_acct_id)
-
-        activation_repo.use_activation_code(body["email"], code)
+        
+        if code != "":
+            activation_repo.use_activation_code(body["email"], code)
         t.commit()
 
     response = jsonify(new_acct.to_api())
--- a/microsetta_private_api/api/_account.py
+++ b/microsetta_private_api/api/_account.py
@@ -79,6 +79,9 @@ def register_account(body, token_info):
             )
             if not can_activate:
                 return jsonify(code=404, message=cause), 404
+            else:
+                activation_repo.use_activation_code(body["email"], code)
+
         if kit_name != "":
             kit_repo = KitRepo(t)
             kit = kit_repo.get_kit_all_samples(kit_name)
@@ -88,9 +91,6 @@ def register_account(body, token_info):
         acct_repo = AccountRepo(t)
         acct_repo.create_account(account_obj)
         new_acct = acct_repo.get_account(new_acct_id)
-        
-        if code != "":
-            activation_repo.use_activation_code(body["email"], code)
         t.commit()
 
     response = jsonify(new_acct.to_api())
--- a/microsetta_private_api/api/_account.py
+++ b/microsetta_private_api/api/_account.py
@@ -59,16 +59,20 @@ def register_account(body, token_info):
     # First register with AuthRocket, then come here to make the account
     new_acct_id = str(uuid.uuid4())
     body["id"] = new_acct_id
+    # Account.from_dict requires a kit_name, even if blank
+    kit_name = body.get("kit_name", "")
+    body["kit_name"] = kit_name
+    code = body.get("code", "")
+    body["code"] = code
+
     account_obj = Account.from_dict(body, token_info[JWT_ISS_CLAIM_KEY],
                                     token_info[JWT_SUB_CLAIM_KEY])
 
-    kit_name = body.get("kit_name", "")
-    code = body.get("code", "")
     if kit_name == "" and code == "":
         return jsonify(
-            code=404,
+            code=400,
             message="Account registration requires "
-                    "valid kit ID or activation code")
+                    "valid kit ID or activation code"), 400
 
     with Transaction() as t:
         activation_repo = ActivationRepo(t)
--- a/microsetta_private_api/api/_account.py
+++ b/microsetta_private_api/api/_account.py
@@ -12,6 +12,7 @@ from microsetta_private_api.api.literals import AUTHROCKET_PUB_KEY, \
 from microsetta_private_api.model.account import Account, AuthorizationMatch
 from microsetta_private_api.model.address import Address
 from microsetta_private_api.repo.account_repo import AccountRepo
+from microsetta_private_api.repo.activation_repo import ActivationRepo
 from microsetta_private_api.repo.kit_repo import KitRepo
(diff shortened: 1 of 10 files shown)


microsetta_private_api/db/patches/0076.sql
```diff
@@ -0,0 +1,17 @@
...
+-- TODO: Expecting the user to use the same email will probably blow up in our face.
+-- Procedure:
+-- 1. Admin user sends requested emails to url
+-- 2. Codes are generated as necessary, written to db, returned to admin
+-- 3. Admin sends codes to distributor service, which passes activation code on to user
+-- 4. User creates account using same email and code.
+CREATE TABLE ag.activation
+(
+    email varchar NOT NULL,
+    code varchar NOT NULL,
```
wasade, line 11: I *think* this is the case but just to check, in the absence of an explicit PK, it is still possible to have an FK relation to this field *because* it has a `unique index`, right?

microsetta_private_api/model/activation_code.py
```diff
@@ -0,0 +1,17 @@
...
+# 4.5 bits of entropy per character,
+# need ~15 to get 64 bits entropy:
+# 1/(2^32) chance of collision by
+# birthday paradox
+CHARSET22 = "ACEFHJKLMNPRTUVWXY3479"
+
+
+class ActivationCode:
+    @classmethod
+    def generate(cls):
```
wasade, line 12:
```suggestion
    @staticmethod
    def generate():
```

microsetta_private_api/repo/activation_repo.py
```diff
@@ -0,0 +1,102 @@
+from microsetta_private_api.model.activation_code import ActivationCode
+from microsetta_private_api.repo.base_repo import BaseRepo
+
+
+class ActivationRepo(BaseRepo):
+    def __init__(self, transaction):
+        super().__init__(transaction)
+
+    @staticmethod
+    def to_api(row):
```
wasade, line 10: This is bypassing a model object, should there be one? This is I think the first definition of `to_api` which relies on arguments rather than instance state

microsetta_private_api/repo/activation_repo.py
```diff
@@ -0,0 +1,102 @@
...
+            "email": row[0],
+            "code": row[1],
+            "activated": row[2]
+        }
+
+    def search_email(self, email_query):
+        with self._transaction.cursor() as cur:
+            cur.execute(
+                "SELECT email, code, activated FROM "
+                "activation WHERE email LIKE %s", (email_query + "%",)
```
wasade, line 21: Could this be `ILIKE`?

microsetta_private_api/repo/activation_repo.py
```diff
@@ -0,0 +1,102 @@
...
+            )
+            return [ActivationRepo.to_api(r) for r in cur.fetchall()]
+
+    def search_code(self, code_query):
+        with self._transaction.cursor() as cur:
+            cur.execute(
+                "SELECT email, code, activated FROM "
+                "activation WHERE code LIKE %s", (code_query + "%")
```
wasade, line 29: Same, since the code is defined entirely in a single case

microsetta_private_api/repo/activation_repo.py
```diff
@@ -0,0 +1,102 @@
...
+            code = self.get_activation_code(email)
+            result[email] = code
+        return result
+
+    def can_activate_with_cause(self, email, code):
+        with self._transaction.cursor() as cur:
+            cur.execute(
+                "SELECT email, code, activated "
+                "FROM activation WHERE "
+                "email = %s AND "
```
wasade, line 77: oh... ignore the ILIKE comments above as that would make this suck

microsetta_private_api/repo/activation_repo.py
```diff
@@ -0,0 +1,102 @@
...
+        result = {}
+        for email in emails:
+            code = self.get_activation_code(email)
+            result[email] = code
+        return result
+
+    def can_activate_with_cause(self, email, code):
+        with self._transaction.cursor() as cur:
+            cur.execute(
+                "SELECT email, code, activated "
```
wasade, line 75: Why return `code`? It looks like it isn't used

microsetta_private_api/api/_account.py
```diff
@@ -61,15 +62,35 @@ def register_account(body, token_info):
...
+            kit = kit_repo.get_kit_all_samples(kit_name)
+            if kit is None:
+                return jsonify(code=404, message="Kit name not found"), 404
 
         acct_repo = AccountRepo(t)
         acct_repo.create_account(account_obj)
         new_acct = acct_repo.get_account(new_acct_id)
+        
+        if code != "":
+            activation_repo.use_activation_code(body["email"], code)
```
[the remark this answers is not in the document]
wasade: nm, should be after an account can actually be created

wasade (commented):
Thanks @dhakim87! This is awesome, added a few comments/questions. For most other endpoints, things like the open API parameters and schemas are referenced rather than inline, do you think that makes sense to do here too?

microsetta_private_api/repo/activation_repo.py
```diff
@@ -0,0 +1,102 @@
...
+        with self._transaction.cursor() as cur:
+            cur.execute(
+                "SELECT code "
+                "FROM activation "
+                "WHERE email=%s", (email,))
+            r = cur.fetchone()
+            if r is not None:
+                # Already an activation code available
+                # for this email, return it.
+                return r[0]
```
wasade, line 43: What happens if the same person receives two collection kits and different time points? The activation code would be the same, and presumably already used? I wonder if, in that case, we may want to (in the UI side) prompt them to login using their account

microsetta_private_api/repo/activation_repo.py
dhakim87: Add model object that can wrap this to_api

microsetta_private_api/api/_account.py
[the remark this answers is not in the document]
dhakim87: Moved it up

microsetta_private_api/repo/activation_repo.py
dhakim87: Fair

microsetta_private_api/repo/activation_repo.py
dhakim87: They're already logged in by the time they hit our account creation, since it happens after authrocket's account creation steps.  So if they already made an account with us, they'll be logged into it already.

PR #315, opened 2021-04-27
Radio Fields

cassidysymons:
1) Added model for the radio button field type to model/vue/vue_field.py
2) Added logic to util/vue_adapter.py to switch fields of type "SINGLE" that have six or fewer options from select to radio


wasade:
Am :+1: after CI, thanks @cassidysymons!

PR #322, opened 2021-05-06
Much of the vioscreens

wasade:
This pull request brings in partial support for representation of Vioscreen FFQ data. The motivation to issue the PR right now is the number of lines of code changed is getting appreciable, and additionally, the interaction with the Vioscreen session data is critical for accurate per-sample reporting with respect to the FFQ status.

Much of this code was pair programmed with @Oliver-Nizet


microsetta_private_api/admin/admin_impl.py
```diff
@@ -440,7 +442,11 @@ def query_barcode_stats(body, token_info):
 
             sample_status = sample_repo.get_sample_status(
                 sample.barcode,
-                sample._latest_scan_timestamp  # noqa
```
dhakim87, line 443: The # noqa was because flake8 doesn't like it when we reference 'private' variables, starting with _, like _latest_scan_timestamp.  Probably have to put this back to get past the build server lint checks, but if not, I'm happy to see it go.

microsetta_private_api/api/microsetta_private_api.yaml
```diff
@@ -1563,6 +1563,40 @@ paths:
               schema:
                 type: array
 
+  '/admin/account_barcode_summary':
+    post:
+      operationId: microsetta_private_api.admin.admin_impl.query_barcode_stats
```
dhakim87, line 1568: Something is wrong.  microsetta_private_api.admin.admin_impl.query_barcode_stats wasn't added in this PR.  Was this accidentally copy pasted from line 1515 and never removed?

microsetta_private_api/db/patches/0078.sql
```diff
@@ -0,0 +1,144 @@
+-- create tables to match new models for sessions and percent energy data
```
dhakim87, line 1: May have to rename file, depending on who merges first, I may have already taken 0078.

microsetta_private_api/db/patches/0078.sql
```diff
@@ -0,0 +1,144 @@
...
+DROP TABLE IF EXISTS ag.vioscreen_percentenergy;
+DROP TABLE IF EXISTS ag.vioscreen_dietaryscore;
+DROP TABLE IF EXISTS ag.vioscreen_foodconsumption;
+DROP TABLE IF EXISTS ag.vioscreen_eatingpatterns;
+DROP TABLE IF EXISTS ag.vioscreen_mpeds;
+DROP TABLE IF EXISTS ag.vioscreen_percentenergy;
+DROP TABLE IF EXISTS ag.vioscreen_foodcomponents;
```
dhakim87, line 8: Yay, finally!

microsetta_private_api/db/patches/0078.sql
```diff
@@ -0,0 +1,144 @@
...
+    endDate timestamp,
+    cultureCode varchar NOT NULL,
+    created timestamp NOT NULL,
+    modified timestamp NOT NULL
+    -- no FK is set to ag.vioscreen_registry.vio_id <-> username
+    -- as ag.vioscreen_registry.vio_id is not assured to be UNIQUE.
+    -- this is due to the historical relationship of 1 FFQ to many 
+    -- samples. 
+);
+CREATE INDEX vio_sess_by_username ON ag.vioscreen_sessions(username);
```
dhakim87, line 25: Unless these names match something exactly on the vioscreen side, sql naming conventions I've seen usually prefer all lowercase with underscore separators, rather than camelcase as is done here.  Any arguments for or against renaming to:  session_id, protocol_id, start_date, end_date, culture_code?

dhakim87: Same comment other tables

microsetta_private_api/db/patches/0078.sql
```diff
@@ -0,0 +1,144 @@
...
+    shortDescription varchar NOT NULL,
+    units varchar NOT NULL
+);
+
+CREATE TABLE ag.vioscreen_percentenergy (
+    id uuid DEFAULT uuid_generate_v4() NOT NULL,
+    sessionId varchar NOT NULL,
+    code varchar NOT NULL,
+    amount float NOT NULL,
+    CONSTRAINT vioscreen_percentenergy_pkey PRIMARY KEY ( id ),
```
dhakim87, line 39: Minor syntactic sugar, I believe you can write "id uuid DEFAULT uuid_generate_v4() NOT NULL PRIMARY KEY" and it will generate this constraint for you.

dhakim87: Though maybe I'm misinterpreting - You want both the uuid to be the unique primary key and separately you want a uniqueness constraint on sessionId and code?  I don't know what sessionId and code are, do you want sessionId to be unique and separately code to be unique, or do you want the pair (sessionId,code) to be unique?  

If the latter, you can update line 44 as CREATE UNIQUE INDEX vio_peren_by_sessionid ON ag.vioscreen_percentenergy(sessionId,code)

microsetta_private_api/db/patches/0078.sql
```diff
@@ -0,0 +1,144 @@
...
+    supplement varchar NOT NULL,
+    frequency varchar NOT NULL,
+    amount varchar NOT NULL,
+    average varchar NOT NULL,
+    CONSTRAINT vioscreen_supplements_pkey PRIMARY KEY ( id ),
+    UNIQUE (sessionId, supplement),
+    FOREIGN KEY (sessionId) REFERENCES ag.vioscreen_sessions (sessionId)
+);
+
+INSERT INTO ag.vioscreen_percentenergy_code
```
dhakim87, line 81: What's the advantage in persisting these lookup tables to the database?  If these never change, why not define them in python?

microsetta_private_api/api/microsetta_private_api.yaml
wasade: Good spot! A merge conflict didn't get handled right, thanks

microsetta_private_api/db/patches/0078.sql
wasade: A fight to the finish...

microsetta_private_api/util/vioscreen.py
```diff
@@ -284,15 +302,21 @@ def foodconsumption(self, id_):
...
+    def user(self, username):
+        return self.get(f'users/{username}')
+
     def sessions(self, vioscreen_user):
         detail = self.get('users/%s/sessions' % vioscreen_user)
-
         # TODO: Can we push waiting for the async result out to flask to
         #  free up the flask worker for more requests?
-        detail = detail.get()  # Wait for async result
+        if self.perform_async:
```
dhakim87, line 318: I think this reverses sync and async- if in async mode, you won't want to wait for detail.get().  If in sync mode, you do want to wait for detail.get() as that makes it synchronous.  

Or does this just always wait synchronously, which is probably what my TODO was complaining about before?

microsetta_private_api/db/patches/0078.sql
wasade: On the fence, but happy to go either way that you prefer. These are the names in the JSON from vioscreen. However, it is inconsistent with the naming in the rest of the database

microsetta_private_api/db/patches/0078.sql
wasade: The pair (sessionId, code) to be unique -- did not know about `CREATE UNIQUE INDEX`, that is super cool, thanks!!

microsetta_private_api/db/patches/0078.sql
wasade: There are some changes that can occur overtime as vioscreen updates its database

microsetta_private_api/util/vioscreen.py
```diff
@@ -333,4 +357,72 @@ def top_food_report(self, session_id):
...
+
+@celery.task(ignore_result=False)
+def update_session_detail():
+    # The interaction with the API is occuring within a celery task
+    # and the recommendation from celery is to avoid depending on synchronous
+    # tasks from within a task. As such, we'll use non-async calls here. See
+    # http://docs.celeryq.org/en/latest/userguide/tasks.html#task-synchronous-subtasks
+
+    # HOWEVER, we could implement a watch and monitor child tasks, but
+    # not sure if that would be a particular benefit here
```
dhakim87, line 374: This looks like a comment I would have written, does it match the implementation here?

microsetta_private_api/util/vioscreen.py
wasade: I could have named this poorly... I had to fight with celery for a bit. I believe this preserves the existing behavior where the `detail` was the result of a `.delay`.

microsetta_private_api/util/vioscreen.py
wasade: I must have been channeling your energy :) we are not doing a watch and monitor, seemed overly complex for what was occuring

wasade:
There is a valid failing test right now that I *think* is due to my basing it off of residual state in my local database. I should have a fix in tomorrow

microsetta_private_api/db/patches/0078.sql
dhakim87: Hmm... so this is like a versioned api and we need to change what we query for as they update?  Maybe add a protocol version column or something if that's really the case, persisting here still feels like overkill to me though.

microsetta_private_api/repo/vioscreen_repo.py
```diff
@@ -1,7 +1,481 @@
...
+        with self._transaction.cursor() as cur:
+            cur.execute(f"""INSERT INTO ag.vioscreen_sessions (
+                               {self._sql_cols}
+                               )
+                           VALUES (%s, %s, %s, %s, %s, %s, %s, %s, %s)
+                           ON CONFLICT (sessionId)
+                           DO UPDATE SET
+                               status = EXCLUDED.status,
+                               endDate = EXCLUDED.endDate,
+                               modified = EXCLUDED.modified""",
```
dhakim87, line 46: I'm not super familiar with this syntax, does this DO UPDATE SET update all the fields, or does it only update status, endDate, and modified?  Are these the only columns which could possibly have changed?

microsetta_private_api/model/vioscreen.py
```diff
@@ -0,0 +1,188 @@
...
+        self.modified = update['modified']
+        self.status = update['status']
+        return self
+
+    @property
+    def is_complete(self):
+        return (self.endDate is not None) and (self.status == 'Finished')
+
+    @classmethod
+    def not_present(cls, username):
```
dhakim87, line 64: This is a weird name for a constructor

dhakim87: I thought this would return a boolean

microsetta_private_api/repo/vioscreen_repo.py
```diff
@@ -1,7 +1,481 @@
...
+            The first index if True indicates the FFQ is completed.
+            The second index if True indicates the FFQ has been started,
+                but may or may not be completed.
+            The third index is the exact status from Vioscreen, or None
+                if there is no FFQ associated with the sample.
+        """
+        with self._transaction.cursor() as cur:
+            cur.execute("""SELECT status
+                           FROM ag.vioscreen_sessions vs
+                           JOIN ag.vioscreen_registry vr
```
dhakim87, line 174: Wow.  I had no idea that the AS was optional, I've always seen this as FROM ag.vioscreen_sessions AS vs

microsetta_private_api/repo/vioscreen_repo.py
```diff
@@ -1,7 +1,481 @@
...
+                return (is_complete, is_taken, status)
+            else:
+                raise ValueError("A sample should not have multiple FFQs")
+
+
+class VioscreenPercentEnergyRepo(BaseRepo):
+    def __init__(self, transaction):
+        super().__init__(transaction)
+
+    def insert_percent_energy(self, vioscreen_percent_energy):
```
dhakim87, line 193: Is it okay that this is an insert rather than an upsert?  You'll never call this on the same id twice right?

microsetta_private_api/repo/vioscreen_repo.py
```diff
@@ -1,7 +1,481 @@
...
+        """
+        # upsert based off of https://stackoverflow.com/a/36799500/19741
+        with self._transaction.cursor() as cur:
+            cur.execute(f"""INSERT INTO ag.vioscreen_sessions (
+                               {self._sql_cols}
+                               )
+                           VALUES (%s, %s, %s, %s, %s, %s, %s, %s, %s)
+                           ON CONFLICT (sessionId)
+                           DO UPDATE SET
+                               status = EXCLUDED.status,
```
wasade, line 44: update startDate

wasade, line 44: Fall back to matching the insert

microsetta_private_api/repo/vioscreen_repo.py
```diff
@@ -1,7 +1,481 @@
...
+        with self._transaction.cursor() as cur:
+            cur.execute(f"""SELECT {self._sql_cols}
+                            FROM ag.vioscreen_sessions
+                            WHERE username = %s""",
+                        (username, ))
+            rows = cur.fetchall()
+            if len(rows) == 0:
+                return None
+            else:
+                return [VioscreenSession(*row) for row in rows]
```
wasade, line 98: Add TODO, test if we can add a unique constraint on `username`

microsetta_private_api/model/vioscreen.py
```diff
@@ -0,0 +1,188 @@
...
+        self.username = username
+        self.protocolId = protocolId
+        self.status = status
+        self.startDate = startDate
+        self.endDate = endDate
+        self.cultureCode = cultureCode
+        self.created = created
+        self.modified = modified
+
+    def update_from_vioscreen(self, update):
```
wasade, line 52: this should update everything

microsetta_private_api/db/patches/0078.sql
```diff
@@ -0,0 +1,144 @@
...
+
+CREATE TABLE ag.vioscreen_percentenergy_code (
+    code varchar PRIMARY KEY,
+    description varchar NOT NULL,
+    shortDescription varchar NOT NULL,
+    units varchar NOT NULL
+);
+
+CREATE TABLE ag.vioscreen_percentenergy (
+    id uuid DEFAULT uuid_generate_v4() NOT NULL,
```
wasade, line 35: declare PRIMARY KEY here

microsetta_private_api/model/vioscreen.py
```diff
@@ -0,0 +1,188 @@
...
+        self.amount = amount
+
+    @classmethod
+    def from_vioscreen(cls, component):
+        return cls(component['code'], component['description'],
+                   component['shortDescription'], component['units'],
+                   component['amount'])
+
+
+class VioscreenPercentEnergy(ModelBase):
```
wasade, line 101:
```suggestion
class VioscreenPercentEnergy(ModelBase):
    _codes = {'some_name': (foo, bar, baz),
                        ...}
```

microsetta_private_api/repo/vioscreen_repo.py
```diff
@@ -1,7 +1,481 @@
...
+        tuple
+            The code, description, short description and units for a
+            particular code
+
+        Raises
+        ------
+        NotFound
+            A NotFound error is raised if the code is unrecognized
+        """
+        with self._transaction.cursor() as cur:
```
wasade, line 281: return [VioscreenPercentEnergy._code[foo] ..]

wasade, line 281: do stuff with the static lookup dictionary

microsetta_private_api/repo/vioscreen_repo.py
```diff
@@ -1,7 +1,481 @@
...
+        vioscreen_dietary_score : VioscreenDietaryScore
+            An instance of a dietary score model
+
+        Returns
+        -------
+        int
+            The number of inserted rows
+        """
+        with self._transaction.cursor() as cur:
+            cur.execute("""SELECT sessionId
```
wasade, line 311: drop insert check

microsetta_private_api/repo/vioscreen_repo.py
```diff
@@ -1,7 +1,481 @@
...
+            and score type
+
+        Raises
+        ------
+        NotFound
+            A NotFound error is raised if the code or score type are
+            unrecognized
+        """
+
+        with self._transaction.cursor() as cur:
```
wasade, line 392: do stuff with the look up dictionary

microsetta_private_api/repo/vioscreen_repo.py
```diff
@@ -1,7 +1,481 @@
...
+        vioscreen_supplements : VioscreenSupplements
+            An instance of a supplements model
+
+        Returns
+        -------
+        int
+            The number of inserted rows
+        """
+        with self._transaction.cursor() as cur:
+            cur.execute("""SELECT sessionId
```
wasade, line 423: drop the insert check

microsetta_private_api/util/vioscreen.py
```diff
@@ -333,4 +357,72 @@ def top_food_report(self, session_id):
...
         #  answers and encode/decode things.  Blah.
-        return base64.b64decode(async_result.get().encode("utf-8"))
+        if self.perform_async:
+            return base64.b64decode(result.get().encode("utf-8"))
+        else:
+            return base64.b64decode(result.encode('utf-8'))
+
+
+@celery.task(ignore_result=False)
+def update_session_detail():
```
wasade, line 367: double check how many beat workers are running, make sure we have > 1

microsetta_private_api/util/vioscreen.py
```diff
@@ -333,4 +357,72 @@ def top_food_report(self, session_id):
...
+        r = VioscreenSessionRepo(t)
+        unfinished_sessions = r.get_unfinished_sessions()
+
+    n_to_get = len(unfinished_sessions)
+    for idx, sess in enumerate(unfinished_sessions, 1):
+        updated = []
+        if sess.sessionId is None:
+            # a session requires a mix of information from Vioscreen's
+            # representation of a user and a session
+            user_detail = vio_api.user(sess.username)
```
wasade, line 394: handle the scenario where vioscreen doesn't know what the username is

microsetta_private_api/util/vioscreen.py
```diff
@@ -333,4 +357,72 @@ def top_food_report(self, session_id):
...
+              "message": "Gathering unfinished sessions..."})
+
+    # obtain our current unfinished sessions to check
+    with Transaction() as t:
+        r = VioscreenSessionRepo(t)
+        unfinished_sessions = r.get_unfinished_sessions()
+
+    n_to_get = len(unfinished_sessions)
+    for idx, sess in enumerate(unfinished_sessions, 1):
+        updated = []
```
wasade, line 390: wrap forloop body in try/except

microsetta_private_api/repo/vioscreen_repo.py
```diff
@@ -270,24 +269,28 @@ def _get_code_info(self, code):
...
-                           FROM ag.vioscreen_percentenergy_code
-                           WHERE code = %s""",
-                        (code,))
-            row = cur.fetchone()
-            if row is not None:
-                return row
-            else:
-                raise NotFound("No such code: " + code)
+        
+        code_lookup = {'%mfatot': ('Percent of calories from Monounsaturated Fat', 'Monounsaturated Fat', '%'),
```
wasade, line 281: :+1: on the use of tuple here

microsetta_private_api/util/vioscreen.py
```diff
@@ -333,4 +358,88 @@ def top_food_report(self, session_id):
...
+        state="SUCCESS",
+        meta={"completion": n_to_get,
+              "status": "SUCCESS",
+              "message": f"{n_to_get} sessions updated"})
+
+    if len(failed_sessions) > 0:
+        # ...and let's make Daniel feel bad about not having a better means to
+        # log what hopefully never occurs
+        payload = ''.join(['%s : %s\n' % (s, m) for s, m in failed_sessions])
+        send_email("danielmcdonald@ucsd.edu", "pester_daniel",
```
dhakim87, line 443: I love it.

dhakim87 (approved):

PR #324, opened 2021-05-14
Per sample barcode summaries including email support

wasade:
Testing out the celery component right now. The directed procedure using a specified set of barcodes via microsetta admin appears to check out locally

### DIFF
--- a/microsetta_private_api/admin/admin_impl.py
+++ b/microsetta_private_api/admin/admin_impl.py
@@ -5,7 +5,6 @@ from flask import jsonify, Response
 
 from microsetta_private_api.config_manager import SERVER_CONFIG
 from microsetta_private_api.model.log_event import LogEvent
-from microsetta_private_api.model.source import Source
 from microsetta_private_api.model.project import Project
 from microsetta_private_api.model.daklapack_order import DaklapackOrder, \
     ORDER_ID_KEY, SUBMITTER_ACCT_KEY
@@ -18,15 +17,15 @@ from microsetta_private_api.repo.sample_repo import SampleRepo
 from microsetta_private_api.repo.source_repo import SourceRepo
 from microsetta_private_api.repo.transaction import Transaction
 from microsetta_private_api.repo.admin_repo import AdminRepo
-from microsetta_private_api.repo.survey_template_repo import SurveyTemplateRepo
 from microsetta_private_api.repo.metadata_repo import (retrieve_metadata,
                                                        drop_private_columns)
-from microsetta_private_api.repo.vioscreen_repo import VioscreenSessionRepo
-from microsetta_private_api.tasks import send_email as celery_send_email
+from microsetta_private_api.tasks import send_email as celery_send_email,\
+    per_sample_summary as celery_per_sample_summary
 from microsetta_private_api.admin.email_templates import EmailMessage
 from microsetta_private_api.util.redirects import build_login_redirect
 from microsetta_private_api.admin.daklapack_communication import \
     post_daklapack_order, send_daklapack_hold_email
+from microsetta_private_api.admin.sample_summary import per_sample
 from werkzeug.exceptions import Unauthorized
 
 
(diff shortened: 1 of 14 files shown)


microsetta_private_api/admin/admin_impl.py
```diff
@@ -397,83 +396,21 @@ def query_email_stats(body, token_info):
...
-                summary[status] = sample_status == status
-
-            summaries.append(summary)
-
-    return jsonify(summaries), 200
+def query_barcode_stats(body, token_info):
+    validate_admin_access(token_info)
+    barcodes = body["sample_barcodes"]
+    if len(barcodes) > 1000:
+        return jsonify({"message": "Too manny barcodes requested"}), 429
```
dhakim87, line 411: Dunno that this counts as Too many requests.  Just one really large one.  Maybe a 400 error code?

microsetta_private_api/db/patches/0080.sql
```diff
@@ -0,0 +1,2 @@
+-- Scans are often looked up by barcode
+CREATE INDEX barcode_scans_barcode_idx ON barcodes.barcode_scans (barcode);
```
dhakim87, line 2: Hahahaha, this is the downside of setting uuids as the primary keys to everything.  We forgot to index this one??

microsetta_private_api/repo/sample_repo.py
```diff
@@ -30,6 +30,31 @@ class SampleRepo(BaseRepo):
...
+        FROM ag.ag_kit_barcodes
+        LEFT JOIN (
+            SELECT barcode, max(scan_timestamp) AS scan_timestamp
+            FROM barcodes.barcode_scans
+            WHERE barcode = %s
+            GROUP BY barcode
+        ) latest_scan
+        ON ag.ag_kit_barcodes.barcode = latest_scan.barcode
+        LEFT JOIN ag.source
+        ON ag.ag_kit_barcodes.source_id = ag.source.id"""
```
dhakim87, line 56: Yeah these n+1 queries for getting scan timestamps are painful.  Does this resolve it well enough for you to be happy or do we need to think about other ways to store this?  Double storing the latest scan timestamp directly in the barcodes table would be messy on the code side, but it would be instantaneous on lookup rather than doing this nasty business in barcode_scans all the time

microsetta_private_api/admin/admin_impl.py
wasade: sounds good, thanks!

microsetta_private_api/db/patches/0080.sql
wasade: these are the barcodes themselves (eg 000004125) but ya i think so.

microsetta_private_api/tasks.py
```diff
@@ -14,9 +19,32 @@ def send_email(email, template_name, template_args):
...
+    template_args = {'date': date, 'project': project},
+    send_basic_email(email,
+                     f"[TMI-summary] Project {project}",
+                     'email/sample_summary',
+                     list(template_args),
+                     template_args,
+                     "EMAIL", "EMAIL_PER_PROJECT_SUMMARY",
+                     attachment_filepath=path,
+                     attachment_filename=filename)
+    os.remove(path)
```
dhakim87, line 50: I'm nervous about this.  Did you try this live?  Even if celery runs synchronously, I don't understand how the email could be sent synchronously, so I'm not sure when the attachment can be deleted.  If the email gets queued up somewhere, the attachment can't get deleted from disk until the email is sent out.

microsetta_private_api/repo/sample_repo.py
wasade: for the scenario where the barcode is known, i think this is fine. the nested query was doing a full table scan, so introducing this constraint limits to 1 row. from `explain analyze`, the difference was pretty clear, and overall the query time went from like 20ms per sample to 0.1 give or take (and on my puny laptop...)

microsetta_private_api/util/email.py
```diff
@@ -79,5 +93,13 @@ def send(cls, to, email_template, email_template_args=None, from_=None):
         message.attach(first)
         message.attach(second)
 
+        if attachment_filepath:
+            with open(attachment_filepath, 'rb') as f:
+                data = f.read()
+                attachment = MIMEApplication(data, Name=attachment_filename)
+            disposition = f'attachment; filename="{attachment_filename}"'
+            attachment['Content-Disposition'] = disposition
+            message.attach(attachment)
```
dhakim87, line 102: Does this copy the attached file into memory or is it still just a path to a file on disk?

microsetta_private_api/tasks.py
wasade: the contents of `path` are read into a buffer in memory within the `send` call itself which is executed synchronously. This point cannot be reached without that first 

...however as i type that out, i now think that an IO buffer would probably work here. how strong do you feel about exploring an alternative here?

dhakim87 (commented):
As long as you've tried the email sending on www-dev already I think it's good.  I'm nervous about the attachments still existing when the email is sent out due to the number of async components in between, but if it works, great.

microsetta_private_api/util/email.py
wasade: the `.read()` pulls into memory

wasade:
Cool thanks! those components in between are all used as sync within task, and attachment is in memory. But, am definitely verifying on www-dev...

wasade:
Just a note: the event log isn't getting triggered on this, and a little more jiggering is needed to make sure that occurs

wasade:
...can confirm email works from www-dev

microsetta_private_api/api/microsetta_private_api.yaml
```diff
@@ -1510,6 +1510,30 @@ paths:
               schema:
                 type: array
 
+  '/admin/account_project_barcode_summary':
+    post:
```
wasade, line 1514: add support for a query parameter to strip barcode information

wasade:
Verify source_email is being pulled as expected

PR #325, opened 2021-05-18
Save preferred language to account table

dhakim87:
I don't see an i18n branch in this repo, and this is going to break compatibility outside of the microsetta-interface i18n branch, so... maybe we should make an i18n branch on this side to merge this into?  

Private api side of setting language preference

### DIFF
--- a/microsetta_private_api/api/tests/test_api.py
+++ b/microsetta_private_api/api/tests/test_api.py
@@ -57,7 +57,7 @@ DUMMY_ACCT_INFO = {
     "email": TEST_EMAIL,
     "first_name": "Jane",
     "last_name": "Doe",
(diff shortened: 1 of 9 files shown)


microsetta_private_api/api/tests/test_api.py
```diff
@@ -601,6 +604,8 @@ def run_query_and_content_required_field_test(self, url, action,
 
                 curr_query_str = urlencode(curr_query_dict)
                 curr_content_json = json.dumps(curr_content_dict)
+
+                print(curr_content_json)
```
wasade, line 608: Could this be removed?

wasade (changes requested):

Issue #321, opened 2021-05-05
Generate translations for emails


wasade:
...and the associated directory structure

wasade:
Fixed in #339

Issue #320, opened 2021-05-05
Add flask-babel


wasade:
...per the title

wasade:
Fixed in #339

PR #339, opened 2021-06-04
Babel

### LINKED ISSUE #319: Add gettext / ngettext to email templates
...per the title

wasade:
Fixes #319, #320, #321


microsetta_private_api/server.py
```diff
@@ -74,6 +76,14 @@ def build_app():
     global babel
     babel = Babel(app.app)
 
+    @babel.localeselector
+    def get_locale():
+        # for unit test support
+        if not flask.has_request_context():
+            return EN_US
+
+        return request.accept_languages.best_match([EN_US, ES_MX])
```
dhakim87, line 85: This is probably wrong for the API, since requests come from microsetta-interface, the accept languages will always be the same regardless of the user.  But that just means we need to grab the lang_tag query parameter and override here.

dhakim87 (approved):
I want this in so I can use the babel integration.  I think the language selector is wrong, but easily updated after this goes in anyway.  

Hooray for the death of .plain email templates.

microsetta_private_api/server.py
wasade: Ya I wasn't enitrely sure how this would need to look, should we cram the preferred language (if account ID exists) into the request object and test against it here?

PR #341, opened 2021-06-04
Fix Country Sorting on Survey in Spanish

cassidysymons:


microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -147,6 +147,10 @@ def get_survey_template(self, survey_id, language_tag):
                                                                language_tag)
                 triggers = self._get_question_triggers(question_id)
 
+                # Quick and dirty fix to correctly sort country names in Spanish
+                if language_tag == "es_MX" and (question_id == 110 or question_id == 148):
```
wasade, line 151: Would it be possible to test against `microsetta_private_api.localization.ES_MX`?

microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -147,6 +147,10 @@ def get_survey_template(self, survey_id, language_tag):
...
+                    responses[1:len(responses)] = sorted(responses[1:len(responses)])
```
wasade, line 152: Just to check, does the participants response still get stored in the DB as expected?

wasade: ...nm this should be good

PR #352, opened 2021-06-17
database methods needed for forthcoming daklapack order polling

AmandaBirmingham:
I'm trying to add the daklapack order polling functionality in chunks to allow reasonable code-reviewing.  This is the first installment, which includes db methods in admin_repo.py to get and set info about already-submitted orders.

### DIFF
--- a/microsetta_private_api/admin/tests/test_admin_repo.py
+++ b/microsetta_private_api/admin/tests/test_admin_repo.py
@@ -885,3 +885,113 @@ class AdminRepoTests(AdminTests):
 
         # NB: all the above happens within a transaction that we then DO NOT
         # commit so the db changes are not permanent
+
+    def make_dummy_dak_orders(self, t):
+        # create some orders;
+        # need a valid submitter id from the account table to input
+        with t.dict_cursor() as cur:
+            cur.execute("SELECT id, first_name, last_name "
+                        "FROM ag.account "
+                        "WHERE account_type = 'admin' "
+                        "ORDER BY id "
+                        "LIMIT 1;")
+            submitter_record = cur.fetchone()
+            submitter_id = submitter_record[0]
+
+            dummy_orders = [('7ed917ef-0c4d-431a-9aa0-0a1f4f41f44b',
+                             submitter_id, 'dummy 1', None,
+                             '{"orderId": '
+                             '"7ed917ef-0c4d-431a-9aa0-0a1f4f41f44b"}',
+                             dateutil.parser.isoparse(
+                                 "2020-10-09T22:43:52.219328Z"),
+                             None, "Sent"),
+                            ('8ed917ef-0c4d-431a-9aa0-0a1f4f41f44b',
+                             submitter_id, 'dummy 2', None,
+                             '{"orderId": '
+                             '"8ed917ef-0c4d-431a-9aa0-0a1f4f41f44b"}',
+                             dateutil.parser.isoparse(
+                                 "2021-10-09T22:43:52.219328Z"),
+                             None, "Pending"),
+                            ('9ed917ef-0c4d-431a-9aa0-0a1f4f41f44b',
+                             submitter_id, 'dummy 3', None,
+                             '{"orderId": '
+                             '"9ed917ef-0c4d-431a-9aa0-0a1f4f41f44b"}',
+                             dateutil.parser.isoparse(
+                                 "2020-10-09T22:43:52.219328Z"),
+                             None, "Error"),
+                            ('0ed917ef-0c4d-431a-9aa0-0a1f4f41f44b',
+                             submitter_id, 'dummy 4', None,
+                             '{"orderId": '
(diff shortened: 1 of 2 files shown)


microsetta_private_api/admin/tests/test_admin_repo.py
```diff
@@ -885,3 +885,113 @@ def test_create_daklapack_order(self):
 
         # NB: all the above happens within a transaction that we then DO NOT
         # commit so the db changes are not permanent
+
+    def make_dummy_dak_orders(self, t):
+        # create some orders;
+        # need a valid submitter id from the account table to input
```
dhakim87, line 891: This is kind of weird.  I think there is no admin on a fresh dev install/db rebuild, which is going to be annoying if the tests start failing.  Not the end of the world though.

microsetta_private_api/repo/admin_repo.py
```diff
@@ -1108,3 +1108,41 @@ def create_daklapack_order(self, daklapack_order):
...
         return order_id
+
+    def get_unfinished_daklapack_order_ids(self):
+        with self._transaction.dict_cursor() as cur:
+            cur.execute(
+                "SELECT dak_order_id "
+                "FROM "
+                "barcodes.daklapack_order "
+                "WHERE last_polling_status NOT IN (%s, %s) "
+                "OR last_polling_status IS NULL "
```
dhakim87, line 1119: NULL is considered IN ("Sent", "Error") ???  That's super annoying.

dhakim87, line 1119: I feel like NULL is already NOT IN ("Sent", "Error"), does SQL disagree?

microsetta_private_api/repo/admin_repo.py
```diff
@@ -1108,3 +1108,41 @@ def create_daklapack_order(self, daklapack_order):
...
+                "ORDER BY last_polling_timestamp DESC;",
+                ("Sent", "Error"))
+            rows = cur.fetchall()
+            return [x[0] for x in rows]
+
+    def get_projects_for_dak_order(self, dak_order_id):
```
dhakim87, line 1125: project is like American Gut Project, British Gut Project?  

What does it mean if a daklapack order is in multiple projects?  Is that okay or is that weird?

dhakim87 (approved):
Welcome back!  Seems generally good, I don't know a ton (/anything at all) about how daklapack ordering works.  

Totally understand use cases for finding unfinished orders and updating status on orders after polling, not sure I get the relationship between projects and daklapack orders.  If that relationship is many to one, who pays and what instruction booklet do we put in the kit?

microsetta_private_api/admin/tests/test_admin_repo.py
AmandaBirmingham: Haha, happily we (you :D ) already thought of that issue at some past point!  The `setup_test_data` method for the `test_admin_repo.py` unit tests sets up a new Dan H user who has admin privileges at the beginning of every test.

microsetta_private_api/repo/admin_repo.py
AmandaBirmingham: It sounded kind of weird to me at first, but it is in fact a hard requirement.  A daklapack order maps to a kit (or a set of kits all having the exact same properties), and a kit can be assigned to multiple projects (e.g., both American Gut and one of the American gut sub-projects).  Also, apparently which project(s) a daklapack order is attached to can have an effect on who has to pay for it :D

microsetta_private_api/repo/admin_repo.py
AmandaBirmingham: :D Yes, I too thought that NULL would covered by "not in sent/error", but postgresql apparently gets all snottily semantic about "you asked me to return *values* other than sent and error and NULL isn't a value, it is a lack of a value, yadda, yadda" ... I can sadly guarantee from my own debugging time that this query does not return nulls without that last OR clause :-|  I found this that suggests I'm not alone: https://stackoverflow.com/questions/41342593/postgresql-how-to-return-null-values-in-not-in-expression

AmandaBirmingham:
@dhakim87 Re "If that [project-to-order] relationship is many to one, who pays and what instruction booklet do we put in the kit?": I don't know anything about who pays, but I think I can answer the part about the instruction booklet.  A given daklapack order is for a single, specific "article" as defined in the `barcodes.daklapack_articles` table, and that table specifies which instruction booklet(s) and which registration card to provide.  

Thinking about it, I can see that this situation allows for the possible error case that an admin manually submitting a daklapack order COULD select an article that specifies sending American Gut instructions but s/he could fail to specify that the order is in fact associated with the American Gut project.  (This shouldn't be a realistic error case for daklapack orders submitted automatically from Fundrazr submissions since the project info will be assigned to the order programmatically, so hopefully we will set that up right once and not have to worry about it again.) I had not acknowledged the possibility for that particular kind of screw-up before, but I think it is low on my list of errors to guard against right now :D

PR #356, opened 2021-07-01
Extend kit creation code, barcodes.kit table for daklapack kits

AmandaBirmingham:
Refactor most of `create_kits` function in `admin_repo.py` into private functions that can be used by both `create_kits` and a new `create_kit` (note singular :) function that takes in barcodes, box id, and kit name specified by Daklapack (instead of generating them internally). Extend `barcodes.kit` table to contain a box id separate from a kit id as well as two different fedex tracking numbers (instead of one).

### DIFF
--- a/microsetta_private_api/db/patches/0083.sql
+++ b/microsetta_private_api/db/patches/0083.sql
@@ -1,0 +1,10 @@
+-- rename the "fedex_tracking" column to specify it is the *outbound* fedex
+-- tracking code, and add an inbound fedex tracking code column too
+-- add a new "box_id" column to the kit table and populate it
+-- (for all records to date) with the kit uuid;
+-- require it be not null
+ALTER TABLE barcodes.kit RENAME COLUMN fedex_tracking TO outbound_fedex_tracking;
+ALTER TABLE barcodes.kit ADD COLUMN inbound_fedex_tracking varchar NULL;
+ALTER TABLE barcodes.kit ADD COLUMN box_id varchar NULL;
+UPDATE barcodes.kit SET box_id = kit_uuid;
+ALTER TABLE barcodes.kit ALTER COLUMN box_id SET NOT NULL;
--- a/microsetta_private_api/repo/admin_repo.py
+++ b/microsetta_private_api/repo/admin_repo.py
@@ -1108,3 +1108,41 @@ class AdminRepo(BaseRepo):
                                            template=None, page_size=100)
 
         return order_id
+
+    def get_unfinished_daklapack_order_ids(self):
+        with self._transaction.dict_cursor() as cur:
+            cur.execute(
+                "SELECT dak_order_id "
+                "FROM "
+                "barcodes.daklapack_order "
+                "WHERE last_polling_status NOT IN (%s, %s) "
+                "OR last_polling_status IS NULL "
+                "ORDER BY last_polling_timestamp DESC;",
+                ("Sent", "Error"))
+            rows = cur.fetchall()
+            return [x[0] for x in rows]
+
+    def get_projects_for_dak_order(self, dak_order_id):
+        with self._transaction.dict_cursor() as cur:
+            cur.execute(
+                "SELECT project_id "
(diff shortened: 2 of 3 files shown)


microsetta_private_api/db/patches/0083.sql
```diff
@@ -0,0 +1,10 @@
+-- rename the "fedex_tracking" column to specify it is the *outbound* fedex
+-- tracking code, and add an inbound fedex tracking code column too
+-- add a new "box_id" column to the kit table and populate it
+-- (for all records to date) with the kit uuid;
+-- require it be not null
+ALTER TABLE barcodes.kit RENAME COLUMN fedex_tracking TO outbound_fedex_tracking;
+ALTER TABLE barcodes.kit ADD COLUMN inbound_fedex_tracking varchar NULL;
+ALTER TABLE barcodes.kit ADD COLUMN box_id varchar NULL;
+UPDATE barcodes.kit SET box_id = kit_uuid;
```
wasade, line 9: Should we also drop the column `kit_uuid`?

microsetta_private_api/repo/admin_repo.py
```diff
@@ -732,8 +732,18 @@ def create_kits(self, number_of_kits, number_of_samples, kit_prefix,
         project_ids : list of int
             Project ids the samples are to be associated with
         """
-        # int ids come in as strings ...
-        project_ids = [int(x) for x in project_ids]
+
+        kit_names = self._generate_novel_kit_names(number_of_kits, kit_prefix)
+        kit_name_and_barcode_tuples_list, new_barcodes = \
+            self._generate_novel_barcodes(
```
wasade, line 738: @dhakim87, do you think these should use `lock`?

microsetta_private_api/repo/admin_repo.py
```diff
@@ -783,73 +808,137 @@ def create_kits(self, number_of_kits, number_of_samples, kit_prefix,
...
+            # create kits in kit table
+            new_kit_uuids = [str(uuid.uuid4()) for x in kit_names]
+            if box_ids is None:
+                box_ids = new_kit_uuids
+            barcode_kit_inserts = list(zip(new_kit_uuids, kit_names, box_ids))
             cur.executemany("INSERT INTO barcodes.kit "
-                            "(kit_id) "
-                            "VALUES (%s)", [(n, ) for n in names])
+                            "(kit_uuid, kit_id, box_id) "
+                            "VALUES (%s, %s, %s)", barcode_kit_inserts)
```
wasade, line 861: Do the fedex tracking numbers come later?

microsetta_private_api/repo/admin_repo.py
dhakim87: Not sure- I don't think they need it?  What resource do you think needs to be locked/what other operation would this conflict with?  

I think the postgres lock mechanism should be thought of as using synchronized/lock blocks in java or C#: we need lock access to particular resource when there are multiple workflows that all need to read and then write based on what was snapshotted, and all access to that resource must be locked.

microsetta_private_api/repo/admin_repo.py
AmandaBirmingham: @wasade Good catch!  I had failed to unshelve one set of changes for this PR <facepalm>, which bring in address and fedex tracking numbers ... they are here now :)

microsetta_private_api/db/patches/0083.sql
wasade: oh I see, that makes sense!

microsetta_private_api/repo/admin_repo.py
wasade: Okay, so assuming we rarely if ever expect concurrent transactions, and the likelihood of a conflict in ID generation is low, then we should be good

microsetta_private_api/repo/admin_repo.py
wasade: thanks!

PR #360, opened 2021-07-22
Daklapack order status checking

AmandaBirmingham:
This code provides the logic of checking the status of outstanding daklapack orders via the daklapack api, recording the current status to our daklapack orders table, creating a new kit for any order that was successfully sent, and gathering descriptive info about any order that errored out.

Note that this code does not include (a) the celery-based scheduling of this status-checking functionality or (b) the (again celery-based) emailing of the descriptive info about any errored orders to someone at microsetta who has to deal with them.  Those code pieces are ready to go and I will put them up as their own PR once this is reviewed, since this seems meaty enough by itself.

This also does not include the automated cancellation with daklapack of any errored order.  I need a bit of help from Bas at daklapack to finish that code off.

### DIFF
--- a/microsetta_private_api/admin/daklapack_communication.py
+++ b/microsetta_private_api/admin/daklapack_communication.py
@@ -68,5 +68,10 @@ def _get_from_daklapack_api(url_suffix):
     oauth_session = _get_daklapack_oauth2_session()
     dak_order_get_url = f"{SERVER_CONFIG['daklapack_api_base_url']}" \
                         f"{url_suffix}"
+
     result = oauth_session.get(dak_order_get_url, headers=DAK_HEADERS)
+    if result.status_code >= 300:
+        raise ValueError(f"Getting {url_suffix} received status code "
+                         f"{result.status_code}: {result.json}")
+
     return result
--- a/microsetta_private_api/admin/daklapack_communication.py
+++ b/microsetta_private_api/admin/daklapack_communication.py
@@ -51,3 +51,22 @@ def send_daklapack_hold_email(daklapack_order):
         email_success = False
 
     return email_success
+
+
+def get_daklapack_orders_status(page_num):
+    return _get_from_daklapack_api(f"/api/orders?Page={page_num}")
+
+
+def get_daklapack_order_details(order_id):
+    return _get_from_daklapack_api(f"/api/orders/{order_id}")
+
+
+def _get_from_daklapack_api(url_suffix):
+    oauth_session = _get_daklapack_oauth2_session()
+    dak_order_get_url = f"{SERVER_CONFIG['daklapack_api_base_url']}" \
+                        f"{url_suffix}"
+    result = oauth_session.get(
+        dak_order_get_url,
+        headers={SERVER_CONFIG["daklapack_subscription_key_name"]:
+                 SERVER_CONFIG["daklapack_subscription_key_val"]})
+    return result
--- a/microsetta_private_api/admin/daklapack_communication.py
+++ b/microsetta_private_api/admin/daklapack_communication.py
@@ -3,6 +3,11 @@ from requests_oauthlib import OAuth2Session
 from microsetta_private_api.config_manager import SERVER_CONFIG
 from microsetta_private_api.tasks import send_basic_email as celery_send_email
 
+DAK_HEADERS = {
+    SERVER_CONFIG["daklapack_subscription_key_name"]:
+        SERVER_CONFIG["daklapack_subscription_key_val"]
+}
+
 ORDER_HOLD_TEMPLATE_PATH = "email/daklapack_fulfillment_hold_request"
 
 
--- a/microsetta_private_api/admin/daklapack_communication.py
+++ b/microsetta_private_api/admin/daklapack_communication.py
@@ -28,9 +28,7 @@ def post_daklapack_order(payload):
     dak_order_post_url = f"{SERVER_CONFIG['daklapack_api_base_url']}" \
                          f"/api/orders"
     result = oauth_session.post(
-        dak_order_post_url, json=payload,
-        headers={SERVER_CONFIG["daklapack_subscription_key_name"]:
-                 SERVER_CONFIG["daklapack_subscription_key_val"]})
+        dak_order_post_url, json=payload, headers=DAK_HEADERS)
     return result
 
 
@@ -54,7 +52,7 @@ def send_daklapack_hold_email(daklapack_order):
 
 
 def get_daklapack_orders_status(page_num):
-    return _get_from_daklapack_api(f"/api/orders?Page={page_num}")
+    return _get_from_daklapack_api(f"/api/orders?Page={int(page_num)}")
 
 
 def get_daklapack_order_details(order_id):
@@ -65,8 +63,5 @@ def _get_from_daklapack_api(url_suffix):
     oauth_session = _get_daklapack_oauth2_session()
     dak_order_get_url = f"{SERVER_CONFIG['daklapack_api_base_url']}" \
                         f"{url_suffix}"
-    result = oauth_session.get(
-        dak_order_get_url,
-        headers={SERVER_CONFIG["daklapack_subscription_key_name"]:
-                 SERVER_CONFIG["daklapack_subscription_key_val"]})
+    result = oauth_session.get(dak_order_get_url, headers=DAK_HEADERS)
     return result
--- a/microsetta_private_api/admin/daklapack_polling.py
+++ b/microsetta_private_api/admin/daklapack_polling.py
@@ -16,6 +16,7 @@ def poll_dak_orders():
 
     with Transaction() as t:
         poll_dak_orders_using_transaction(t)
+        t.commit()
 
 
 # This method exists (as a separate entity from poll_dak_orders)
--- a/microsetta_private_api/admin/daklapack_polling.py
+++ b/microsetta_private_api/admin/daklapack_polling.py
@@ -1,3 +1,4 @@
+import json
 from datetime import datetime
 from microsetta_private_api.repo.transaction import Transaction
 from microsetta_private_api.repo.admin_repo import AdminRepo
@@ -212,6 +213,14 @@ def _store_single_sent_kit(admin_repo, order_proj_ids, single_article_dict):
     created_kit_info = admin_repo.create_kit(
         kit_name, box_id, address_dict, outbound_fedex_code,
         inbound_fedex_code, device_barcodes, order_proj_ids)
+
+    # addresses returned from the create are strings (that in this case
+    # hold representation of json); return them as json instead.
+    # Not doing graceful error handling here because if any of these
+    # keys/structures don't exist, something is wrong and we *should* error
+    for i in range(len(created_kit_info["created"])):
+        address_str = created_kit_info["created"][i]["address"]
+        created_kit_info["created"][i]["address"] = json.loads(address_str)
 
     return created_kit_info
 
--- a/microsetta_private_api/admin/daklapack_polling.py
+++ b/microsetta_private_api/admin/daklapack_polling.py
@@ -1,0 +1,208 @@
+from datetime import datetime
+from microsetta_private_api.repo.transaction import Transaction
+from microsetta_private_api.repo.admin_repo import AdminRepo
+from microsetta_private_api.admin import daklapack_communication as dc
+
+OUTBOUND_DEV_KEY = "outBoundDelivery"
+INBOUND_DEV_KEY = "inBoundDelivery"
+COLLECTION_DEVICE_TYPES = ["2point5ml_etoh_tube", "7ml_etoh_tube",
+                           "neoteryx_kit"]
+SENT_STATUS = "Sent"
+ERROR_STATUS = "Error"
+
+
+def poll_dak_orders():
+    """Get open orders' status from daklapack and process accordingly"""
+
+    with Transaction() as t:
+        poll_dak_orders_using_transaction(t)
+
+
+# This method exists (as a separate entity from poll_dak_orders)
+# exists only to make testing easier
+def poll_dak_orders_using_transaction(t):
+    """Use transaction to get open orders' status from daklapack and process"""
+
+    results = {SENT_STATUS: [], ERROR_STATUS: []}
+    admin_repo = AdminRepo(t)
+
+    # from our db, get list of daklapack orders that haven't been completed
+    open_dak_order_ids = admin_repo.get_unfinished_daklapack_order_ids()
+    curr_page = 0
+
+    # as long as there are still orders we need status info on
+    while len(open_dak_order_ids) > 0:
+        curr_page += 1
+
+        # call daklapack api to get another "page" of order statuses
+        dak_orders_response = dc.get_daklapack_orders_status(curr_page)
+        dak_orders_data = dak_orders_response.json["data"]
+
+        # for each order for which daklapack provided a status
+        for curr_datum in dak_orders_data:
+            curr_order_id = curr_datum["orderId"]
+
+            # ignore any order that is not an open order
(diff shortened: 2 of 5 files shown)


microsetta_private_api/admin/daklapack_polling.py
```diff
@@ -0,0 +1,208 @@
...
+
+                # note extend not append--avoiding nested lists
+                results[curr_status].extend(per_article_info)
+
+                # if curr_status == ERROR_STATUS:
+                # TODO: cancel the errored order w daklapack
+                # end if error
+            # end if error or sent
+
+            # remove current order id from the list of order ids that we
```
dhakim87, line 69: This is kind of lame- I take it their api does not allow filtering the orders we ask for in any way, so looping over pages of them is the only way?  

Is any kind of sorting available?  Even a sort by most recent order would put the unprocessed ones at the front of the list and save on the number of pages we have to grab if there's lots of orders.

microsetta_private_api/admin/daklapack_communication.py
```diff
@@ -51,3 +51,22 @@ def send_daklapack_hold_email(daklapack_order):
         email_success = False
 
     return email_success
+
+
+def get_daklapack_orders_status(page_num):
+    return _get_from_daklapack_api(f"/api/orders?Page={page_num}")
```
dhakim87, line 57: This is a little weird, but page_num is not escaped- if someone sends page_num="17&xxx=yyy", it lets them modify what gets sent (with our credentials) to the daklapack /api/orders endpoint.  I don't think anything needs to be changed here, but it must be the case that this function can only be called by administrator users (or celery tasks or whatever)- it cannot be exposed to api users or allows a minor privilege escalation.

dhakim87: Just a straight int(page_num) cast would fix the potential security non issue, as trivial as it is.

microsetta_private_api/admin/daklapack_communication.py
```diff
@@ -51,3 +51,22 @@ def send_daklapack_hold_email(daklapack_order):
...
+    return _get_from_daklapack_api(f"/api/orders/{order_id}")
+
+
+def _get_from_daklapack_api(url_suffix):
+    oauth_session = _get_daklapack_oauth2_session()
+    dak_order_get_url = f"{SERVER_CONFIG['daklapack_api_base_url']}" \
+                        f"{url_suffix}"
+    result = oauth_session.get(
+        dak_order_get_url,
+        headers={SERVER_CONFIG["daklapack_subscription_key_name"]:
```
dhakim87, line 70: What do they need this header for, I thought the connections used oauth2???  Also, if header construction is part of every request, might be worth wrapping, but only at 2 usages so far so meh.

microsetta_private_api/admin/daklapack_polling.py
```diff
@@ -0,0 +1,209 @@
...
+SENT_STATUS = "Sent"
+ERROR_STATUS = "Error"
+
+
+def poll_dak_orders():
+    """Get open orders' status from daklapack and process accordingly"""
+
+    with Transaction() as t:
+        poll_dak_orders_using_transaction(t)
+        t.commit()
```
dhakim87, line 19: 👍

microsetta_private_api/admin/daklapack_polling.py
```diff
@@ -0,0 +1,209 @@
...
+
+    results = {SENT_STATUS: [], ERROR_STATUS: []}
+    admin_repo = AdminRepo(t)
+
+    # from our db, get list of daklapack orders that haven't been completed
+    open_dak_order_ids = admin_repo.get_unfinished_daklapack_order_ids()
+    curr_page = 0
+
+    # as long as there are still orders we need status info on
+    while len(open_dak_order_ids) > 0:
```
dhakim87, line 35: What happens if we believe there is an order id that they have never heard of?  Does this just infinite loop forever asking for page after page after page?  I think this while loop should be a for loop that loops through all of their pages (assuming there is a way to ask how many pages there are?)

microsetta_private_api/admin/daklapack_polling.py
```diff
@@ -0,0 +1,209 @@
...
+        curr_page += 1
+
+        # call daklapack api to get another "page" of order statuses
+        dak_orders_response = dc.get_daklapack_orders_status(curr_page)
```
dhakim87, line 39: Is there a way to check if such a page will exist?  What happens if we ask for page 15 but there are only 10 pages?

microsetta_private_api/admin/daklapack_polling.py
```diff
@@ -0,0 +1,209 @@
...
+            if status == SENT_STATUS:
+                # (Per Daniel 2021-07-01, each instance of a daklapack article
+                # represents exactly one kit, no more or less.)
+                curr_output = _store_single_sent_kit(
+                    admin_repo, order_proj_ids, curr_article_instance)
+            elif status == ERROR_STATUS:
+                curr_output = _gather_article_error_info(
+                    order_id, create_date, curr_article_instance)
+            else:
+                raise ValueError(f"Order {order_id} has an unexpected status: "
```
dhakim87, line 123: Recommend adding a try_catch to poll_dak_orders?  If this case is hit, the transaction won't be committed- nothing will be written to the database, and if this is all run by a background celery worker, I don't know that anybody will be checking logs regularly or which log this would even end up in, so I don't know how we track down error cases like this.  

Same issue may exist for any network errors occurring in communications with daklapack?  May be worth writing status info for all the things we are able to pull, but try catch at the level of individual orders or pages to ensure that we don't quit the worker if one order fails?

microsetta_private_api/admin/daklapack_polling.py
```diff
@@ -0,0 +1,209 @@
...
+        # complexity, like a populated containerItems list that itself
+        # contains scannable items, on to infinity.  HOWEVER, microsetta
+        # has defined each article to equal exactly one kit, so it should
+        # not be necessary to dig into that.
+
+        # figure out what *kind* of barcoded thing this is and capture
+        # its barcode to the right field if it is a kind we care about
+        curr_scannable_type = curr_scannable["type"]
+        curr_barcode = curr_scannable["barcode"]
+        if curr_scannable_type in COLLECTION_DEVICE_TYPES:
```
dhakim87, line 172: Is it the case that every scannable kit will have exactly 1 thing of type box and 1 thing of type registration card?  If not, this code is hopelessly wrong.

dhakim87 (approved):
I think this covers all of the happy cases.  I tend not to trust network code not to have obscure error cases though, so I would worry a bit about daklapacks api returning unhandled error codes or http statuses failing out or just having run of the mill network failures.  

The concurrency guarantees made by their api aren't clear to me- is it possible for the same order to appear twice on different pages if something is added before it while you're processing a page?  Is it possible for an order to be skipped as you cross pages if an order is deleted?  I don't think any of that really matters all too much as long as this script is run consistently in the background, but that's a good reason to ensure that any failure cases bubble up to the top of the celery worker and get logged in a place someone is looking at them.  

The biggest worry for me is that some network failure or database inconsistency silently crashes or spinlocks the celery worker.  As long as that type of case is properly handled, I'm happy with these changes.

microsetta_private_api/admin/daklapack_polling.py
AmandaBirmingham: No, their api does not allow us to filter the list of orders by providing a bulk list of ids we care about. Now that I look at it again, their pre-release documentation does says that we can filter by status, but doesn't provide syntax ... I will ask our contact there for details.  Fortunately, he has previously confirmed that the *default* sort for the orders get is that latest orders are shown first, so we should be seeing the orders we are most likely to need new status info on first.

microsetta_private_api/admin/daklapack_communication.py
AmandaBirmingham: nice catch!  int(page_num) it is :)

microsetta_private_api/admin/daklapack_communication.py
AmandaBirmingham: What they need it for I do not know, but they seem very clear on the fact that they do need it ... I will break it out so it isn't repeated.

microsetta_private_api/admin/daklapack_polling.py
AmandaBirmingham: Ugh, good point!  They have not revealed to me a way to find out how many pages there are ... maybe their endpoint will error if I ask for more pages than they have?  You are quite correct that I'd better find out and figure out how to prevent infinite looping.

microsetta_private_api/admin/daklapack_polling.py
AmandaBirmingham: @dhakim87 That is my understanding; @wasade can you confirm or deny that we can expect that every kit ordered from daklapack will have exactly one box and exactly one registration card?

In theory this code would be ok with there being *no more than one* of each ... however, while we can provide None for a kit's box id (in which case it will be set to the kit's uuid), passing a None for the kit id will cause the database insert to fail as that can't be null and has no default ...

AmandaBirmingham:
@dhakim87  thank you for the very useful comments! @wasade big picture, how would you like errors from this background process to be handled?  Is there a log they should be routed to that will get regular attention?  Should they be collected and emailed somewhere the way the order failures are? @dhakim87  how do you handle errors bubbling up from the celery processes you run in the background?  Should I be following whatever model you use?

microsetta_private_api/admin/daklapack_polling.py
wasade: Each kit will have one box and one registration card

microsetta_private_api/admin/daklapack_polling.py
```diff
@@ -0,0 +1,209 @@
...
+                       "article_code": curr_article_instance["articleCode"],
+                       "sent_to_address":
+                           curr_article_instance["sendInformation"],
+                       # per specifications, orders submitted by microsetta api
+                       # through the daklapack interface put the name of the
+                       # microsetta user who submitted the order into the
+                       # order's "companyName" field
+                       "order_submitter":
+                           curr_article_instance["sendInformation"]
+                           ["companyName"],
```
wasade, line 203: Is this an implicit string concatenation or a getitem using `companyName`?

microsetta_private_api/admin/tests/test_daklapack_polling.py
```diff
@@ -0,0 +1,724 @@
...
+                      'firstName': 'John J Doe',
+                      'insertion': '',
+                      'lastName': '',
+                      'phone': '505-784-5252',
+                      'postalCode': '88103',
+                      'state': 'NM'},
+                  'status_description': None}],
+            'Sent':
+                [{'created': [
+                    {'address': '{"firstName": "Natalia J Phillips", '
```
wasade, line 485: The expected value associated with the key `address` is of `str` not `dict`?

wasade (commented):
Thanks @AmandaBirmingham!!! Mostly just questions

microsetta_private_api/admin/daklapack_polling.py
AmandaBirmingham: I've checked the latest api docs, and we still can't filter by a list of order numbers and sorting by status, while possible, won't help us.  I think this approach, with the default sort (which puts the most recent orders first) is the best we're gonna get.

microsetta_private_api/admin/daklapack_polling.py
AmandaBirmingham: I've verified that asking for a page number greater than the actual number of pages of orders there are returns a response with an empty data list.  I've added code to break out when that happens, even if there are order ids we still haven't found.

microsetta_private_api/admin/daklapack_polling.py
AmandaBirmingham: See new findings above :)

microsetta_private_api/admin/daklapack_polling.py
AmandaBirmingham: Ok, good to have it confirmed that every kit should have exactly one kit and one registration card.  I have added a little logic to ensure the code throws an error if daklapack passes us kit info that has *more than one* of either of these values ... should not ever happen but if it does we probably want to know because something is agley.

microsetta_private_api/admin/daklapack_polling.py
```diff
@@ -0,0 +1,209 @@
...
+def poll_dak_orders():
+    """Get open orders' status from daklapack and process accordingly"""
+
+    with Transaction() as t:
+        poll_dak_orders_using_transaction(t)
+        t.commit()
+
+
+# This method exists (as a separate entity from poll_dak_orders)
+# exists only to make testing easier
```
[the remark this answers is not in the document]
AmandaBirmingham: This method ended up being removed in the refactoring for more robust error-case handling.

microsetta_private_api/admin/tests/test_daklapack_polling.py
AmandaBirmingham: Correct because this return value is pulled out of the new records added to the db, and the "address" field of the barcodes.kit table predates daklapack and is defined as a varchar, not a jsonb.  We could redefine it if you like, but we'd presumably have to make sure any existing data was changed into json format.  Alternately, I could turn the value from the db into a dict before returning it. Thoughts?

microsetta_private_api/admin/daklapack_polling.py
AmandaBirmingham: a getItem, getting the ["companyName"] field in the dictionary stored in curr_article_instance["sendInformation"]

microsetta_private_api/admin/daklapack_communication.py
```diff
@@ -51,3 +54,19 @@ def send_daklapack_hold_email(daklapack_order):
...
+
+def get_daklapack_order_details(order_id):
+    return _get_from_daklapack_api(f"/api/orders/{order_id}")
+
+
+def _get_from_daklapack_api(url_suffix):
+    oauth_session = _get_daklapack_oauth2_session()
+    dak_order_get_url = f"{SERVER_CONFIG['daklapack_api_base_url']}" \
+                        f"{url_suffix}"
+    result = oauth_session.get(dak_order_get_url, headers=DAK_HEADERS)
```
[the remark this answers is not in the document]
AmandaBirmingham: I added code to check for >2XX status codes and, if detected, raise an error (which should then be handled by the error handling in the calling functions).

microsetta_private_api/admin/daklapack_polling.py
```diff
@@ -0,0 +1,209 @@
...
+    open_dak_order_ids = admin_repo.get_unfinished_daklapack_order_ids()
+    curr_page = 0
+
+    # as long as there are still orders we need status info on
+    while len(open_dak_order_ids) > 0:
+        curr_page += 1
+
+        # call daklapack api to get another "page" of order statuses
+        dak_orders_response = dc.get_daklapack_orders_status(curr_page)
+        dak_orders_data = dak_orders_response.json["data"]
```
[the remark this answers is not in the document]
AmandaBirmingham: Now >2XX error will cause get_daklapack_orders_status to raise ValueError which will be caught by try/except

microsetta_private_api/admin/daklapack_polling.py
AmandaBirmingham: Lots more error handling added.  Errors collected for emailing.

microsetta_private_api/admin/tests/test_daklapack_polling.py
wasade: I think the last option, casting to `dict` before `return` is easiest here and is likely what a consumer would want. Would that be quick to change?

AmandaBirmingham: Done

PR #364, opened 2021-08-10
Address Verification via Melissa

cassidysymons:
Added functionality to verify addresses using Melissa's global address API


microsetta_private_api/api/microsetta_private_api.yaml
```diff
@@ -1144,6 +1144,57 @@ paths:
         '404':
           $ref: '#/components/responses/404NotFound'
 
+  '/admin/verify_address':
+    get:
+      operationId: microsetta_private_api.admin.admin_impl.address_verification
+      tags:
+        - Admin
+      summary: Verify a postal address
```
AmandaBirmingham, line 1152: I think this is a little unclear to a user not immersed in the address verification process.  "Verify" what?  Perhaps "Verify the correctness of a postal address" or "Verify that a postal address really exists" or "Verify that postal address is in the required format for FedEx" or ...?  (I don't know which it is :)

microsetta_private_api/api/microsetta_private_api.yaml
```diff
@@ -1144,6 +1144,57 @@ paths:
...
+          description: Second line of address to verify
+          schema:
+            type: string
+        - in: query
+          name: city
+          description: City of address to verify
+          schema:
+            type: string
+        - in: query
+          name: state
```
AmandaBirmingham, line 1171: What should be passed in if the country in question does not have states/provinces and/or doesn't include them in their mailing addresses?

microsetta_private_api/repo/melissa_repo.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def create_record(self, address_1, address_2, city, state, postal, \
+        country):
+        """
+        Create a record before pinging Melissa so we can include unique ID
+
+        Parameters
+        ----------
+        address_1 - Primary street address
+        address_2 - Secondary street address
```
AmandaBirmingham, line 15: Some indication/example of what goes in the primary and in the secondary would be a nice-to-have.  As a potential user, I'd also love some guidance on which of these fields is required (e.g., can I have an address_2 without an address_1?) and what to pass in for the ones that are optional (presumably state and postal code, as those aren't universally used ... is address_2 also optional?)

microsetta_private_api/db/patches/0084.sql
```diff
@@ -0,0 +1,24 @@
+CREATE TABLE ag.melissa_address_queries (
+id uuid PRIMARY KEY DEFAULT uuid_generate_v4(),
```
AmandaBirmingham, line 2: The general convention in this db is to include the name of the table in the primary key column of a table; this is useful because otherwise we end up with every table having a column named "id" and it becomes hard to keep them straight :). Therefore, I suggest this column be renamed to "melissa_address_query_id" or "address_query_id" or something ...

microsetta_private_api/repo/melissa_repo.py
```diff
@@ -0,0 +1,129 @@
...
+    def check_duplicate(self, address_1, address_2, postal, country):
+        """
+        Check if an address has already been verified to avoid duplicate
+            queries against the Melissa API
+
+        Parameters
+        ----------
+        address_1 - Primary street address
+        address_2 - Secondary street address
+        postal - Postal code
```
AmandaBirmingham, line 54: I see state is not input here; can you add a code comment on why not?

microsetta_private_api/repo/melissa_repo.py
```diff
@@ -0,0 +1,129 @@
...
+                            AND source_address_2 = %s
+                            AND source_postal = %s
+                            AND source_country = %s)
+                            OR (result_address_1 = %s
+                            AND result_address_2 = %s
+                            AND result_postal = %s
+                            AND result_country = %s)""",
+                            (address_1, address_2, postal, country,
+                            address_1, address_2, postal, country))
+            row = cur.fetchone()
```
AmandaBirmingham, line 73: Hmm, now I think I'm confused ... contrary to what the docstring states, this doesn't look like it returns a boolean in the case where a duplicate record does exist.  It is probably undesirable to have the code return a boolean (False) if a duplicate doesn't exist but to return a row if a duplicate does exist;  I suggest changing it to return None if there is no duplicate and row(s) if there is/are duplicates. 
 I say row(s) because I see no clear reason why we can assume there will only be ONE record that matches the select ... can you speak to that?

microsetta_private_api/repo/melissa_repo.py
```diff
@@ -0,0 +1,129 @@
...
+
+        Parameters
+        ----------
+        record_id - Unique ID generated from create_record function
+        source_url - URL we used to query Melissa (for record-keeping/debug)
+        result_raw - Full response from Melissa
+        result_codes - Codes indicating whether address is valid, why, and
+                        what fields were updated
+        result_good - Our internal good/bad flag based on result_codes
+        formatted_address - The fully formatted address returned from API
```
AmandaBirmingham, line 94: Why do we store the formatted_address as well as storing each piece of the formatted address individually?

microsetta_private_api/server_config.json
```diff
@@ -22,5 +22,7 @@
   "daklapack_subscription_key_name": "dak-specific-key",
   "daklapack_subscription_key_val": "123abc456d89z",
   "daklapack_api_base_url": "https://example.com/dak_api",
-  "daklapack_service_email": "dak_fake@example.com"
+  "daklapack_service_email": "dak_fake@example.com",
+  "melissa_url": "http://address.melissadata.net/v3/WEB/GlobalAddress/doglobaladdress",
```
AmandaBirmingham, line 26: Is this url public info?  If not, we would want to replace it in this json with a made up one as for daklapack_api_base_url ...

microsetta_private_api/util/melissa.py
```diff
@@ -0,0 +1,104 @@
...
+                            "admarea": state,
+                            "postal": postal,
+                            "ctry": country}
+
+            url = SERVER_CONFIG["melissa_url"] + "?%s" % \
+                urllib.parse.urlencode(url_params)
+
+            response = requests.get(url)
+            if response.ok is False:
+                raise Exception("Error connecting to address verification API")
```
AmandaBirmingham, line 43: As discussed in Zoom, would be nice to have the error here include more info, such as the response.status_code and possibly its text.  Also would be nice to make the exception type more specific.

microsetta_private_api/util/melissa.py
```diff
@@ -0,0 +1,104 @@
...
+
+            response_raw = response.text
+            response_obj = json.loads(response_raw)
+            if "Records" in response_obj.keys():
+                record_obj = response_obj["Records"][0]
```
AmandaBirmingham, line 48: Without knowing anything about the Melissa api, I feel a bit worried about taking just the first item in a list object with a plural name ("Records").  Can you add a code comment to reassure the user about why we only ever want the first one?

microsetta_private_api/util/melissa.py
```diff
@@ -0,0 +1,104 @@
...
+
+                r_formatted_address = record_obj["FormattedAddress"]
+                r_codes = record_obj["Results"]
+                r_good = False
+
+                codes = r_codes.split(",")
+                for code in codes:
+                    if(code in GOOD_CODES):
+                        r_good = True
```
AmandaBirmingham, line 57: This is probably a short list so not a big deal but: once you get to `r_good = True`, you can `break` out of the codes loop as if one code is good there's no need to examine the remaining codes.

microsetta_private_api/util/melissa.py
```diff
@@ -0,0 +1,104 @@
...
+                r_address_1 = record_obj["AddressLine1"]
+                r_address_2 = record_obj["AddressLine2"]
+                r_city = record_obj["Locality"]
+                r_state = record_obj["AdministrativeArea"]
+                r_postal = record_obj["PostalCode"]
+                r_country = record_obj["CountryName"]
+                r_latitude = record_obj["Latitude"]
+                r_longitude = record_obj["Longitude"]
+    
+                melissa_repo.update_results(record_id, url, response_raw,
```
AmandaBirmingham, line 68: If I'm reading the code right, `update_results` returns a boolean indicating whether a record was really updated, but that boolean isn't being captured/examined here.  I think it should be because if `update_results` returns false, that means that the info we think is being recorded actually isn't and that is a pathological case ... either the record_id is bogus or someone deleted a record from a table that isn't supposed to be deleted from or something is being input in a wrong format or ... whatever the reason, I think we would need to know about it and raise an exception.

AmandaBirmingham (changes requested):
Requests for a few small tweaks. Looks good!

microsetta_private_api/server_config.json
cassidysymons: Yes, the URL is public information so there's no concern about having it visible in the code.

microsetta_private_api/api/microsetta_private_api.yaml
cassidysymons: Good catch, and this has an impact well beyond the scope of the address verification. At the moment, the state field in the database is being used as a sort of catch-all for whatever geographic unit fits between city and postal code - for the US and MX, it's state, while for GB it's storing the county in that field. As we grow into other countries, that's likely to lead to problems. Going to raise this as an issue on Slack to discuss on the AGP API call, as it's an issue that needs to be decided so we can adjust our approach for data collection (signup forms), data verification (this particular code), data storage (database), and third-party usage (e.g. shipping with Daklapak).

microsetta_private_api/api/microsetta_private_api.yaml
cassidysymons: Clarified the intent of the function in the most recent commit.

microsetta_private_api/api/microsetta_private_api.yaml
cassidysymons: I added a 'required' flag to each parameter, which fits the needs of the function based on the countries within our current scope of use.

microsetta_private_api/repo/melissa_repo.py
cassidysymons: Added clarifying information to the parameters.

microsetta_private_api/db/patches/0084.sql
cassidysymons: I renamed the id field to a more unique and descriptive name. I also added indices to the table for the check_duplicates function.

microsetta_private_api/repo/melissa_repo.py
cassidysymons: Added comment in code explaining which fields are necessary and which are superfluous for the sake of duplicate checks.

microsetta_private_api/repo/melissa_repo.py
cassidysymons: I had changed the code after I wrote the docstring. That has been updated to reflect what the function actually returns.

microsetta_private_api/repo/melissa_repo.py
cassidysymons: I added a comment in the code addressing this, but essentially, ease of use for potential future implementations. I'm not opposed to removing it, but my view is that database space on this scale (text strings for thousands of records) is relatively cheap and it allows us more flexibility in how we use the address verification and correction functionality.

microsetta_private_api/util/melissa.py
cassidysymons: I expanded the exception messages and, in the case of the potential for an HTTP failure, added response code and text.

microsetta_private_api/util/melissa.py
cassidysymons: Added comment to code explaining that decision.

microsetta_private_api/util/melissa.py
cassidysymons: Good call, and done.

microsetta_private_api/util/melissa.py
cassidysymons: Good call, I missed that. Added a check for whether it succeeded and the code now raises an exception if it failed to update.

PR #365, opened 2021-08-13
BUG: fallback to default on best_match

wasade:
...avoid a possible `None` for language


dhakim87 (commented):
Definitely approve of setting default on best_match and tests for invalid tags, but

I don't understand why normalize is necessary.  All calls are coming in through microsetta-interface, which sets language tag to whatever we specify.  Where the heck did a tag with the wrong casing come from???

wasade:
Worth having the normalize to be defensive or should it be dropped?

dhakim87 (commented):
I'd rather remove the normalize if we can track down who is sending up "en_us", "en_us" is used as a key sent to babel so being passed up may mean we show the wrong language consent form or something because we forgot to do the lookup.

PR #380, opened 2021-09-17
EBI Accession urls returned in sample model objects

dhakim87:
Samples now report ebi accession url info.


microsetta_private_api/api/_sample.py
```diff
@@ -54,7 +56,35 @@ def read_sample_association(account_id, source_id, sample_id, token_info):
...
+        accession_urls = []
+        for barcode_info in qiita_data:
+            experiment_accession = barcode_info["ebi_experiment_accession"]
+            accession_urls.append(
+                "https://www.ebi.ac.uk/ena/browser/view/" +
+                experiment_accession +
+                "?show=reads")
+
+        sample.set_accession_urls(accession_urls)
+    except Exception as e:
```
wasade, line 84: It would be pleasant to error gracefully for not found (see https://github.com/qiita-spots/qiita_client/blob/0e04e78578c8924f65e12ef6813ebadd2886c5e9/qiita_client/qiita_client.py#L310-L311)

microsetta_private_api/api/_sample.py
```diff
@@ -54,7 +56,35 @@ def read_sample_association(account_id, source_id, sample_id, token_info):
...
+        print("Failed to communicate with qiita", e)
```
wasade, line 85: This message won't make it into any logs

microsetta_private_api/api/_sample.py
```diff
@@ -54,7 +56,35 @@ def read_sample_association(account_id, source_id, sample_id, token_info):
         if sample is None:
             return jsonify(code=404, message="Sample not found"), 404
 
-        return jsonify(sample.to_api()), 200
+    # Check qiita for accession data and fill it in
+    qclient = QiitaClient(
```
wasade, line 60: Construction establishes a session, should we instead have a single instance (e.g, `from microsetta_private_api.something import qclient`) and construct this object only a single time? See https://github.com/qiita-spots/qiita_client/blob/0e04e78578c8924f65e12ef6813ebadd2886c5e9/qiita_client/qiita_client.py#L171

dhakim87: Done

microsetta_private_api/api/_sample.py
dhakim87: Hmm... ideally we would just log a message about qiita being down and not show any accession urls.  Do you want QiitaClient's NotFoundError, BadRequestError, or ForbiddenError's to bubble up to the user?  Is there a way to write BadRequest or Forbidden to the log since neither of these should be happening?

wasade: not found would make sense to boil up as it means the sample sequecne data isn't available in qiita. bad request, forbidden and runtime errors are problems we need to know about but the user does not

I'm also not sure what the best way to this detail into these into the logs would be, would it be possible to track that down? it might be as simple as "from flask import logger" etc

wasade:
Note test failures seem real

wasade:
yay! reviewing

microsetta_private_api/api/_sample.py
```diff
@@ -54,7 +59,44 @@ def read_sample_association(account_id, source_id, sample_id, token_info):
...
+                experiment_accession +
+                "?show=reads")
+        sample.set_accession_urls(accession_urls)
+    except NotFoundError:
+        # I guess qiita doesn't know about this barcode,
+        # so probably no ebi accession info
+        pass
+    except BadRequestError as e:
+        # How do I log these to gunicorn??
+        print("Couldn't communicate with qiita", e)
```
wasade, line 87:
```suggestion
```

microsetta_private_api/api/_sample.py
```diff
@@ -54,7 +59,44 @@ def read_sample_association(account_id, source_id, sample_id, token_info):
...
+        app.logger.warning("Couldn't communicate with qiita", exc_info=True)
+    except ForbiddenError as e:
+        # How do I log these to gunicorn??
+        print("Couldn't communicate with qiita", e)
```
wasade, line 91:
```suggestion

```

microsetta_private_api/api/_sample.py
```diff
@@ -54,7 +59,44 @@ def read_sample_association(account_id, source_id, sample_id, token_info):
...
+        app.logger.warning("Couldn't communicate with qiita", exc_info=True)
+    except RuntimeError as e:
+        # How do I log these to gunicorn??
+        print("Failed to communicate with qiita", e)
```
wasade, line 95:
```suggestion

```

microsetta_private_api/api/tests/test_api.py
```diff
@@ -547,7 +548,13 @@ def client(request):
         with patch("microsetta_private_api.api."
                    "verify_jwt") as mock_v:
             mock_v.side_effect = mock_verify
-            yield client
+            # This is a bit moronic, but since qiita won't be available
```
wasade, line 551: i'll keep this in the interest of getting this merged, but in the future we should consider `unittest.skipIf` so we can provide real credentials when needed

wasade (changes requested):
ive got the changes staged, will commit via gh

wasade (approved):

microsetta_private_api/api/_sample.py
```diff
@@ -54,7 +59,41 @@ def read_sample_association(account_id, source_id, sample_id, token_info):
...
+        # I guess qiita doesn't know about this barcode,
+        # so probably no ebi accession info
+        pass
+    except BadRequestError as e:
+        # How do I log these to gunicorn??
+        app.logger.warning("Couldn't communicate with qiita", exc_info=True)
+    except ForbiddenError as e:
+        # How do I log these to gunicorn??
+        app.logger.warning("Couldn't communicate with qiita", exc_info=True)
+    except RuntimeError as e:
```
wasade, line 91:
```suggestion
    except BadRequestError:
        # How do I log these to gunicorn??
        app.logger.warning("Couldn't communicate with qiita", exc_info=True)
    except ForbiddenError:
        # How do I log these to gunicorn??
        app.logger.warning("Couldn't communicate with qiita", exc_info=True)
    except RuntimeError:
```

PR #388, opened 2021-10-04
Migrate barcode

wasade:
We have a need to migrate barcodes from one source to another on an existing project. 

These modifications update the internal references, and also account for the `source_barcodes_surveys` table

### DIFF
--- a/microsetta_private_api/repo/sample_repo.py
+++ b/microsetta_private_api/repo/sample_repo.py
@@ -118,18 +118,12 @@ class SampleRepo(BaseRepo):
             survey_ids = tuple([r[0] for r in cur.fetchall()])
 
             if len(survey_ids) > 0:
-                # if this was a previously assigned sample...
-                # remove stale relations
+                # if this was a previously assigned sample, then remove any
+                # stale relations
                 cur.execute("""DELETE FROM ag.source_barcodes_surveys
                                WHERE barcode=%s
                                    AND survey_id IN %s""",
                             (barcode, survey_ids))
-
-                # update any vioscreen associations to reflect the source
-                cur.execute("""UPDATE ag.vioscreen_registry
-                               SET source_id=%s
-                               WHERE sample_id=%s""",
-                            (source_id, sample_id))
 
             # collect new survey associations if they exist
             cur.execute("""SELECT survey_id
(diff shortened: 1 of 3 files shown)


microsetta_private_api/repo/sample_repo.py
```diff
@@ -107,12 +107,48 @@ def _update_sample_association(self, sample_id, source_id,
...
-                        "ag_kit_barcode_id = %s",
+            barcode = existing_sample.barcode
+
+            # collect old survey associations if they exist
+            cur.execute("""SELECT survey_id
+                           FROM ag.ag_login_surveys
+                           JOIN ag.ag_kit_barcodes USING (source_id)
+                           WHERE ag_kit_barcode_id=%s""",
+                        (sample_id, ))
+            survey_ids = tuple([r[0] for r in cur.fetchall()])
```
dhakim87, line 118: Why tuple() this list?

wasade: I thought psycopg needed it as `tuple` for `IN` clauses

microsetta_private_api/repo/sample_repo.py
```diff
@@ -107,12 +107,48 @@ def _update_sample_association(self, sample_id, source_id,
...
+
+            if len(survey_ids) > 0:
+                # if this was a previously assigned sample, then remove any
+                # stale relations
+                cur.execute("""DELETE FROM ag.source_barcodes_surveys
```
dhakim87, line 123: Why is this query so different from the one on line 112?  Are we guaranteed that ag_login_surveys JOIN ag_kit_barcodes has the same set of survey ids linked as ag_source_barcodes_surveys?  

If things are out of sync, does this orphan a bunch of rows?

wasade: I'm not sure I follow, the query on like 112 is a selection whereas this one is a deletion? 

`source_barcodes_surveys` as a FK on `survey_id` to `ag_login_surveys`. It would be possible for `ag_login_surveys` JOIN `ag_kit_barcodes` to have a superset of survey IDs; `source_barcodes_surveys` is guaranteed to be a subset. I am not aware of a reason why the sets would differ in practice though

dhakim87 (commented):
I -think- this is fine.  This is a pretty complicated operation though, and I'm unclear if it should require admin intervention.  Checking that would require making it separate from the old operation.  

I'm somewhat worried that we'll have a few things out of sync between the source_barcodes_surveys path and the ag_login_surveys path which would make surveys disappear when barcodes are moved from source to source, but honestly those would already have to be in a bad state so I'm not sure what else to do.  

One thing to check before this is merged:  the original _update_sample_association didn't appear to set survey information at all, but now it looks like it will.  If you create a new source, and add a new barcode to it, I believe this function will get called.  Can we ensure that that path doesn't do anything stupid like associate all the surveys to the sample twice?

dhakim87 (approved):

wasade:
Thanks!

PR #396, opened 2021-10-20
Use pkg_resources for datapath

wasade:


wasade:
cc @AmandaBirmingham I think this should resolve the issue, but let's wait for the tests

PR #402, opened 2021-11-01
TST/API: campaign is_member support

wasade:
Adds support for testing if a user is a member of a campaign. 

~~*DO NOT* merge this yet, there is a possible test failure which may not be captured by CI that I'm verifying right now but could also be due to stale local state.~~


wasade:
@cassidysymons, same code as before, are you 👍 on this?

cassidysymons:
@wasade looks good

wasade:
thanks!

PR #403, opened 2021-11-03
Fundrazr transaction support

wasade:
Support for:

* CRUD operations for FundRazr transactions
* A minimal API client for querying FundRazr for campaign and payment information
* A celery beat task for obtaining new transactions
* Support for load of historical transaction data
* Use of central `interested_users` and `campaign` information
* Support for anonymous transactions
* Responsive to new campaigns or perks w/o manual modification
* Transactions are reasonably general so tentatively we can support other mechanisms

What is not done yet (and would prefer to resolve in a separate pull request):
* Add a celery beat task to perform address verification, emailing a user if not present
* An API endpoint to allow a user to correct their address 
* API endpoints to support transaction querying for administrative interfaces


microsetta_private_api/db/patches/0090.sql
```diff
@@ -0,0 +1,132 @@
...
+    remote_campaign_id VARCHAR PRIMARY KEY,
+    internal_campaign_id uuid NOT NULL,
+    currency VARCHAR NOT NULL,
+    UNIQUE (remote_campaign_id, internal_campaign_id),
+    CONSTRAINT fk_campaign FOREIGN KEY (internal_campaign_id) REFERENCES barcodes.campaigns (campaign_id)
+);
+
+INSERT INTO barcodes.campaigns 
+    (title) 
+    VALUES ('The Microsetta Initiative');
```
wasade, line 20: TODO: denote other known fundrazr campaigns

microsetta_private_api/db/patches/0090.sql
```diff
@@ -0,0 +1,132 @@
...
+INSERT INTO barcodes.fundrazr 
+    SELECT '31l0S2' AS remote_campaign_id, campaign_id AS internal_campaign_id, 'usd' AS currency
+    FROM barcodes.campaigns
+    WHERE title='The Microsetta Initiative';
+
+CREATE TYPE tmi_transaction_status AS ENUM ('received', 'valid-address', 'invalid-address', 'shipment-requested');
```
wasade, line 26: QUESTION: are these statuses sufficient?

microsetta_private_api/db/patches/0090.sql
```diff
@@ -0,0 +1,132 @@
...
+
+CREATE TABLE barcodes.fundrazr_transaction (
+    id VARCHAR PRIMARY KEY, 
+    remote_campaign_id VARCHAR,
+    created TIMESTAMP NOT NULL,
+    amount FLOAT NOT NULL,
+    net_amount FLOAT NOT NULL,
+    currency VARCHAR NOT NULL,
+    fundrazr_status VARCHAR NOT NULL,
```
wasade, line 35: COMMENT: Per their docs, this can take multiple values. From follow up, we *only* care about status == "completed". 

TODO: revise client to only pull completed, remove this field, update model, etc

microsetta_private_api/db/patches/0090.sql
```diff
@@ -0,0 +1,132 @@
...
+);
+
+CREATE TABLE barcodes.fundrazr_perk_to_daklapack_article (
+    perk_id VARCHAR NOT NULL,
+    dak_article_code INTEGER NOT NULL,
+    CONSTRAINT pk_perk_to_dak PRIMARY KEY (perk_id, dak_article_code),
+    CONSTRAINT fk_perk_to_dak FOREIGN KEY (dak_article_code) REFERENCES barcodes.daklapack_article (dak_article_code)
+);
+
+-- This is not a complete list, but what is available right now. A
```
wasade, line 91: TODO: make this complete

microsetta_private_api/db/patches/0090.sql
```diff
@@ -0,0 +1,132 @@
...
+    dak_article_code INTEGER NOT NULL,
+    CONSTRAINT pk_perk_to_dak PRIMARY KEY (perk_id, dak_article_code),
+    CONSTRAINT fk_perk_to_dak FOREIGN KEY (dak_article_code) REFERENCES barcodes.daklapack_article (dak_article_code)
+);
+
+-- This is not a complete list, but what is available right now. A 
+-- minimal set are added, relative to the Microsetta campaign
+-- (31l0S2) so we can exercise links against daklapack
+
+-- >>> res = requests.get('https://api.fundrazr.com/v1/campaigns?organization=873K6', headers=d)
```
wasade, line 95: COMMENT: the organization ID is okay to share publicly per Fundrazr team

microsetta_private_api/model/fundrazr.py
```diff
@@ -0,0 +1,230 @@
...
+        self.contact_email = contact_email
+        self.shipping_address = shipping_address
+        self.account = account
+        self.subscribe_to_updates = subscribe_to_updates
+        self.claimed_items = claimed_items
+        self.phone_number = phone_number
+        self.message = message
+        self.tmi_status = tmi_status
+
+    def copy(self):
```
wasade, line 127: TODO: add a unit test as this copy is complex

microsetta_private_api/model/fundrazr.py
```diff
@@ -0,0 +1,230 @@
...
+                                         for item in structured[CLAIMED_ITEMS]]
+
+        if SHIPPING_ADDRESS in structured:
+            shipping = Shipping.from_api(**structured[SHIPPING_ADDRESS])
+            structured[SHIPPING_ADDRESS] = shipping
+
+        return cls(**structured)
+
+    @classmethod
+    def from_db(cls, trn, items):
```
wasade, line 193: TODO: add unit tests as this method is complex. This is also rather awkward, and possibly should be expressed instead as a private member method of the `FundRazr` repo object

microsetta_private_api/model/tests/test_fundrazr_transaction.py
```diff
@@ -0,0 +1,185 @@
...
+        self.assertEqual(obs.shipping_address.first_name, 'person')
+        self.assertEqual(obs.claimed_items, None)
+
+    def test_order_with_multiple_items(self):
+        obs = Payment.from_api(**MULTIPLE_ITEMS)
+        self.assertEqual(len(obs.claimed_items), 3)
+        self.assertEqual(obs.claimed_items[0].title,
+                         'USPS Priority Mail Shipping')
+        self.assertEqual(obs.claimed_items[0].quantity, 5)
+
```
wasade, line 182: TODO: add `test_copy`

microsetta_private_api/repo/fundrazr.py
```diff
@@ -0,0 +1,325 @@
...
+              'payer_last_name', 'payer_email', 'account_type',
+              'contact_email', 'subscribed_to_updates', 'phone_number',
+              'message', 'tmi_status')
+
+    _SHIPPING = ('shipping_first_name', 'shipping_last_name',
+                 'shipping_address1', 'shipping_address2',
+                 'shipping_city', 'shipping_country', 'shipping_state',
+                 'shipping_postal')
+
+    # enum statuses. status transitions as follows:
```
wasade, line 31: COMMENT: this should have parity with the database enum

microsetta_private_api/repo/fundrazr.py
```diff
@@ -0,0 +1,325 @@
...
+                inserts = [(payment.transaction_id, i.id, i.quantity)
+                           for i in items]
+                try:
+                    cur.executemany("""INSERT INTO
+                                           barcodes.fundrazr_transaction_perk
+                                       (transaction_id, perk_id, quantity)
+                                       VALUES (%s, %s, %s)""",
+                                    inserts)
+                except psycopg2.errors.ForeignKeyViolation:
+                    # this would indicate a synchronization issue, where
```
wasade, line 116: TODO: add an idempotent method that can be called *upstream* to check / add perks and note in docstring this should be called prior to this method

microsetta_private_api/repo/fundrazr.py
```diff
@@ -0,0 +1,325 @@
...
+                    detail = [i.to_api() for i in items]
+                    raise UnknownItem(f"One or more perks are unknown:\n"
+                                      f"{json.dumps(detail, indent=2)}")
+
+        return True
+
+    def update_transaction(self, payment):
+        """Update the specific detail about a transaction
+
+        We only allow update of shipping information for a transaction, which
```
wasade, line 127: COMMENT: I believe this is the only information we would ever want to update in a transaction but please let me know if I'm wrong about that

microsetta_private_api/repo/fundrazr.py
```diff
@@ -0,0 +1,325 @@
...
+                payment.status,
+                payment.payer_first_name,
+                payment.payer_last_name,
+                payment.payer_email,
+                payment.account,
+                payment.contact_email,
+                payment.subscribe_to_updates,
+                payment.phone_number,
+                payment.message,
+                self.TMI_STATUS_RECEIVED)
```
wasade, line 87: QUESTION: we force an added transaction to be received, which I believe is correct as we go forward. However, representation of historical transaction data will be hard. I think we should remark all historical as "shipment-requested" and issue shipment orders out-of-band for those who we know are waiting on a kit, I worry doing otherwise will be insane. Thoughts?

microsetta_private_api/repo/fundrazr.py
```diff
@@ -0,0 +1,325 @@
...
+
+        sql = ("""UPDATE barcodes.fundrazr_transaction
+                  SET tmi_status=%s
+                  WHERE id=%s""",
+               (status, payment.transaction_id))
+
+        with self._transaction.cursor() as cur:
+            cur.execute(*sql)
+
+        payment = payment.copy()
```
wasade, line 238: COMMENT: it wouldn't be *terrible* to just inplace update the status...

microsetta_private_api/repo/fundrazr.py
```diff
@@ -0,0 +1,325 @@
...
+        payment.tmi_status = status
+        return payment
+
+    def get_transactions(self, before=None, after=None, transaction_id=None,
+                         tmi_status=None, contact_email=None):
```
wasade, line 243: QUESTION: is there value in other criteria here?

microsetta_private_api/repo/fundrazr.py
```diff
@@ -0,0 +1,325 @@
...
+            cur.execute(*sql)
+            res = cur.fetchall()
+
+        return [self._payment_from_transaction(id_) for id_ in res]
+
+    def _payment_from_transaction(self, id_):
+        """Construct a payment instance from a transaction ID"""
+        fields = self._ORDER + self._SHIPPING
+        fields_formatted = ','.join(fields)
+        transaction_sql = (f"""SELECT {fields_formatted}
```
wasade, line 305: TODO: express this as a single query with ARRAY_AGG

microsetta_private_api/util/fundrazr.py
```diff
@@ -0,0 +1,15 @@
+from microsetta_private_api.celery_utils import celery
+from microsetta_private_api.repo.transaction import Transaction
+from microsetta_private_api.repo.fundrazr import FundRazr as FundRazrRepo
+from microsetta_private_api.client.fundrazr import FundRazrClient
+
+
+@celery.task(ignore_result=True)
+def get_fundrazr_transactions():
```
wasade, line 8: COMMENT: this is pseudocode at this point

microsetta_private_api/util/fundrazr.py
```diff
@@ -0,0 +1,15 @@
...
+    c = FundRazrClient()
+    with Transaction() as t:
+        fr = FundRazrRepo(t)
+        last_transaction = fr.last_transaction()
```
wasade, line 12: TODO: add this method

microsetta_private_api/db/patches/0090.sql
cassidysymons: Clarifying question - does "received" mean the transaction was received by TMI, or the shipment was received by the user? If it's the latter, it might be worth updating to "shipment-received" for clarity.

microsetta_private_api/repo/fundrazr.py
cassidysymons: On the first questions, I could see value in allowing for other transaction types, provided it's not a ton of extra work. If other transaction types do come into play, there will be work integrating them into the system, but building that flexibility in now seems wise.

On the second question, I think centralized status information would go a long way in streamlining reporting. It's more beneficial to sponsored projects than Fundrazr specifically, but being able to view the process as a funnel - rom expression of interest to getting a kit to sending it back - and having an immediate & clear picture of where someone is in that process is helpful. All of the info can be gleaned from individual components, but centralizing it simplifies the process.

microsetta_private_api/repo/fundrazr.py
cassidysymons: Agree with that method. Any idea of the scope of pending kits right now? It would be worth doing a full audit as we implement this to minimize outstanding liabilities - especially ones that might be untracked - going forward.

microsetta_private_api/repo/fundrazr.py
cassidysymons: I can see potential use cases for querying transactions by campaign.

microsetta_private_api/db/patches/0090.sql
wasade: In this context, the intent was "transaction received"

microsetta_private_api/repo/fundrazr.py
wasade: (1) 👍 , will build a proposed approach here. (2) also 👍. I'll make some modifications as a proposed central solution -- seems well worth it.

microsetta_private_api/repo/fundrazr.py
wasade: Pending is on the scale of 100-200, as even though we shut down, some people still wanted to donate. Kat did audit transactions for > March 2020 but we'll need to make sure the audit is up to date once we ingress

microsetta_private_api/repo/fundrazr.py
wasade: 👍

microsetta_private_api/repo/fundrazr.py
wasade: Deferring on (2), it will be a complex PR in its own right to fully centralize this information. We can reduce the content of the fundrazr transaction table though as much of the content overlaps with interested_users, and we can rely on the status flags there (at least partially if not entirely for htis).

microsetta_private_api/client/tests/test_fundrazr.py
```diff
@@ -0,0 +1,41 @@
+from unittest import TestCase, main, skipIf
+
+from microsetta_private_api.client.fundrazr import FundrazrClient
+from microsetta_private_api.config_manager import SERVER_CONFIG
+
+
+class FundrazrClientTests(TestCase):
```
wasade, line 7: Finalization of tests is blocked by Fundrazr at the moment while we work out how best to interact with their staging system

wasade (commented):
Good for final review

microsetta_private_api/db/patches/0090.sql
```diff
@@ -0,0 +1,103 @@
...
+    payer_first_name VARCHAR NOT NULL,
+    payer_last_name VARCHAR NOT NULL,
+    payer_email VARCHAR,
+    account_type VARCHAR NOT NULL,
+    message VARCHAR,
+    subscribed_to_updates BOOLEAN NOT NULL,
+    CONSTRAINT fk_transaction_campaign FOREIGN KEY (remote_campaign_id) REFERENCES campaign.transaction_source_to_campaign (remote_campaign_id),
+    CONSTRAINT fk_transaction_user FOREIGN KEY (interested_user_id) REFERENCES campaign.interested_users (interested_user_id)
+);
+
```
cassidysymons, line 57: Perhaps I'm missing it elsewhere, but is there a field reflecting whether a transaction has been fulfilled? It will be valuable for historical transactions so we can mark them all fulfilled on ingress, but it would also be useful for managing process flow as new transactions enter the system.

microsetta_private_api/db/patches/0090.sql
```diff
@@ -0,0 +1,103 @@
+CREATE SCHEMA campaign;
+ALTER TABLE barcodes.interested_users SET SCHEMA campaign;
+ALTER TABLE barcodes.campaigns SET SCHEMA campaign;
+ALTER TABLE barcodes.campaigns_projects SET SCHEMA campaign;
+ALTER TABLE ag.melissa_address_queries SET SCHEMA campaign;
```
cassidysymons, line 5: The Melissa code needs to be updated to reflect the schema change. Nothing is actually using that code in production yet, so if you don't want to further bloat this PR, perhaps the quickest solution would be to remove this one line? I can move ag.melissa_address_queries into the campaign schema and update the code with my next PR that will cover address verification for Fundrazr transactions.

microsetta_private_api/db/patches/0090.sql
wasade: We have no way to determine this for historical data. It would be good for future data though -- good call

microsetta_private_api/db/patches/0090.sql
wasade: Sounds good, can do

microsetta_private_api/db/patches/0090.sql
wasade: A transaction can have partial fulfillment as well, and some transactions the notion of fulfillment doesn't make sense (e.g., straight donation). Should this be out of scope?

cassidysymons: Yeah, since this isn't intended to be fully end-to-end, it probably makes the most sense to punt it and discuss how to handle it with the next phase of development.

PR #404, opened 2021-11-23
Private API functionality for interested users.

cassidysymons:


microsetta_private_api/repo/interested_user_repo.py
```diff
@@ -0,0 +1,108 @@
...
+                "SELECT address_1, address_2, city, state, postal_code, "
+                "country "
+                "FROM barcodes.interested_users WHERE interested_user_id = %s",
+                (interested_user_id,)
+            )
+            r = cur.fetchone()
+            if r is None:
+                return None
+            else:
+                melissa_response = verify_address(r['address_1'],
```
wasade, line 73: This issues a remote resource call, and should be wrapped in a try/except. It would be beneficial to do this outside of the transaction context to avoid holding the transaction open unnecessarily

microsetta_private_api/repo/interested_user_repo.py
```diff
@@ -0,0 +1,108 @@
+from microsetta_private_api.repo.base_repo import BaseRepo
+from microsetta_private_api.exceptions import RepoException
+from microsetta_private_api.util.melissa import verify_address
+
+
+class InterestedUserRepo(BaseRepo):
+    def __init__(self, transaction):
+        super().__init__(transaction)
+
+    def insert_interested_user(self, **kwargs):
```
wasade, line 10: why not pass in an instance of `InterestedUser`?

wasade (changes requested):
Thanks @cassidysymons

Can unit tests be added to exercise / verify behavior on the repo object? It would be good to include a test for attempting to verify an interested user that has already been verified and assert behavior is as expected. 

Integration tests against the exposed routes would be valuable as well

microsetta_private_api/repo/interested_user_repo.py
cassidysymons: I'm adding a try/except block but I'm unsure how to entirely move it out of the transaction context since it's a database operation. There are other remote resource calls that leave a transaction open in a similar fashion (e.g. Daklapack) and it seems inherently necessary. Can you expand on what you were thinking?

wasade: Oh I see, you're right -- I didn't expect the db operation outside of `MelissaRepo`. Makes sense to leave as is

microsetta_private_api/repo/interested_user_repo.py
cassidysymons: It seemed more flexible to use the kwargs approach since interested users could end up coming from a variety of different avenues (submission form, Fundrazr, others TBD). I'm open to changing it to an instance of the InterestedUser model if you feel that's better, though.

cassidysymons:
In regards to a unit test to validate the behavior when attempting to verify an already-verified address - Melissa has no concept of a testing mode, all requests must utilize real data and a valid license key. The ramifications of this are two-fold:
1) Unit testing for the Melissa repo/functionality mostly needs to be mocked up, rather than being able to test the remote resource.
2) Anything that interacts with Melissa - for example, interested_users - can then interact with the mock process.

If we want to proceed that way, I can set up the mock testing for Melissa, then link interested user testing in to that.

wasade:
Sorry for the delay, just saw the comment. Why not `patch` the `verify_address` method? https://github.com/biocore/microsetta-private-api/pull/404/files#diff-8f4fe0085812e80c6dca2c22037d070c32252604607e2117afe53d9a25cbdb45R77 We might be saying the same thing though

microsetta_private_api/repo/interested_user_repo.py
wasade: Thanks! For other repos with complex arguments, we have been utilizing model objects (eg vioscreen). A benefit here is future modification to the model doesn't change the public method signature

microsetta_private_api/repo/tests/test_interested_user_repo.py
```diff
@@ -0,0 +1,114 @@
...
+        interested_user = InterestedUser.from_dict(dummy_user)
+        with Transaction() as t:
+            interested_user_repo = InterestedUserRepo(t)
+            user_id = \
+                interested_user_repo.insert_interested_user(interested_user)
+            obs = interested_user_repo.verify_address(user_id)
+            self.assertTrue(obs is None)
+            t.rollback()
+
+
```
wasade, line 112: Could a test for `test_verify_address_not_previously_verified_isvalid` and `test_verify_address_not_previously_verified_isinvalid` which mocks out `microsetta_private_api.util.melissa.verify_address` using a `unittest.patch` be added so that we can cover the remaining logic of the `InterestedUserRepo.verify_address` method?

microsetta_private_api/repo/tests/test_interested_user_repo.py
```diff
@@ -0,0 +1,114 @@
...
+            "address_checked": True
+        }
+        interested_user = InterestedUser.from_dict(dummy_user)
+        with Transaction() as t:
+            interested_user_repo = InterestedUserRepo(t)
+            user_id = \
+                interested_user_repo.insert_interested_user(interested_user)
+            obs = interested_user_repo.verify_address(user_id)
+            self.assertTrue(obs is None)
+            t.rollback()
```
wasade, line 110:
```suggestion
```

A rollback is implicit; commit needs to be explicit

wasade (changes requested):

microsetta_private_api/model/interested_user.py
```diff
@@ -0,0 +1,42 @@
...
+        self.converted_to_account = kwargs.get('converted_to_account', False)
+        self.converted_to_account_timestamp = \
+            kwargs.get('converted_to_account_timestamp')
+        self.over_18 = kwargs.get('over_18', False)
+
+    def to_api(self):
+        return self.__dict__.copy()
+
+    @classmethod
+    def from_dict(cls, values_dict):
```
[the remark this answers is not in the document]
cassidysymons: Different models in the existing code base handle it differently - e.g. project.py is similar to what I wrote while others enumerate the parameters in the constructor similar to your example. Would you prefer to use the latter going forward?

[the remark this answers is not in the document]
wasade: fair, good point, precedence ftw, let's leave as is

microsetta_private_api/repo/tests/test_interested_user_repo.py
wasade: It looks like there are mocks on `microsetta_private_api.repo.interested_user_repo.verify_address` rather than `microsetta_private_api.util.melissa.verify_address`. Doing the mocking at the lower level will ensure the database operations within `microsetta_private_api.repo.interested_user_repo.verify_address` are exercised

cassidysymons: I just dove into this and as far as I can tell, the way I have it patched is exercising the database operations in `microsetta_private_api.repo.interested_user_repo.verify_address`. I added a `t.commit()` to `test_verify_address_not_verified_is_valid` and the latitude, longitude, address_checked, and address_valid fields all update as intended in the `interested_users` table.

When I change the patch to `microsetta_private_api.util.melissa.verify_address`, it attempts to connect to the Melissa API (and fails, as expected, since the license key is removed). Am I misunderstanding what you were requesting or is the mocking working as intended?

wasade: I think I see that now, thanks

wasade:
I'm :+1: following a `mv` on `0090.sql` to a new patch number

wasade:
The failure is unrelated, and weird. I'm merging and will investigate the CI issue after

PR #408, opened 2021-12-11
Summary by project

charles-cowart:
A little bit of new code was needed after all to support requests for summaries by project id.
This code is still a work in progress. It needs flake8 and unittests.
Should be ready before Monday.


wasade (commented):
This seems good, thanks! I'll hold off merge as its remarked WIP

PR #412, opened 2022-01-13
Allow for replacing verify_jwt

wasade:
Provide a means to substitute `verify_jwt` as to facilitate integration testing between -interface and -private-api


dhakim87 (approved):
verify_jwt does more than I thought.  Seems good the way this is, was it possible to make a _verify_jwt(token, verify_signature) that just passes down true or false to jwt.decode?  Seems to be a lot of overlap otherwise, though I imagine it makes creation of fake tokens slightly harder.

wasade:
Thanks!! 

I agree there is replication. But, for the point you noted, it felt like "the right thing to do" as this is a point of risk

PR #423, opened 2022-03-08
Delete account

wasade:
Implements an admin endpoint to delete an account if safe, scrubbing otherwise, such that non-identifiable details are retained

### DIFF
--- a/microsetta_private_api/admin/admin_impl.py
+++ b/microsetta_private_api/admin/admin_impl.py
@@ -775,3 +775,50 @@ def qiita_barcode_query(body, token_info):
     )
 
     return jsonify(qiita_data), 200
+
+
+def delete_account(account_id, token_info):
+    validate_admin_access(token_info)
+
+    with Transaction() as t:
+        acct_repo = AccountRepo(t)
+        src_repo = SourceRepo(t)
+        samp_repo = SampleRepo(t)
+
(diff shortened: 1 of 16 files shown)


wasade:
I believe this is now good for review

microsetta_private_api/admin/admin_impl.py
```diff
@@ -775,3 +776,60 @@ def qiita_barcode_query(body, token_info):
...
+            for sample in samples:
+                # we scrub rather than disassociate in the event that the
+                # sample is in our freezers but not with an up-to-date scan
+                samp_repo.scrub(account_id, source.id, sample.id)
+
+            surveys = sar_repo.list_answered_surveys(account_id, source.id)
+            if has_samples:
+                # if we have samples, we need to scrub survey / source
+                # free text
+                print(surveys)
```
wasade, line 816:
```suggestion
                print(surveys)
```
```suggestion
```

wasade, line 816:
```suggestion
                print(surveys)
```
```suggestion
```

wasade:
@cassidysymons suggested checking with how authrocket will be handled

@cassidysymons noted connection to interested users, and cleaning up that table -- an issue will be created as that is out of scope here.

TODO: add remaining docstrings

PR #427, opened 2022-03-29
Celery task to geocode legacy accounts

cassidysymons:
Changes:
1) New Celery task to geocode legacy accounts using Melissa's Global Address API
2) Minor tweaks to handle edge cases in Melissa implementation

Questions/Points of Discussion:
1) How cautiously do we want to throttle the checks per minute? Under normal conditions, each check takes a fraction of a second, but there's no real urgency to blow through the whole database as fast as possible.
2) What's the most effective way to log Exceptions that the interaction with Melissa might throw?


wasade:
@cassidysymons this is good but any chance a `git mv 0097.sql 0098.sql` could be performed?

cassidysymons:
@wasade done

PR #433, opened 2022-04-08
update description on all

wasade:
@AmandaBirmingham, when deployed to staging, the final `ALTER TABLE barcodes.daklapack_article ALTER COLUMN detailed_description SET NOT NULL;` was failing. It seems there may have been state present not reflected in the test database + patches? 

I think the proposed change here addresses it, when you have a chance could you take a look?


AmandaBirmingham (approved):
Looks good to me

PR #445, opened 2022-08-01
Fixed Duplicate Source Name and Email Check

Ram6911:


microsetta_private_api/repo/source_repo.py
```diff
@@ -258,3 +258,19 @@ def scrub(self, account_id, source_id):
...
+
+    def get_duplicate_source_name_email(self, account_id, source_name, email):
+
+        with self._transaction.dict_cursor() as cur:
+            cur.execute("SELECT " + SourceRepo.read_cols +
+                        " FROM "
+                        "source "
+                        "WHERE "
+                        "source.account_id = %s AND "
+                        "source.participant_email = %s AND "
```
Ram6911, line 270: Instead of AND we should use OR

microsetta_private_api/api/_source.py
```diff
@@ -155,3 +155,16 @@ def create_human_source_from_consent(account_id, body, token_info):
...
+
+def check_duplicate_source_name_email(account_id, body):
+    with Transaction() as t:
+        source_repo = SourceRepo(t)
+        source_name = body['participant_name']
+        email = body['participant_email']
+        source = source_repo.get_duplicate_source_name_email(
+            account_id, source_name, email)
+        if source['source_duplicate']:
+            source["message"] = DUPLICATE_SOURCE_AND_EMAIL
```
Ram6911, line 168: Change DUPLICATE_SOURCE_AND_EMAIL to TRUE

microsetta_private_api/repo/source_repo.py
Ram6911: Done

microsetta_private_api/api/_source.py
Ram6911: Done

microsetta_private_api/api/_source.py
```diff
@@ -155,3 +155,16 @@ def create_human_source_from_consent(account_id, body, token_info):
...
+
+def check_duplicate_source_name_email(account_id, body):
+    with Transaction() as t:
+        source_repo = SourceRepo(t)
+        source_name = body['participant_name']
+        email = body['participant_email']
+        source = source_repo.get_duplicate_source_name_email(
+            account_id, source_name, email)
+        if source['source_duplicate']:
+            source["message"] = TRUE
```
cassidysymons, line 168: Please change the return value to an actual boolean (true/false). The goal is to return a boolean, rather than a string, so that if we need to change the message that's displayed, we can do it at the Interface level rather than the API level.

microsetta_private_api/api/literals.py
```diff
@@ -14,3 +14,10 @@
 ACCT_NOT_FOUND_MSG = "Account not found"
 SRC_NOT_FOUND_MSG = "Source not found"
 INVALID_TOKEN_MSG = "Invalid token"
+TRUE = "It looks like you're creating a new source " \
```
cassidysymons, line 17: This literal can be removed.

microsetta_private_api/api/microsetta_private_api.yaml
```diff
@@ -168,6 +168,45 @@ paths:
         '422':
           $ref: '#/components/responses/422UnprocessableEntity'
 
+  '/accounts/{account_id}/source_email_match':
```
cassidysymons, line 171: Please change this to '/accounts/{account_id}/check_duplicate_source' - we want the name to be as clear and descriptive as possible.

microsetta_private_api/repo/source_repo.py
```diff
@@ -258,3 +258,19 @@ def scrub(self, account_id, source_id):
...
+    def get_duplicate_source_name_email(self, account_id, source_name, email):
+
+        with self._transaction.dict_cursor() as cur:
+            cur.execute("SELECT " + SourceRepo.read_cols +
+                        " FROM "
+                        "source "
+                        "WHERE "
+                        "source.account_id = %s AND "
+                        "source.participant_email = %s OR "
+                        "source.source_name = %s",
```
cassidysymons, line 271: This query isn't achieving what we're looking for. When you're using both AND and OR statements in a conditional, you need to use parentheses to meet the condition. We're looking for any source that matches account_id AND (participant email OR source_name).

microsetta_private_api/api/microsetta_private_api.yaml
```diff
@@ -168,6 +168,45 @@ paths:
...
+                type: object
+                properties:
+                  source_match:
+                    type: boolean
+        '401':
+          $ref: '#/components/responses/401Unauthorized'
+        '403':
+          $ref: '#/components/responses/403Forbidden'
+        '404':
+          $ref: '#/components/responses/404NotFound'
```
cassidysymons, line 208: Can this function actually return all of these HTTP codes?

cassidysymons (changes requested):

microsetta_private_api/api/_source.py
```diff
@@ -155,3 +154,13 @@ def create_human_source_from_consent(account_id, body, token_info):
     # NB: Don't expect to handle errors 404, 422 in this function; expect to
     # farm out to `create_source`
     return create_source(account_id, source, token_info)
+
+
+def check_duplicate_source_name_email(account_id, body):
```
cassidysymons, line 159: Can you add _validate_account_access() to this function please?

cassidysymons (changes requested):

cassidysymons (approved):

PR #449, opened 2022-08-15
Translate email subjects

SyedAli-789:
expose all of the email subject lines for user-facing emails to the translation mechanisms

### DIFF
--- a/microsetta_private_api/admin/admin_impl.py
+++ b/microsetta_private_api/admin/admin_impl.py
@@ -9,7 +9,6 @@ from microsetta_private_api.model.project import Project
 from microsetta_private_api.model.daklapack_order import DaklapackOrder, \
     ORDER_ID_KEY, SUBMITTER_ACCT_KEY, ADDR_DICT_KEY
 from microsetta_private_api.exceptions import RepoException
-from microsetta_private_api.repo import account_repo
 from microsetta_private_api.repo.account_repo import AccountRepo
 from microsetta_private_api.repo.activation_repo import ActivationRepo
 from microsetta_private_api.repo.event_log_repo import EventLogRepo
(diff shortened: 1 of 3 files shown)


microsetta_private_api/admin/admin_impl.py
```diff
@@ -281,6 +282,10 @@ def send_email(body, token_info):
         contact_name = None
         activation_code = None
         language = localization.EN_US
+        acct_repo = account_repo.AccountRepo(t)
+        account = acct_repo.get_account(account_id)
```
cassidysymons, line 286: account_id is None, so this will never actually work - it needs to be further down in the code and how it's invoked depends on the value of body["issue_type"]

microsetta_private_api/admin/email_templates.py
```diff
@@ -1,4 +1,5 @@
 from enum import Enum
+from gettext import gettext
```
cassidysymons, line 2:
```suggestion
from flask_babel import gettext
```

microsetta_private_api/util/email.py
```diff
@@ -83,7 +84,7 @@ def send(cls, to, email_template, email_template_args=None, from_=None,
         message['To'] = to
         message['From'] = from_ or cls.from_
         message['Reply-To'] = cls.reply_to
-        message['Subject'] = email_template.subject
+        message['Subject'] = gettext(email_template.subject)
```
cassidysymons, line 87: The subject lines are already wrapped in gettext() in email_templates.py - wouldn't this code mean they're being subjected to two layers of gettext()?

cassidysymons (changes requested):

cassidysymons (approved):

PR #452, opened 2022-08-30
remove kit id and code from activation form

raginirai553:


raginirai553:
Lint error fixed

microsetta_private_api/model/account.py
```diff
@@ -30,7 +30,7 @@ def from_dict(input_dict, auth_iss, auth_sub):
                 input_dict['address']['post_code'],
                 input_dict['address']['country_code']
             ),
-            input_dict['kit_name'],
+            "",
```
cassidysymons, line 33: Rather than passing in a blank string here, can you update the class to remove the property completely? The relevant unit and integration tests will need to be updated as well.

microsetta_private_api/api/tests/test_api.py
```diff
@@ -683,6 +697,8 @@ def validate_dummy_acct_response_body(self, response_obj,
         expected_dict[ACCT_TYPE_KEY] = ACCT_TYPE_VAL
         expected_dict[CREATION_TIME_KEY] = real_creation_time
         expected_dict[UPDATE_TIME_KEY] = real_update_time
+        print(expected_dict)
+        print(response_obj)
```
cassidysymons, line 701: Please make sure to remove debug output before submitting pull requests.

cassidysymons (changes requested):

microsetta_private_api/model/account.py
raginirai553: Removing the property completely means many downstream changes like in DB and most importantly in source creation, sample allocation etc. Created_with_kit_id on account DB is referenced and used further in listing and looking up samples. We probably will need to rewrite codes for source and sample creation if we wish to completely remove it. If you wish I can continue to work to achieve that

microsetta_private_api/api/tests/test_integration.py
```diff
@@ -664,7 +664,7 @@ def test_create_new_account(self):
                 "email": FAKE_EMAIL,
                 "first_name": "Jane",
                 "last_name": "Doe",
-                "kit_name": "jb_qhxqe",
+                # "kit_name": "jb_qhxqe",
```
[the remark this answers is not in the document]
raginirai553: Will remove it soon

microsetta_private_api/api/tests/test_api.py
raginirai553: sure, will do that

raginirai553: done

microsetta_private_api/model/account.py
cassidysymons: Please look into it and write up specifications on what all would need to be changed downstream and how long you think it would take to implement those changes.

microsetta_private_api/api/tests/test_integration.py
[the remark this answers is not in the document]
raginirai553: done

microsetta_private_api/db/patches/0063.sql
```diff
@@ -38,4 +38,3 @@ UPDATE ag.account SET created_with_kit_id=subquery.kit_id
 UPDATE ag.account SET created_with_kit_id=''
     WHERE created_with_kit_id IS NULL;
 
-ALTER TABLE ag.account ALTER COLUMN created_with_kit_id SET NOT NULL;
```
cassidysymons, line 41: We can't change old database patches, we need to create a new database patch that makes the desired change to the existing structure.

microsetta_private_api/model/account.py
```diff
@@ -30,16 +30,15 @@ def from_dict(input_dict, auth_iss, auth_sub):
...
         )
         return result
 
     def __init__(self, account_id, email,
                  account_type, auth_issuer, auth_sub,
                  first_name, last_name,
-                 address, created_with_kit_id,
-                 language,
+                 address, language,
+                 created_with_kit_id=None,
```
cassidysymons, line 41: Is there any reason why you're leaving the created_with_kit_id in the Account model?

cassidysymons (changes requested):

microsetta_private_api/model/account.py
raginirai553: the existing database has that column in the account table. Intent is to not make any changes to avoid downtime and unknown dependency issues.It is to ensure accounts created earlier continue to function as is. Removing column in table might result in downtime and functionalities of old accounts.Let me know if any concerns.

microsetta_private_api/db/patches/0063.sql
raginirai553: Writing a new patch for this setting.

microsetta_private_api/model/account.py
raginirai553: like there is an admin action which retrieves account details based on created_with_kit_id

PR #460, opened 2022-09-22
Metadata Survey Updates

cassidysymons:
Updating the Metadata repo to reflect recent survey changes:
1) Adding MyFoodRepo, the Polyphenol FFQ, and the Spain FFQ to the ignored survey templates, as they're external surveys and can't be retrieved from the database.
2) Adding all of the fields in the locally hosted Cooking Oils & Oxalate-rich Foods survey to the EBI_REMOVE constant, as they don't exist in Qiita.

### DIFF
--- a/microsetta_private_api/repo/metadata_repo/_repo.py
+++ b/microsetta_private_api/repo/metadata_repo/_repo.py
@@ -189,19 +189,26 @@ def _fetch_survey_template(template_id):
         observed, the survey responses should not be considered valid.
     """
     with Transaction() as t:
+        error = None
+
         survey_template_repo = SurveyTemplateRepo(t)
         info = survey_template_repo.get_survey_template_link_info(
             template_id)
 
         # For local surveys, we generate the json representing the survey
-        survey_template = survey_template_repo.get_survey_template(
-            template_id, "en_US")
-        survey_template_text = vue_adapter.to_vue_schema(survey_template)
+        try:
+            survey_template = survey_template_repo.get_survey_template(
+                template_id, "en_US")
+        except NotFound as e:
+            error = e
 
-        info = info.to_api(None, None)
-        info['survey_template_text'] = survey_template_text
+        if error is None:
+            survey_template_text = vue_adapter.to_vue_schema(survey_template)
 
-        return info, None
+            info = info.to_api(None, None)
+            info['survey_template_text'] = survey_template_text
+
+        return info, error
 
 
 def _to_pandas_dataframe(metadatas, survey_templates):
--- a/microsetta_private_api/repo/metadata_repo/_repo.py
+++ b/microsetta_private_api/repo/metadata_repo/_repo.py
@@ -22,8 +22,10 @@ jsonify = json.dumps
 # ignore list.
 TEMPLATES_TO_IGNORE = {10001, 10002, 10003, 10004, None}
 
+# TODO 2022-10-03
 # Adding questions from Cooking Oils & Oxalate-rich Foods survey
-# to ignore list as they don't exist in Qiita
+# to ignore list as they don't exist in Qiita (OILS_*). We're blocked on
+# pushing them, pending an update to Qiita's API.
 EBI_REMOVE = ['ABOUT_YOURSELF_TEXT', 'ANTIBIOTIC_CONDITION',
               'ANTIBIOTIC_MED', 'PM_NAME', 'PM_EMAIL',
               'BIRTH_MONTH', 'CAT_CONTACT', 'CAT_LOCATION',
@@ -184,7 +186,7 @@ def _fetch_survey_template(template_id):
     -------
     dict
         The survey structure as returned from the private API
-    dict or None
+    string or None
         Any error information associated with the retreival. If an error is
         observed, the survey responses should not be considered valid.
     """
@@ -200,7 +202,7 @@ def _fetch_survey_template(template_id):
             survey_template = survey_template_repo.get_survey_template(
                 template_id, "en_US")
         except NotFound as e:
-            error = e
+            error = repr(e)
 
         if error is None:
             survey_template_text = vue_adapter.to_vue_schema(survey_template)
--- a/microsetta_private_api/repo/metadata_repo/_repo.py
+++ b/microsetta_private_api/repo/metadata_repo/_repo.py
@@ -18,8 +18,12 @@ jsonify = json.dumps
 # the vioscreen survey currently cannot be fetched from the database
 # there seems to be some detached survey IDs -- see 000089779
 # that account has a long and unusual history though
-TEMPLATES_TO_IGNORE = {10001, None}
+# Adding the MyFoodRepo, Polyphenol FFQ, and Spain FFQs to the
+# ignore list.
+TEMPLATES_TO_IGNORE = {10001, 10002, 10003, 10004, None}
 
+# Adding questions from Cooking Oils & Oxalate-rich Foods survey
+# to ignore list as they don't exist in Qiita
 EBI_REMOVE = ['ABOUT_YOURSELF_TEXT', 'ANTIBIOTIC_CONDITION',
               'ANTIBIOTIC_MED', 'PM_NAME', 'PM_EMAIL',
               'BIRTH_MONTH', 'CAT_CONTACT', 'CAT_LOCATION',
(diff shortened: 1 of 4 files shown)


wasade:
Can we add the fields to Qiita so they are accessible for analysis?

cassidysymons:
> Can we add the fields to Qiita so they are accessible for analysis?

I'm all for that as long as it won't cause any problems. Is that something I can/should do, or do I request it from someone else?

wasade:
I thought the code was written to automatically add to qiita but may be misremembering

wasade:
It's not, see https://github.com/biocore/microsetta-private-api/blob/3fd63e9151047a7e24be52432525eb692ddc9447/microsetta_private_api/repo/qiita_repo.py#L74-L77

@antgonza, is there an easy way to add metadata columns to a qiita study w/o actually needing to create and specify the metadata? In this case, we have variables that need to exist in qiita, but we dont yet have samples that express those variables

antgonza:
Programmatically: You could add a fake sample with all the columns you need and then delete it; this will keep the extra columns around. 

If this is only going to be performed once: You can open the DB and insert into [QIITA_COLUMN_NAME](https://github.com/qiita-spots/qiita/blob/master/qiita_db/metadata_template/base_metadata_template.py#L55) the new columns you want to add, like [this](https://github.com/qiita-spots/qiita/blob/master/qiita_db/metadata_template/base_metadata_template.py#L859).

Note that this will add new and fully empty columns to all the samples - IMOO not useful for analysis.

Hope this helps.

wasade:
Thanks! Is the programmatic method exposed and usable by an API? Is it possible to set a default value for historical samples where the variable is not collected (e.g., `Missing: not collected` or something)? I'm hoping there is an automated solution here rather than asking you to do it manually whenever new columns need to be specified 

cc @sejsong, this surrounds new questions to the questionnaire

antgonza:
Yes for [adding/updating samples/metadata](https://qiita.ucsd.edu/static/doc/html/dev/rest.html#qiita-internal-rest-api) but not for deleting (I think deleting samples can be problematic). No way to define a default value, you would need to add that value for all historical samples.

IMOO the best would be that the TMI pushes whatever changes it requires directly, either automatically or by hand ... and if there is no end-point add one (as long as it makes sense for the rest of the system).

wasade:
If I understand the existing API correctly, the metadata category has to already exist? See https://github.com/qiita-spots/qiita/blob/master/qiita_pet/handlers/rest/study_samples.py#L171-L179

Would the Qiita dev team be able to expose endpoints to allow for creation of metadata variables?

antgonza:
Good point, I didn't remember that - I wonder who added that restriction 😜.

Anyway, I'll be happy to review/merge new PRs with endpoints.

sejsong:
@antgonza is this something you might be able to do? I'm not sure we have the expertise and bandwidth on the TMI dev team to do this the most efficiently.
I'm assuming this is something that will be useful for Qiita more widely?

antgonza:
I'm not sure if it will be useful for other Qiita users (as they do not interact directly with the API - they would need access to the UCSD systems to do it). Also, that original functionality was added specifically for and by the AGP/TMI-dev-team. Now IMOO the best will be that whoever implements that functionality for this repo should issue a PR to qiita to make sure the 2 PRs work properly; does that make sense? 

Anyway, I think having an internal meeting to decide what and whom should add new code + timelines for TMI would be beneficial.

wasade:
Just spoke with Se Jin about the motivation for this. It seems surprising that samples are failing from the other surveys, but plausible. Definitely recommend testing against qiita-rc. I thought the code was setup to fail on a per survey basis, not per barcode basis, but there may be edge cases I'm misremembering

cassidysymons:
A "No such survey" error is causing a hardfail in the qiita metadata push:

qiita metadata push errors

[ { "hardfail": "Traceback (most recent call last):\n File \"/Users/api_user/microsetta-private-api-2022.9/microsetta_private_api/tasks.py\", line 76, in update_qiita_metadata\n n_pushed, error = qiita.push_metadata_to_qiita()\n File \"/Users/api_user/microsetta-private-api-2022.9/microsetta_private_api/repo/qiita_repo.py\", line 90, in push_metadata_to_qiita\n formatted, error = retrieve_metadata(to_push)\n File \"/Users/api_user/microsetta-private-api-2022.9/microsetta_private_api/repo/metadata_repo/_repo.py\", line 109, in retrieve_metadata\n survey_templates, st_errors = _fetch_observed_survey_templates(fetched)\n File \"/Users/api_user/microsetta-private-api-2022.9/microsetta_private_api/repo/metadata_repo/_repo.py\", line 159, in _fetch_observed_survey_templates\n survey, error = _fetch_survey_template(template_id)\n File \"/Users/api_user/microsetta-private-api-2022.9/microsetta_private_api/repo/metadata_repo/_repo.py\", line 190, in _fetch_survey_template\n survey_template = survey_template_repo.get_survey_template(\n File \"/Users/api_user/microsetta-private-api-2022.9/microsetta_private_api/repo/survey_template_repo.py\", line 136, in get_survey_template\n raise NotFound(\"No such survey\")\nwerkzeug.exceptions.NotFound: 404 Not Found: No such survey\n" } ]

wasade:
Thanks. It may be good to wrap `_fetch_survey_template` so errors in the future can be gracefully caught and reported, it looks like this is bringing down the entire metadata push which is bad

cassidysymons:
@wasade I added a try/catch block in `_fetch_survey_template` to gracefully handle the NotFound exception. Given the urgency to get the metadata push back up and the timeline for having Qiita changes live and available to use, do you see any issues with merging this code as-written? It's admittedly a temporary solution, but it resolves our issue for the time being and we can revisit the Qiita changes in a few weeks.

wasade:
I think that strategy works, I'm unsure if `e` needs to be `str` or `Exception`. However, it is going to be painful to recover the responses and update the metadata in Qiita so this is a bandaid and it will hurt when it gets ripped off.

cassidysymons:
@wasade When you have a minute can you please do a final review on this?

microsetta_private_api/repo/metadata_repo/_repo.py
```diff
@@ -37,7 +41,10 @@
               'COVID_SYMPTOMS_OTHER', 'FERMENTED_CONSUMED_OTHER',
               'FERMENTED_OTHER', 'FERMENTED_PRODUCE_COMMERCIAL_OTHER',
               'FERMENTED_PRODUCE_PERSONAL_OTHER',
-              'OTHER_ANIMALS_FREE_TEXT']
+              'OTHER_ANIMALS_FREE_TEXT', 'OILS_FREQUENCY_VEGETABLE',
+              'OILS_FREQUENCY_ANIMAL', 'OILS_FREQUENCY_OTHER',
+              'OILS_FREQUENCY_MARGARINE', 'OILS_FREQUENCY_OXALATE'
+              'OILS_FREQUENCY_SOY']
```
wasade, line 47: Can a `TODO` note be added regarding when these were added, and that we're blocked pending an update on Qiita's API?

microsetta_private_api/repo/metadata_repo/_repo.py
```diff
@@ -182,19 +189,26 @@ def _fetch_survey_template(template_id):
...
+        if error is None:
+            survey_template_text = vue_adapter.to_vue_schema(survey_template)
 
-        info = info.to_api(None, None)
-        info['survey_template_text'] = survey_template_text
+            info = info.to_api(None, None)
+            info['survey_template_text'] = survey_template_text
 
-        return info, None
+        return info, error
```
wasade, line 211: Does it work to return type `Exception` here? I don't know how flexible the system is with the return types but it would be good to make sure that this does work. I think other areas of the code return `dict` or `None`

wasade (changes requested):
I'm not sure if the error return is valid. It may be nice to assert it with a unit test

microsetta_private_api/repo/metadata_repo/_repo.py
cassidysymons: Done.

cassidysymons:
> I'm not sure if the error return is valid. It may be nice to assert it with a unit test

I added a unit test that intentionally calls the function with a remote survey ID to observe that an error is returned with a value other than "None." Since I've specifically cast the Exception to string before returning it, I don't think we need any further test coverage there. What are your thoughts?

microsetta_private_api/repo/metadata_repo/tests/test_repo.py
```diff
@@ -167,6 +167,14 @@ def test_fetch_survey_template(self):
         self.assertEqual(survey, exp)
         self.assertEqual(errors, None)
 
+    def test_fetch_survey_template_remote(self):
+        # attempt to fetch info for Vioscreen survey
+        survey, errors = _fetch_survey_template(10001)
+
+        # verify that _fetch_survey_template returns an error, reflecting
+        # that it's a remote survey for which we can't extract local data
+        self.assertNotEqual(errors, None)
```
wasade, line 176: Shouldn't errors be somethign different than None?

cassidysymons: I'm asserting that errors isn't None in this case using assertNotEqual, which represents the intended behavior. We don't strictly care what the content of the error is, but having a non-None value there will prevent the metadata push from attempting to use that survey.

wasade: Oh, ya read that fast :) thanks

PR #461, opened 2022-09-27
API: add a method to obtain barcode -> vioscreen users

wasade:
We do not have an exposed method to obtain a barcode to vioscreen user association, which is necessary to transform extracted Vioscreen data for analysis


cassidysymons:
I'm +1 on this, anything else you wanted to add or am I good to merge?

wasade:
I think it's good, thanks!

PR #462, opened 2022-09-29
Nested daklapack responses

AmandaBirmingham:
Supports capturing kit info from within one level of nesting in daklapack responses, for kits nested under a "Container-content"-type scannableKitItem.  Note that other types and/or levels of nesting remain unsupported.


microsetta_private_api/admin/daklapack_polling.py
```diff
@@ -140,41 +146,88 @@ def process_order_articles(admin_repo, order_id, status, create_date):
...
+
+            # DON'T change this to "is_multi_item" bc False != None here
+            if is_multi_item is False:
+                raise ValueError(MULTI_ITEM_ERR_MSG)
+            else:
+                is_multi_item = True
+
+            curr_items = curr_scannable[CONTAINER_ITEMS_KEY]
+            for curr_item in curr_items:
+                curr_item_details = curr_item["containerItemDetails"]
```
wasade, line 198: Many of the other keys are global variables, should this one be as well?

microsetta_private_api/admin/daklapack_polling.py
```diff
@@ -140,41 +146,88 @@ def process_order_articles(admin_repo, order_id, status, create_date):
...
+                    admin_repo, order_proj_ids, order_id, outbound_fedex_code,
+                    inbound_fedex_code, address_dict, curr_item_details)
+                created_kits.append(curr_kit_info)
+
+        elif curr_scannable_type in SINGLE_KIT_TYPES:
+            is_multi_item = False
+
+            # in theory, at this point I could break from the loop, but I'm
+            # continuing to loop over all the kit items so I can detect if
+            # single-item and multi-item info are both present (which is Bad)
```
wasade, line 210: +1

wasade (commented):
Just a minor comment -- I think this is good, but will defer to @cassidysymons for final review

microsetta_private_api/admin/daklapack_polling.py
AmandaBirmingham: I usually make them global variables if I use them in more than one place, which I didn't with this one.  However, I'm happy to make 'em all global variables if that is preferable.

cassidysymons:
I'm comfortable with the logic on global variables. Code was tested on staging and correctly handles Daklapack's updated application of their API logic relative to TMI's needs.

PR #470, opened 2022-10-27
Scrub bug

### LINKED ISSUE #469: Bug on account delete when a user has secondary surveys
Originally noted by @cassidysymons, as he noted:

"I just discovered an interesting bug in the account deletion functionality. When a user has taken any of the three optional surveys that are remote - MyFoodRepo, Spanish FFQ, or Polyphenol FFQ - deleting their account will fail. Digging into it, all of the registry tables for those external surveys are set up similarly (I believe I patterned SFFQ and PFFQ after MFR), with the key aspect being that account_id is a foreign key reference ag.account.id and more importantly, is not null."

On inspection, the criteria to delete rather than scrub is based on the presence of samples associated to the account, and does not explicitly test for surveys which may depend on account / source relationships.

wasade:
Fixes #469 

To do so, we needed to add admin-intended deletion methods for the external surveys in order support the testing framework. 

If a user has any external surveys, their account is scrubbed rather than deleted so the relationships can be preserved.

### DIFF
--- a/microsetta_private_api/admin/admin_impl.py
+++ b/microsetta_private_api/admin/admin_impl.py
@@ -796,6 +796,7 @@ def delete_account(account_id, token_info):
         src_repo = SourceRepo(t)
         samp_repo = SampleRepo(t)
         sar_repo = SurveyAnswersRepo(t)
+        template_repo = SurveyTemplateRepo(t)
 
         acct = acct_repo.get_account(account_id)
         if acct is None:
@@ -806,6 +807,7 @@ def delete_account(account_id, token_info):
                 return None, 204
 
         sample_count = 0
+        account_has_external = False
         sources = src_repo.get_sources_in_account(account_id)
 
         for source in sources:
@@ -813,6 +815,11 @@ def delete_account(account_id, token_info):
 
             has_samples = len(samples) > 0
             sample_count += len(samples)
+            has_external = template_repo.has_external_surveys(account_id,
+                                                              source.id)
+
+            if has_external:
+                account_has_external = True
 
             for sample in samples:
                 # we scrub rather than disassociate in the event that the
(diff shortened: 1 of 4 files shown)


microsetta_private_api/admin/admin_impl.py
```diff
@@ -826,15 +833,17 @@ def delete_account(account_id, token_info):
                 for survey_id in surveys:
                     sar_repo.scrub(account_id, source.id, survey_id)
                 src_repo.scrub(account_id, source.id)
-            else:
-                # if we do not have associated samples, then the source
-                # is safe to delete
+
+            if not has_samples and not has_external:
```
cassidysymons, line 837: It looks like sources that don't have samples but do have external surveys (`if not has samples and has_external`) is being missed here. Should line 830 be changed to `if has_samples or has_external`?

microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -339,6 +339,37 @@ def set_myfoodrepo_id(self, account_id, source_id, mfr_id):
                            WHERE account_id=%s AND source_id=%s""",
                         (mfr_id, account_id, source_id))
 
+    def delete_myfoodrepo(self, account_id, source_id):
+        """Intended for admin use, remove MyFoodRepo entries
+
+        This method is idempotent
+
+        This method deletes ALL surveys associated with an account and source
```
cassidysymons, line 347: This comment in each of the delete_[remote survey] functions should probably be re-worded since these functions only delete specific surveys, rather than all surveys associated with the account/source.

cassidysymons (changes requested):

microsetta_private_api/admin/admin_impl.py
wasade: I think the current logic is correct? We only want to delete a source if and only if it does not have samples and does not have an external survey?

microsetta_private_api/repo/survey_template_repo.py
wasade: Done

microsetta_private_api/admin/admin_impl.py
cassidysymons: If the source has samples (line 830), it gets scrubbed (on line 835). If the source has neither samples nor external surveys (line 837), it gets deleted (on line 842). There's no condition for a source that does not have samples but does have external surveys, so it would be neither deleted nor scrubbed.

wasade: ah i see, done, good spot

cassidysymons: Great, thanks. I'll push it to staging for testing whenever Jeff get the filesystem issues resolved.

PR #476, opened 2022-11-01
Adding Data and Bio Consent through DB

raginirai553:


microsetta_private_api/repo/consent_repo.py
```diff
@@ -0,0 +1,236 @@
...
+                        "VALUES( %s, %s, %s, %s, %s, "
+                        "%s, %s) ",
+                        _consent_document_to_row(consent))
+            return cur.rowcount == 1
+
+    def get_all_consent_documents(self, tag):
+        consent_docs = []
+
+        with self._transaction.dict_cursor() as cur:
+            cur.execute("SELECT " + self.doc_read_cols + " FROM "
```
cassidysymons, line 89: The way this query is written, it will return every single document for a given language tag, even if some of them have been replaced. We only want to return the most recent version of each document type.

microsetta_private_api/api/_consent.py
```diff
@@ -1,18 +1,68 @@
...
+
+def get_consent_doc(account_id, consent_id, token_info):
+    _validate_account_access(token_info, account_id)
+
+    with Transaction() as t:
+        consent_repo = ConsentRepo(t)
+        documents = consent_repo.get_consent_document(consent_id)
+        data = [x.to_api() for x in documents]
+
+    return jsonify(data), 200
```
cassidysymons, line 37: Shouldn't this return a 404 error if data is None, rather than an HTTP code of 200?

microsetta_private_api/repo/consent_repo.py
```diff
@@ -0,0 +1,236 @@
...
+                return _row_to_consent_document(r)
+
+    def is_consent_required(self, source_id, consent_type):
+        with self._transaction.dict_cursor() as cur:
+            cur.execute("SELECT " + self.signature_read_cols + " FROM "
+                        "ag.consent_audit INNER JOIN ag.consent_documents "
+                        "USING (consent_id) "
+                        "WHERE consent_audit.source_id = %s and "
+                        "consent_documents.consent_type "
+                        "LIKE %s ORDER BY sign_date DESC",
```
cassidysymons, line 117: Why are you using LIKE rather than = here?

microsetta_private_api/api/_consent.py
```diff
@@ -1,18 +1,68 @@
...
+    return jsonify(consent_form), 200
+
+
+def get_consent_doc(account_id, consent_id, token_info):
+    _validate_account_access(token_info, account_id)
+
+    with Transaction() as t:
+        consent_repo = ConsentRepo(t)
+        documents = consent_repo.get_consent_document(consent_id)
+        data = [x.to_api() for x in documents]
```
cassidysymons, line 35: Since get_consent_document() can only return one result, wouldn't it be clearer to express without the for/in loop?

microsetta_private_api/api/_consent.py
```diff
@@ -1,18 +1,68 @@
...
+
+    return jsonify(data), 200
+
+
+def check_consent_signature(account_id, source_id, consent_type, token_info):
+    _validate_account_access(token_info, account_id)
+
+    res = True
```
cassidysymons, line 43: Is it necessary to have a default state here? Or to turn the question around, in what situation will ConsentRepo.is_consent_required not return an actionable result?

microsetta_private_api/api/_consent.py
```diff
@@ -1,18 +1,68 @@
...
+
+    with Transaction() as t:
+        consent_repo = ConsentRepo(t)
+        res = consent_repo.is_consent_required(source_id, consent_type)
+
+    return jsonify({"result": res}), 200
+
+
+def sign_consent_doc(account_id, source_id, consent_type, body, token_info):
```
cassidysymons, line 52: Since ConsentRepo.sign_consent can raise NotFound, it would be good to catch those and express them appropriately in the response, if they were to occur.

microsetta_private_api/db/patches/0106.sql
```diff
@@ -0,0 +1,2364 @@
...
+    source_id uuid NOT NULL,
+    date_time TIMESTAMP WITH TIME ZONE NOT NULL,
+    PRIMARY KEY (signature_id),
+    FOREIGN KEY (consent_id) REFERENCES ag.consent_documents (consent_id),
+    FOREIGN KEY (source_id) REFERENCES ag.source (id)
+);
+
+INSERT INTO ag.account 
+    ("id", "email", "account_type", "first_name", "last_name",
+     "street", "city", "post_code", "country_code", "preferred_language",
```
cassidysymons, line 27: There's no reason to insert street, city, post_code, country_code, preferred_language, address_verified, or cannot_geocode for the consent dummy account.

microsetta_private_api/db/patches/0106.sql
```diff
@@ -0,0 +1,2364 @@
...
+     "address_verified", "cannot_geocode")
+     VALUES ('000fc4cd-8fa4-db8b-e050-8a800c5d81b7', 'microsetta+consentdummy@ucsd.edu', 'standard', 'demo', 'demo', 'demo', 'demo', '46839', 'US', 'en_US', 'true', 'false');
+
+INSERT INTO ag.consent_documents ("consent_type", "locale", "date_time", "consent_content", "reconsent_required", "account_id")
+    VALUES ('Adolescent Consent - Biospecimen', 'en_US', NOW(), '<p class="consent_title">
```
cassidysymons, line 32: For all of the various consent types, can we stick to [age-group]_[type]? The various permutations should be as follows:
adolescent_biospecimen
adolescent_data
adult_biospecimen
adult_data
child_biospecimen
child_data
parent_biospecimen
parent_data

Including "Consent" in the string is redundant for certain types and inaccurate for others (child/adolescent docs are technically assent, not consent).

Please also make this change in the Interface PR so that everything matches up between the two.

microsetta_private_api/model/consent.py
```diff
@@ -0,0 +1,86 @@
+from datetime import datetime
+
+
+class ConsentDocument:
+    @staticmethod
+    def from_dict(input_dict, account_id, consent_id):
+        consent_type = input_dict["consent_type"]
+        locale = input_dict["locale"]
+        date_time = datetime.now()
```
cassidysymons, line 9: Shouldn't the date_time be coming from the record in the database rather than now()?

microsetta_private_api/model/consent.py
```diff
@@ -0,0 +1,86 @@
...
+    def from_dict(input_dict, source_id, signature_id):
+        date_time = datetime.now()
+
+        parent_name_1 = input_dict.get("parent_1_name", None)
+
+        parent_name_2 = input_dict.get("parent_2_name", None)
+
+        deceased_parent = input_dict.get("deceased_parent", None)
+
+        assent_obtainer = input_dict.get("assent_obtainer", None)
```
cassidysymons, line 52:
```suggestion
        parent_name_1 = input_dict.get("parent_1_name")
        parent_name_2 = input_dict.get("parent_2_name")
        deceased_parent = input_dict.get("deceased_parent")
        assent_obtainer = input_dict.get("assent_obtainer")
```
get() will return None by default if the key doesn't exist, no need to declare it.

microsetta_private_api/repo/consent_repo.py
```diff
@@ -0,0 +1,236 @@
...
+
+
+def _row_to_consent_document(r):
+    return ConsentDocument(
+        r["consent_id"],
+        r["consent_type"],
+        r["locale"],
+        r["date_time"],
+        r["consent_content"],
+        getattr(r, 'account_id', None),
```
cassidysymons, line 27: It should be impossible for a consent document to not have an account_id.

cassidysymons (changes requested):

microsetta_private_api/api/_consent.py
raginirai553: Corrected!

microsetta_private_api/api/_consent.py
raginirai553: Corrected!

microsetta_private_api/db/patches/0106.sql
raginirai553: Corrected!

microsetta_private_api/model/consent.py
raginirai553: Corrected!

microsetta_private_api/repo/consent_repo.py
raginirai553: @cassidysymons The query is designed always fetch only one record of latest consent document for each type. self.doc_read_cols contains distinct constraint on consent_type ensuring unique document for each consent further combined with order by clause as per date_time in descending order ensuring that we always get latest consent documents.

microsetta_private_api/repo/consent_repo.py
raginirai553: We are using like query instead of equals becasue the exact consent type is not known at this instance. There are broadly two classifications of consent i.e., data and biospecimen. While checking the consent status for any source, the age range for the source is not known at the interface end. Therefore, one cannot make any exact calls for particular consent from the interface. However, the broader category i.e., data or biospecimen is known. Although there are workarounds including extraction of age_range from the source table given source id, programattically create string for exact consent_document type and look for the said document in the query. However, to avoid the workaround, accurate match can be found on basis of consent_type LIKE and source_id, since source_id is a foreign key in consent_audit table. 
Further, there may exist multiple consents provided by a source but we intend to find the consent from one of the broader categories, that too latest signed.

microsetta_private_api/api/_consent.py
raginirai553: Removed the default initialization.

microsetta_private_api/api/_consent.py
raginirai553: Try-except block added to handle errors appropriately.

microsetta_private_api/db/patches/0106.sql
raginirai553: Since all these fields have NOT NULL constraints applied in the database, it was necessary to include these in the DB Patch.

microsetta_private_api/model/consent.py
raginirai553: The purpose of this static method is to create consent document object from the post request by admin to create new consent and therefore, date_time is generated using now(). This method is not used right now and is a provision for future.

microsetta_private_api/repo/consent_repo.py
raginirai553: This was written for testing purposes. Changes have been reverted.

microsetta_private_api/repo/consent_repo.py
raginirai553: @cassidysymons I have tested it few times already and it always returns the latest consent content

cassidysymons: Sorry, I missed the distinct clause in read_cols. Looks good.

microsetta_private_api/api/_consent.py
```diff
@@ -31,17 +32,18 @@ def get_consent_doc(account_id, consent_id, token_info):
...
         consent_repo = ConsentRepo(t)
-        documents = consent_repo.get_consent_document(consent_id)
-        data = [x.to_api() for x in documents]
+        document = consent_repo.get_consent_document(consent_id)
+        data = document.to_api()
+
+    if len(data) == 0:
+        return jsonify(code=404, message=CONSENT_DOC_NOT_FOUND_MSG) , 404
 
     return jsonify(data), 200
```
cassidysymons, line 41: This can be more concise:

```
    document = consent_repo.get_consent_document(consent_id)
    
if document:
    data = document.to_api()
    return jsonify(data), 200
else:
    return jsonify(code=404, message=CONSENT_DOC_NOT_FOUND_MSG) , 404
```

microsetta_private_api/db/patches/0106.sql
cassidysymons: Ah, for some reason I thought those fields were nullable. Please just change address_verified to false.

microsetta_private_api/repo/consent_repo.py
raginirai553: No, its all working as expected

microsetta_private_api/api/_consent.py
raginirai553: Updated!

microsetta_private_api/db/patches/0106.sql
raginirai553: Changed!

Issue #468, opened 2022-10-26
Bug with project association on campaign creation


cassidysymons:
When creating a campaign, this block of code in the create_campaign function is behaving inconsistently:

                cur.executemany(
                    "INSERT INTO campaign.campaigns_projects ("
                    "campaign_id,project_id"
                    ") VALUES (%s, %s) ",
                    [(campaign_id, pid) for pid in associated_projects]
                )

In some cases, it treats a single ID properly (e.g. 155 is inserted as 155), while in others, it treats each numeral in the ID as a different entry (e.g. 157 is inserted as three entries - 1, 5, and 7). Need to refactor this code to explicitly treat a comma as the only delimiter, rather than relying on default behavior.

wasade:
Maybe add 

```python
if isinstance(associated_projects, str):
    associated_projects = [associated_projects, ]
```

immediately prior?

PR #477, opened 2022-11-02
Changes for THDMI Japan signups

cassidysymons:
The scope of this PR is to support THDMI Japan signups using the Interested Users form on the existing platform. The actual THDMI Japan campaign will be launched on the new version of the TMI platform, so these PRs explicitly exclude Japanese translations for the consent documents and surveys.

To prevent the edge case of an existing account changing their account to Japanese and/or someone with an existing Kit ID/Activation Code from creating an account and selecting Japanese, I've prevented it from displaying as a language option in the Account Details screen.

Added:
- Confirmation email for THDMI signups
- True/False flag in campaign.campaigns table to determine whether confirmation email should be sent
- ja_JP locale

Fixed:
- Issue #468

### DIFF
--- a/microsetta_private_api/api/_interested_user.py
+++ b/microsetta_private_api/api/_interested_user.py
@@ -6,6 +6,7 @@ from microsetta_private_api.repo.interested_user_repo import InterestedUserRepo
 from microsetta_private_api.repo.transaction import Transaction
 from microsetta_private_api.exceptions import RepoException
 from microsetta_private_api.repo.campaign_repo import CampaignRepo
+from microsetta_private_api.tasks import send_email
 
 
 def create_interested_user(body):
--- a/microsetta_private_api/api/_interested_user.py
+++ b/microsetta_private_api/api/_interested_user.py
@@ -5,6 +5,7 @@ from microsetta_private_api.model.interested_user import InterestedUser
 from microsetta_private_api.repo.interested_user_repo import InterestedUserRepo
 from microsetta_private_api.repo.transaction import Transaction
 from microsetta_private_api.exceptions import RepoException
+from microsetta_private_api.repo.campaign_repo import CampaignRepo
 
 
 def create_interested_user(body):
@@ -28,6 +29,25 @@ def create_interested_user(body):
                 code=400,
                 message="Failed to create interested user."
             ), 400
+
+        if interested_user_id:
+            campaign_repo = CampaignRepo(t)
+            campaign_info =\
+                campaign_repo.get_campaign_by_id(interested_user.campaign_id)
+
+            if campaign_info.send_thdmi_confirmation:
+                try:
+                    # Send a confirmation email
+                    # TODO: Add more intelligent locale determination.
+                    # Punting on that since our current campaign use cases
+                    # are only a single language.
+                    send_email(interested_user.email,
+                               "submit_interest_confirmation",
+                               {"contact_name": interested_user.first_name},
+                               campaign_info.language_key)
+                except:  # noqa
+                    # try our best to email
+                    pass
 
         t.commit()
 
--- a/microsetta_private_api/api/_interested_user.py
+++ b/microsetta_private_api/api/_interested_user.py
@@ -31,24 +31,23 @@ def create_interested_user(body):
                 message="Failed to create interested user."
             ), 400
 
-        if interested_user_id:
-            campaign_repo = CampaignRepo(t)
-            campaign_info =\
-                campaign_repo.get_campaign_by_id(interested_user.campaign_id)
+        campaign_repo = CampaignRepo(t)
+        campaign_info =\
+            campaign_repo.get_campaign_by_id(interested_user.campaign_id)
 
-            if campaign_info.send_thdmi_confirmation:
-                try:
-                    # Send a confirmation email
-                    # TODO: Add more intelligent locale determination.
-                    # Punting on that since our current campaign use cases
-                    # are only a single language.
-                    send_email(interested_user.email,
-                               "submit_interest_confirmation",
-                               {"contact_name": interested_user.first_name},
-                               campaign_info.language_key)
-                except:  # noqa
-                    # try our best to email
-                    pass
+        if campaign_info.send_thdmi_confirmation:
+            try:
+                # Send a confirmation email
+                # TODO: Add more intelligent locale determination.
+                # Punting on that since our current campaign use cases
+                # are only a single language.
+                send_email(interested_user.email,
+                           "submit_interest_confirmation",
+                           {"contact_name": interested_user.first_name},
+                           campaign_info.language_key)
+            except:  # noqa
+                # try our best to email
+                pass
 
         t.commit()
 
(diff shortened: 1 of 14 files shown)


microsetta_private_api/api/_interested_user.py
```diff
@@ -29,6 +31,25 @@ def create_interested_user(body):
...
 
+        if interested_user_id:
+            campaign_repo = CampaignRepo(t)
+            campaign_info =\
+                campaign_repo.get_campaign_by_id(interested_user.campaign_id)
+
+            if campaign_info.send_thdmi_confirmation:
+                try:
+                    # Send a confirmation email
+                    # TODO: Add more intelligent locale determination.
```
wasade, line 42: Would it be valuable to create a github issue on this?

microsetta_private_api/db/patches/0103.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE campaign.campaigns ADD COLUMN send_thdmi_confirmation BOOLEAN DEFAULT FALSE;
```
wasade, line 1: Could this also be remarked as `NOT NULL`?

microsetta_private_api/repo/campaign_repo.py
```diff
@@ -86,24 +88,37 @@ def create_campaign(self, **kwargs):
...
+                # This code is not ideal but the various campaign creation
+                # methods pass several different datatypes in.
+                if isinstance(associated_projects, tuple):
+                    associated_projects = list(associated_projects)
+                elif isinstance(associated_projects, str):
+                    if "," in associated_projects:
+                        associated_projects = associated_projects.split(",")
+                    else:
+                        associated_projects = [associated_projects, ]
+                elif isinstance(associated_projects, int):
```
wasade, line 119: Does this really happen? Isn't the query below sensitive to whether an associated project entry is `int` or `str`?

microsetta_private_api/templates/email/submit_interest_confirmation.jinja2
```diff
@@ -0,0 +1,249 @@
...
+    </style>
+</head>
+<body class="c10">
+<p class="c1 c5"><span class="c0"></span></p>
+<p class="c1">
+    <span class="c0">{{ _("Dear") }} {{contact_name |e}},</span>
+</p>
+<p class="c1"><span class="c0">&nbsp;</span></p>
+<p class="c8 c9">
+    <span class="c4">{{ _("On behalf of Danone Nutricia Research and The Microsetta Initiative (TMI) at UC San Diego, we would like to thank you for your interest in participating in our research study THDMI - The Human Diets and Microbiome Initiative.") }}</span>
```
wasade, line 224: I'm not QC'ing the content -- assuming that already went through R&E

microsetta_private_api/repo/campaign_repo.py
```diff
@@ -86,24 +88,37 @@ def create_campaign(self, **kwargs):
...
                  accepting_participants, language_key_alt, title_alt,
-                 instructions_alt)
+                 instructions_alt, send_thdmi_confirmation)
             )
             campaign_id = cur.fetchone()[0]
 
             if campaign_id is None:
                 raise RepoException("Error inserting campaign into database")
             else:
+                # This code is not ideal but the various campaign creation
```
wasade, line 110: This is the block associated with bug #468. Our standard practice is to add regression tests that trigger identified bugs, so we can verify the bug is fixed, and so we can ensure it doesn't occur again in the future. Would it be possible to add a test that triggers #468?

wasade (changes requested):
Thanks! Looks pretty good, but had a few small requests and questions along the way

microsetta_private_api/api/_interested_user.py
cassidysymons: #478

microsetta_private_api/templates/email/submit_interest_confirmation.jinja2
wasade: Makes sense

microsetta_private_api/repo/campaign_repo.py
cassidysymons: It doesn't happen from browser-based usage, but internally, unit tests were providing int, tuple, and list in various scenarios. Ideally, it would be good to standardize internal usage but I'd prefer not to get mired in that at the moment.

wasade: Fair

microsetta_private_api/db/patches/0103.sql
cassidysymons: Done.

wasade: Thanks!

microsetta_private_api/repo/campaign_repo.py
```diff
@@ -86,24 +88,37 @@ def create_campaign(self, **kwargs):
...
+                # methods pass several different datatypes in.
+                if isinstance(associated_projects, tuple):
+                    associated_projects = list(associated_projects)
+                elif isinstance(associated_projects, str):
+                    if "," in associated_projects:
+                        associated_projects = associated_projects.split(",")
+                    else:
+                        associated_projects = [associated_projects, ]
```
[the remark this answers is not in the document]
cassidysymons: Good call, switched to just performing a split regardless of whether a comma is present.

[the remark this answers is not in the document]
wasade: Thanks!

microsetta_private_api/api/_interested_user.py
```diff
@@ -29,6 +31,25 @@ def create_interested_user(body):
                 message="Failed to create interested user."
             ), 400
 
+        if interested_user_id:
```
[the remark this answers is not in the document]
wasade: Thanks!

microsetta_private_api/db/patches/0103.sql
cassidysymons: Adding the NOT NULL constraint broke numerous tests (https://github.com/biocore/microsetta-private-api/actions/runs/3387613254/jobs/5628583258). Considering it's a non-essential field, has a default value, and we already have a boolean column in this table that's nullable with a default value (accepting participants), I'd recommend reverting the NOT NULL constraint rather than modifying all of these functions to explicitly declare a value for the column. Thoughts?

wasade: It's a small pain now vs. a potential large pain later. If you're slammed, I can issue a PR against this PR to resolve it

cassidysymons: All good, realized I could just add a default False value in the get() function right before the insert statement.

wasade: Great, thanks!

microsetta_private_api/repo/campaign_repo.py
cassidysymons: Added `test_create_campaign_projects_bug` to verify several possible scenarios. Are there any other scenarios I should specifically exercise in that test?

wasade: Thank you! I usually just do enough to trigger the bug. 

My general bug workflow is:

- determine the most likely lowest point in the stack where the bug is occurring
- implement a test to recreate the bug
- verify bug can be triggered by test
- write code to fix the bug

PR #466, opened 2022-10-11
Return previous survey results

charles-cowart:
Missing unittests

PR #474, opened 2022-10-29
migrating survey to updated structure

charles-cowart:
Work in Progress.
Triggers need to be added.
Some remapped questions need to be remade into new questions due to constraints and new/changed/deleted options.
Migraines category needs to be merged and removed. Basic testing w/UI needed.

PR #489, opened 2022-11-22
Survey database update v2

charles-cowart:
Alternative approach to survey/data migration that preserves retired survey templates/groups.
Working on final four broken unittests and adding additional tests as appropriate.


microsetta_private_api/db/migration_support.py
```diff
@@ -742,6 +742,17 @@ def migrate_96(TRN):
                     (r[0], hsi))
         TRN.execute()
 
+    @staticmethod
+    def migrate_107(TRN):
```
wasade, line 746: It doesn't look like this makes any changes, is it necessary?

microsetta_private_api/db/patches/0107.sql
```diff
@@ -0,0 +1,1643 @@
...
+update ag.survey_question set american = 'How often do you cook and consume home cooked meals? (Exclude ready-to-eat meals like boxed macaroni and cheese, ramen noodles, lean cuisine)' where survey_question_id = 57;
+update ag.survey_question set american = 'How often do you consume ready-to-eat meals (e.g. macaroni and cheese, ramen noodles, lean cuisine)?' where survey_question_id = 58;
+update ag.survey_question set american = 'How often do you eat food prepared at a restaurant, including carry-out/take-out?' where survey_question_id = 59;
+update ag.survey_question set american = 'How often do you eat at least 2 servings of whole grains in a day? (1 serving = 1 slice of 100% whole grain bread; 1 cup whole grain cereal like Shredded Wheat, Wheaties, Grape Nuts, high fiber cereals, or oatmeal; 3-4 whole grain crackers; 1⁄2 cup brown rice or whole wheat pasta)' where survey_question_id = 91;
+update ag.survey_question set american = 'How often do you consume at least 2-3 servings of fruit in a day? (1 serving = 1⁄2 cup (1 serving = 1⁄2 cup fruit juice.)' where survey_question_id = 61;
```
wasade, line 149: Unusual character "1⁄2"

microsetta_private_api/db/patches/0107.sql
```diff
@@ -0,0 +1,1643 @@
+-- STEP 1: add survey_template_id to ag_login_surveys table and populate it.
+alter table ag.ag_login_surveys add column survey_template_id integer;
```
wasade, line 2: Can an english description of what the table creations are doing be added please?

microsetta_private_api/repo/survey_answers_repo.py
```diff
@@ -104,13 +103,127 @@ def survey_template_id_and_status(self, survey_answers_id):
...
+        if status is not None:
+            return [(survey_id, SurveyTemplateRepo.VIOSCREEN_ID)]
+
+        with self._transaction.cursor() as cur:
+            # Polyphenol FFQ
+            try:
+                # If survey_id isn't a valid UUID, a ValueError will be
+                # raised.
+                uuid.UUID(survey_id)
+                cur.execute("SELECT EXISTS (SELECT polyphenol_ffq_id FROM "
```
charles-cowart, line 149: Remove potential SQL injection vectors

microsetta_private_api/db/migration_support.py
```diff
@@ -753,6 +764,7 @@ def migrate_96(TRN):
         # "0082.sql": migrate_82.__func__
         # ...
         "0096.sql": migrate_96.__func__,
+        "0105.sql": migrate_107.__func__,
```
charles-cowart, line 767: 0107.sql

wasade (changes requested):
Many of the comments from #474 appear outstanding, could those please be addressed as well?

microsetta_private_api/repo/survey_answers_repo.py
charles-cowart: No potential SQL-injection vectors appeared when reviewing the changes in this PR. Tentatively calling it closed.

microsetta_private_api/db/migration_support.py
charles-cowart: Fixed.

microsetta_private_api/db/migration_support.py
charles-cowart: Will remove this last, once it's demonstrated that it's not needed.

microsetta_private_api/db/migration_support.py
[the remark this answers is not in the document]
charles-cowart: Noted.

microsetta_private_api/db/patches/0107.sql
charles-cowart: Removed all non-ASCII characters and replaced them with their ASCII equivalents, except where 'crème fraiche' was used.

microsetta_private_api/db/patches/0107.sql
charles-cowart: Done!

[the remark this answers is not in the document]
charles-cowart: Working on it.

microsetta_private_api/repo/survey_answers_repo.py
wasade: Pass in arguments to `.execute`, do not use string formatting

charles-cowart: Just to be clear, we're arguing there's a difference between:
> cur.execute("SELECT EXISTS (SELECT polyphenol_ffq_id FROM ag.polyphenol_ffq_registry WHERE polyphenol_ffq_id='%s')" % survey_id)
and
> cur.execute("SELECT EXISTS (SELECT polyphenol_ffq_id FROM ag.polyphenol_ffq_registry WHERE polyphenol_ffq_id=%s)" % (survey_id,))

wasade: Yes, see https://stackoverflow.com/a/45128973/19741

wasade: We are not arguing: this pull request will not be merged with unsafe use of SQL.

charles-cowart: @wasade right, I didn't mean to imply we were arguing, just that there is in fact a difference between those two similar looking statements.

microsetta_private_api/repo/survey_answers_repo.py
```diff
@@ -104,13 +103,140 @@ def survey_template_id_and_status(self, survey_answers_id):
...
+
+    def _local_survey_template_ids_from_survey_ids2(self, survey_ids):
+        with self._transaction.cursor() as cur:
+            # note these ids are unique string ids, not integer ids.
+            ids = [f"'{x}'" for x in survey_ids]
+
+            sql = ("SELECT survey_id, survey_template_id FROM "
+                   "ag.ag_login_surveys WHERE survey_id IN %s", (ids,))
+
+            cur.execute(sql)
```
wasade, line 190: This doesn't require a tuple expansion?

microsetta_private_api/repo/survey_answers_repo.py
```diff
@@ -104,13 +103,140 @@ def survey_template_id_and_status(self, survey_answers_id):
...
+
+            sql = ("select a.barcode, a.survey_id, b.survey_question_id, "
+                   "c.survey_group, d.survey_id as survey_template_id from "
+                   "source_barcodes_surveys a, survey_answers b, "
+                   "group_questions c, surveys d where a.survey_id = "
+                   "b.survey_id and b.survey_question_id = "
+                   "c.survey_question_id and c.survey_group = d.survey_group"
+                   " and b.survey_id in %s", (ids,))
+
+            cur.execute(sql)
```
wasade, line 221: Same here re: tuple expansion? This is passing `cur.execute((foo, bar))`, but it should be `cur.execute(foo, bar)`

I haven't checked other areas of code so this may occur in additional places

microsetta_private_api/repo/survey_answers_repo.py
charles-cowart: doh. You literally caught that at the same time as CI. nice!

charles-cowart: Actually it didn't require the dereference *; instead it needed to be converted into a proper tuple from a list.

microsetta_private_api/repo/survey_answers_repo.py
charles-cowart: three in total. just fixed them and ran them through tests locally. Should be going through CI right now.

microsetta_private_api/db/patches/0107.sql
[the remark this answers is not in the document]
charles-cowart: Added update statements to populate remove survey rows in ag_login_surveys, added no null constraint to the survey_template_id column, and updated inserts in the code and tests to support the new requirement. Tentatively closing.

microsetta_private_api/repo/survey_answers_repo.py
charles-cowart: @wasade tentatively resolved. all tests working.

wasade: Great, thanks!

microsetta_private_api/repo/survey_answers_repo.py
charles-cowart: tentatively closed (see related comment above).

charles-cowart:
@cassidysymons would you still prefer this PR to point to master-overhaul instead of master? I'm reviewing old comments from the v1 pr.

cassidysymons:
@charles-cowart Yes, please point it at master-overhaul. Thanks!

cassidysymons:
@charles-cowart Why are you still migrating the old surveys into new templates? Didn't the idea of retiring templates eliminate the need for that migration?

charles-cowart:
@cassidysymons The idea of retiring templates was to allow us to migrate data without having to perform a lot of CRUD operations over the legacy survey data. Otherwise we'd need to delete legacy template information and migrate all legacy responses over to new templates. This is a large operation over valuable data and can't be undone, as Daniel pointed out.

The update code is relatively straightforward to audit, reuses survey create and delete functions, and has unit tests. This should help to inspire confidence in its robustness. Creating new surveys allows us to migrate and test non-destructively with respect to the legacy data.

get_survey_metadata() doesn't return survey_ids along with the survey metadata it returns, so migrating it on the fly seem s possible, but it would break the service if a user or developer (now or in the future) expects a BASIC_INFO template to exist and it doesn't.

wasade:
I'm also surprised by the need for migration. The `ag.survey_answers` table does not care about what template a `survey_question_id` is associated with. I thought the plan was to update the associations with `ag.survey_question` and `ag.group_questions`, and to not touch `ag.survey_answers`

cassidysymons:
@charles-cowart When we discussed the strategy of retiring templates, we discussed adding survey_template_id to ag_login_surveys, so that the only migration necessary would be adding the template_id to that table, rather than creating new records for the old responses. Did something come up that made that unfeasible?

@wasade Based on [the comments for this migration that Dan Hakim wrote](https://github.com/biocore/microsetta-private-api/commit/9f96eb59ef749100d476dfbc3d54df2fdcb40361), we do need to be able to determine the survey_template_id from a record in ag_login_surveys. However, retroactively adding it via a migration seems like the least invasive option at our disposal.

wasade:
I thought we were making `survey_template_id` explicit in `ag_login_surveys` in this PR. Maybe we're mincing words, but I wouldn't consider that a migration as we aren't moving data but creating it.

cassidysymons:
That was my understanding too, so I think we only have a semantics difference. I think the change would need to use the migration_support.py functionality as it would be incredibly complex to achieve entirely via sql, but yes, it's creating data, not moving it.

charles-cowart:
> @charles-cowart When we discussed the strategy of retiring templates, we discussed adding survey_template_id to ag_login_surveys, so that the only migration necessary would be adding the template_id to that table, rather than creating new records for the old responses. Did something come up that made that unfeasible?

My understanding was that we'd be able to create the new surveys on-demand if we called this new functionality in get_survey_metadata(), leaving the legacy data alone. Or we could call it once for every barcode in migration. The advantages of the latter is we process a barcode once and only once, without having to add a column or some other metadata to denote that we've migrated a barcode and don't do it again on-demand. However we always needed to maintain the constraint that a survey_id is associated with one and only one template_id permanently. It breaks the code if we don't, and it makes the data inconsistent in survey_answers.

charles-cowart:
> That was my understanding too, so I think we only have a semantics difference. I think the change would need to use the migration_support.py functionality as it would be incredibly complex to achieve entirely via sql, but yes, it's creating data, not moving it.

The population of the new survey_template_id column in ag_login_surveys is actually done in the 0108.sql file. I generate three temp tables at the top of the file and use the successive results to populate the column before deleting them.

cassidysymons:
If we add the survey_template_id to ag_login_surveys, and leave all of the old survey templates intact (but retired), why do we need to do any sort of migration?

charles-cowart:
Confirmed the lone error in tests is caused by comparing the observed results of _migrate_responses() with the expected values. The results are identical, but self.assertDictEqual(obs, filled_surveys) fails because elements are in different internal order. Will look into and resolve after the call.

charles-cowart:
> If we add the survey_template_id to ag_login_surveys, and leave all of the old survey templates intact (but retired), why do we need to do any sort of migration?

We aren't migrating data, but we do need to copy it. For survey x of template 'Primary Questionnaire', we need to take those 100 odd questions, filter out the 'Undetermined' non-answers, create new templates with the right subset of questions in each, padding out those templates with their own 'Undetermined' answers for the questions that were not answered, and create them with survey X's timestamp, source id, sample_id, account_id, etc.

charles-cowart:
Closed #466 as this PR supersedes it. Specifically, if we remove the component that migrates old surveys into new templates, what is left is code that’s basically a better version of the code written in #466 to populate surveys with legacy answers, with better tests. Since they both touch read_survey_template() in the same way, it’s really one or the other. The only thing left to keep from #466 was the percentage completed code, so I folded it in. I also added additional tests for the Qiita-related functionality as requested in #466.

microsetta_private_api/admin/tests/test_admin_api.py
```diff
@@ -815,60 +816,218 @@ def test_metadata_qiita_compatible_invalid(self):
...
-        self.assertIn('about_yourself_text', obs)
+        self.assertEqual(result['000069747']['birth_year'], '1984')
 
-    def test_metadata_qiita_compatible_valid_no_private(self):
-        data = json.dumps({'sample_barcodes': ['000004216', '000004213']})
+        # clean up by deleting the survey we added for testing.
+        with Transaction() as t:
+            sar = SurveyAnswersRepo(t)
+            sar.delete_answered_survey(account_id, survey_id)
+
```
charles-cowart, line 901: add commit here as well.

microsetta_private_api/admin/tests/test_admin_api.py
```diff
@@ -815,60 +816,218 @@ def test_metadata_qiita_compatible_invalid(self):
...
+    def test_metadata_qiita_compatible_valid_private(self):
+        # All the old PM_ values have been retired. We've migrated
+        # 502: GENDER_v2 to PM_GENDER_v2. 502 is in template 10. Since there
+        # aren't existing template 10 surveys, create one, set 502, and confirm
+        # it appears in query results.
```
charles-cowart, line 906: We will want to keep GENDER_v2 as-is and push it to Qiita. It's not private an (BImore.
We have no plans for any private short names at this time? - privacy should not be based on PM_ and instead should be based on EBI submissions. see: EBI_REMOVE (EBI_REMOVE alone is not ground truth for privacy. Any free-text answer is also private)
About-yourself (116) needs to be unretired and stored in template 10.

microsetta_private_api/admin/tests/test_admin_api.py
```diff
@@ -815,60 +816,218 @@ def test_metadata_qiita_compatible_invalid(self):
...
+                                    '?include_private=True',
                                     content_type='application/json',
                                     data=data,
                                     headers=MOCK_HEADERS)
+
+        # clean up by deleting the survey we added for testing.
+        with Transaction() as t:
+            sar = SurveyAnswersRepo(t)
+            sar.delete_answered_survey(account_id, survey_id)
+
```
charles-cowart, line 960: Add commit here too.

microsetta_private_api/api/tests/test_api.py
```diff
@@ -1817,9 +1860,9 @@ def test_edit_sample_locked(self):
         self.assertEqual(400, post_resp.status_code)
 
         # if sample date is greater than 30 days
+        # For most two month spans, the total number of days is 61.
         now = datetime.datetime.now()
-        delta = relativedelta(months=now.month+2)
-        date = now+delta
+        date = now + datetime.timedelta(days=61)
```
charles-cowart, line 1865: Lost Cassidy's fix.

microsetta_private_api/api/tests/test_integration.py
```diff
@@ -1600,11 +1637,11 @@ def test_survey_localization(self):
...
             result = repo.get_answered_survey(ACCT_ID, HUMAN_ID,
                                               survey_id, 'en_US')
-            self.assertEqual(result['107'], 'Male',
+            self.assertEqual(result['111'], 'March',
                              "Couldn't read answer from db in en_US")
             result = repo.get_answered_survey(ACCT_ID, HUMAN_ID,
                                               survey_id, 'en_GB')
-            self.assertEqual(result['107'], 'Wizard',
+            self.assertEqual(result['111'], 'March',
                              "Couldn't read answer from db in en_GB")
```
charles-cowart, line 1645: Consider something that's actually different

microsetta_private_api/repo/metadata_repo/_repo.py
```diff
@@ -233,9 +235,24 @@ def _to_pandas_dataframe(metadatas, survey_templates):
     transformed = []
 
     multiselect_map = _construct_multiselect_map(survey_templates)
+
     for metadata in metadatas:
+        # metadata is a dict representing a barcode's metadata.
         try:
+            # each as_series represents the aggregated shortname, value
```
charles-cowart, line 242: Any code interacting w/Qiita should return the value from the survey with the smallest time delta to the timestamp associated with the sample (see the timestamp in ag_login_surveys). With respect to the UI, we want the latest answer.

microsetta_private_api/repo/metadata_repo/_repo.py
```diff
@@ -383,20 +402,19 @@ def _to_pandas_series(metadata, multiselect_map):
...
 
         if template in collected:
+            # HACK: there exist some samples that have duplicate surveys. This
+            # is unusual and unexpected state in the database, and has so far
+            # only been observed only with the surfers survey. The hacky
+            # solution is to only gather the results once
             continue
-        else:
-            collected.add(template)
+
```
charles-cowart, line 416: explain that behavior has changed and that this is no longer a hack but a feature. however for legacy data we don't know how to resolve it perhaps? because we don't have the collected timestamps. Are there timestamps associated with all surveys?

microsetta_private_api/repo/survey_answers_repo.py
```diff
@@ -207,7 +206,8 @@ def get_answered_survey(self, ag_login_id, source_id,
 
     def submit_answered_survey(self, ag_login_id, source_id,
                                language_tag, survey_template_id, survey_model,
-                               survey_answers_id=None):
+                               survey_answers_id=None,
+                               creation_time=None):
```
charles-cowart, line 210: remove creation_time as a parameter.

microsetta_private_api/repo/survey_answers_repo.py
```diff
@@ -230,12 +230,21 @@ def submit_answered_survey(self, ag_login_id, source_id,
...
 
         with self._transaction.cursor() as cur:
-
             # Log that the user submitted this survey
-            cur.execute("INSERT INTO ag_login_surveys "
-                        "(ag_login_id, survey_id, source_id) "
-                        "VALUES(%s, %s, %s)",
-                        (ag_login_id, survey_answers_id, source_id))
+            if creation_time:
+                cur.execute("INSERT INTO ag_login_surveys "
```
charles-cowart, line 235: this can be adjusted as well since we are removing creation_time

microsetta_private_api/repo/admin_repo.py
```diff
@@ -1156,60 +1164,64 @@ def get_survey_metadata(self, sample_barcode, survey_template_id=None):
...
+        template_ids = list(set([x[1] for x in template_ids]))
 
         all_survey_answers = []
-        for answer_id in answer_ids:
-            answer_model = survey_answers_repo.get_answered_survey(
-                account_id,
-                source_id,
-                answer_id,
-                "en_US"
-            )
```
charles-cowart, line 1197: Is this code vestigial?

microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -929,3 +1026,354 @@ def has_external_surveys(self, account_id, source_id):
...
+    def migrate_responses_by_barcode(self, barcode, survey_template_id):
+        '''
+        Get all survey responses associated with a barcode and
+         migrate them. Will pull from past results if they are found,
+         including legacy template answers. Only latest results are returned.
+        :param account_id: An account
+        :param barcodes: a barcode to search by
+        :return: A filled survey_template dict
+        '''
+        if int(survey_template_id) in self.SURVEY_INFO:
```
wasade, line 1039: Please ensure `survey_template_id` is the correct type before calling this method

microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -929,3 +1026,354 @@ def has_external_surveys(self, account_id, source_id):
                     return True
 
         return False
+
+    def migrate_responses_by_barcode(self, barcode, survey_template_id):
+        '''
+        Get all survey responses associated with a barcode and
+         migrate them. Will pull from past results if they are found,
+         including legacy template answers. Only latest results are returned.
+        :param account_id: An account
```
wasade, line 1035: This isn't a parameter for the method

microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -929,3 +1026,354 @@ def has_external_surveys(self, account_id, source_id):
...
+        with self._transaction.cursor() as cur:
+            # first, confirm account_id is valid.
+            '''
+            cur.execute("""SELECT ag_login_id AS account_id
+                           FROM   ag.ag_login_surveys
+                           WHERE  ag_login_id = %s""", (account_id,))
+            cur.fetchall()
+            '''
+
+            if cur.rowcount == 0:
```
wasade, line 1107: it looks like this conditional is based on a stale variable as the `cur` call is commented out

wasade: ...rather unused `cur`

microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -929,3 +1026,354 @@ def has_external_surveys(self, account_id, source_id):
...
+
+            rows = cur.fetchall()
+
+            # create an empty template to fill-in.
+            results = self._generate_empty_survey(survey_template_id)
+
+            for row in rows:
+                question_id = str(row[0])
+                response = row[1]
+                response_type = row[3]
```
wasade, line 1121: are these indices assured to be stable given that a `SELECT *` is used? What about a dict cursor or enforcing the order by specifying the columns in the select?

microsetta_private_api/repo/tests/test_survey_template_repo.py
```diff
@@ -504,3 +510,640 @@ def test_has_external_surveys(self):
...
+                sr.create_source(HUMAN_SOURCE)
+            except UniqueViolation:
+                pass
+
+            t.commit()
+
+            return HUMAN_SOURCE
+
+    def _clean_up(self, account_id, human_source, survey_ids):
+        with Transaction() as t:
```
charles-cowart, line 570: potentially merge these back into single test method

microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -929,3 +1026,354 @@ def has_external_surveys(self, account_id, source_id):
...
+                else:
+                    # STRING and TEXT responses may need their enclosing
+                    # brackets removed. For now, pass them on as is:
+                    # e.g.: ["Free text - <encrypted text>"]
+                    results[question_id] = response
+
+            return results
+
+    def migrate_responses(self, account_id, survey_template_id):
+        '''
```
charles-cowart, line 1145: Make sure we distinguish between different sources in an account_id. Let's not use account_id and use only source_id so the answers are user-associated and not account associated.

microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -929,3 +1026,354 @@ def has_external_surveys(self, account_id, source_id):
...
+                    results[question_id] = response
+                else:
+                    # STRING and TEXT responses may need their enclosing
+                    # brackets removed. For now, pass them on as is:
+                    # e.g.: ["Free text - <encrypted text>"]
+                    results[question_id] = response
+
+            return results
+
+    def migrate_responses(self, account_id, survey_template_id):
```
wasade, line 1144: This should be by source not account

microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -929,3 +1026,354 @@ def has_external_surveys(self, account_id, source_id):
...
+        with self._transaction.cursor() as cur:
+            # first, confirm account_id is valid.
+            '''
+            cur.execute("""SELECT ag_login_id AS account_id
+                           FROM   ag.ag_login_surveys
+                           WHERE  ag_login_id = %s""", (account_id,))
+            cur.fetchall()
+            '''
+
+            if cur.rowcount == 0:
```
wasade, line 1217: Same comment here with the unused `cur`

microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -929,3 +1026,354 @@ def has_external_surveys(self, account_id, source_id):
...
+            rows = cur.fetchall()
+
+            # create an empty template to fill-in.
+            results = self._generate_empty_survey(survey_template_id)
+            total_question_count = len(results)
+
+            for row in rows:
+                question_id = str(row[0])
+                response = row[1]
+                response_type = row[4]
```
wasade, line 1232: Same comment here regarding magic values

microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -929,3 +1026,354 @@ def has_external_surveys(self, account_id, source_id):
...
+            return results
+
+    def get_template_ids_from_survey_ids(self, survey_ids):
+        '''
+        returns a list of (template_id, survey_id) tuples.
+        :param survey_ids: A list of survey ids.
+        :return: A list of (survey_id, survey_template_id) tuples.
+        '''
+        if not isinstance(survey_ids, list):
+            raise ValueError("survey_ids is not a list of survey ids")
```
wasade, line 1312: Please have the caller of this method ensure the data are of the correct type

microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -929,3 +1026,354 @@ def has_external_surveys(self, account_id, source_id):
...
+
+        if len(survey_ids) == 0 or '' in survey_ids or None in survey_ids:
```
wasade, line 1314: Please handle this upstream

microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -929,3 +1026,354 @@ def has_external_surveys(self, account_id, source_id):
...
+    def _remote_survey_template_id_from_survey_id(self, survey_id):
+        # VIOscreen
+        vioscreen_repo = VioscreenRepo(self._transaction)
+        status = vioscreen_repo._get_vioscreen_status(survey_id)
+        if status is not None:
+            return [(survey_id, SurveyTemplateRepo.VIOSCREEN_ID)]
+
+        with self._transaction.cursor() as cur:
+            try:
+                # If survey_id isn't a valid UUID, a ValueError will be
```
wasade, line 1336: There are valid survey IDs which are not UUID

microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -929,3 +1026,354 @@ def has_external_surveys(self, account_id, source_id):
...
+        res = self._local_survey_template_ids_from_survey_ids(survey_ids)
+
+        # check to see if any of the survey ids are associated with a remote
+        # survey.
+        for survey_id in survey_ids:
+            res += self._remote_survey_template_id_from_survey_id(survey_id)
+
+        return res
+
+    def _remote_survey_template_id_from_survey_id(self, survey_id):
```
wasade, line 1327: I'm not sure I undersand this function. Don't we know what template a survey ID is associated with and therefore don't need to test its existence in multiple registries?

microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -929,3 +1026,354 @@ def has_external_surveys(self, account_id, source_id):
...
+                        (survey_id,))
+
+            if cur.fetchone()[0] is True:
+                return [(survey_id, SurveyTemplateRepo.SPAIN_FFQ_ID)]
+
+        return []
+
+    def _local_survey_template_ids_from_survey_ids(self, survey_ids):
+        with self._transaction.cursor() as cur:
+            # note these ids are unique string ids, not integer ids.
```
wasade, line 1374: The survey ID specifically right?

microsetta_private_api/repo/tests/test_survey_template_repo.py
```diff
@@ -504,3 +510,640 @@ def test_has_external_surveys(self):
...
+            with self.assertRaises(ValueError):
+                sar._generate_empty_survey(0)
+            with self.assertRaises(ValueError):
+                sar._generate_empty_survey(-1)
+            with self.assertRaises(ValueError):
+                sar._generate_empty_survey(100)
+
+            # handle remote template id
+            with self.assertRaises(ValueError):
+                sar._generate_empty_survey(10001)
```
wasade, line 536: Can the static variable for this survey be used?

microsetta_private_api/repo/tests/test_survey_template_repo.py
```diff
@@ -504,3 +510,640 @@ def test_has_external_surveys(self):
...
+            HUMAN_INFO.assent_obtainer = None
+            HUMAN_INFO.age_range = '18-plus'
+
+            HUMAN_SOURCE = Source('ffffffff-ffff-ffff-aaaa-aaaaaaaaaaaa',
+                                  account_id,
+                                  Source.SOURCE_TYPE_HUMAN,
+                                  'test person',
+                                  HUMAN_INFO)
+
+            try:
```
wasade, line 560:
```suggestion
```

microsetta_private_api/repo/tests/test_survey_template_repo.py
```diff
@@ -504,3 +510,640 @@ def test_has_external_surveys(self):
...
+                sr.create_source(HUMAN_SOURCE)
+            except UniqueViolation:
+                pass
```
wasade, line 563:
```suggestion
```

microsetta_private_api/repo/tests/test_survey_template_repo.py
```diff
@@ -504,3 +510,640 @@ def test_has_external_surveys(self):
...
+            HUMAN_INFO.age_range = '18-plus'
+
+            HUMAN_SOURCE = Source('ffffffff-ffff-ffff-aaaa-aaaaaaaaaaaa',
+                                  account_id,
+                                  Source.SOURCE_TYPE_HUMAN,
+                                  'test person',
+                                  HUMAN_INFO)
+
+            try:
+                sr.create_source(HUMAN_SOURCE)
```
wasade, line 561:
```suggestion
            sr.create_source(HUMAN_SOURCE)
```

microsetta_private_api/repo/tests/test_survey_template_repo.py
wasade: Please ensure the test database is cleaned up

microsetta_private_api/repo/tests/test_survey_template_repo.py
```diff
@@ -504,3 +510,640 @@ def test_has_external_surveys(self):
...
+                                                   human_source.id,
+                                                   'en_US',
+                                                   10,
+                                                   survey_10)
+
+            t.commit()
+
+            return survey_id
+
+    def test_migrate_responses(self):
```
wasade, line 607: I'm not reviewing this test as we had discussed making this function per source not per account

wasade (changes requested):

microsetta_private_api/api/tests/test_integration.py
charles-cowart: Note: Low priority.

microsetta_private_api/db/migration_support.py
charles-cowart: Removed.

microsetta_private_api/admin/tests/test_admin_api.py
charles-cowart: Resolved.

microsetta_private_api/admin/tests/test_admin_api.py
charles-cowart: Resolved.

microsetta_private_api/repo/survey_answers_repo.py
charles-cowart: Resolved.

microsetta_private_api/repo/survey_answers_repo.py
charles-cowart: Resolved.

microsetta_private_api/repo/metadata_repo/_repo.py
```diff
@@ -14,12 +14,14 @@
...
-
 # the vioscreen survey currently cannot be fetched from the database
 # there seems to be some detached survey IDs -- see 000089779
 # that account has a long and unusual history though
 # Adding the MyFoodRepo, Polyphenol FFQ, and Spain FFQs to the
 # ignore list.
+# do not add legacy template_ids (1-7) to this dict at this time.
+# there are users of this dict that process valid grabs on legacy template
+# dagta.
 TEMPLATES_TO_IGNORE = {10001, 10002, 10003, 10004, None}
```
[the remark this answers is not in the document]
charles-cowart: Resolved.

microsetta_private_api/repo/survey_template_repo.py
charles-cowart: Resolved.

microsetta_private_api/repo/survey_template_repo.py
charles-cowart: Well the SELECT * is just a thin wrapper over the real query to allow ordering by creation time and question id. The UNION of the two queries inside is what explicitly defines the columns in the result. However I'll keep it in mind. This query will need to be reworked anyway to allow for closest time delta to the survey timestamp.

microsetta_private_api/api/tests/test_api.py
charles-cowart: Resolved.

microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -329,9 +420,10 @@ def set_myfoodrepo_id(self, account_id, source_id, mfr_id):
                         "ag_login_id, "
                         "survey_id, "
                         "vioscreen_status, "
-                        "source_id) "
-                        "VALUES(%s, %s, %s, %s)",
-                        (account_id, mfr_id, None, source_id))
+                        "source_id, "
+                        "survey_template_id) "
+                        "VALUES(%s, %s, %s, %s, %s)",
+                        (account_id, mfr_id, None, source_id, 10002))
```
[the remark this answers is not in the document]
charles-cowart: Resolved

microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -560,9 +654,11 @@ def create_spain_ffq_entry(self, account_id, source_id):
...
                             "survey_id, "
                             "vioscreen_status, "
-                            "source_id) "
-                            "VALUES(%s, %s, %s, %s)",
-                            (account_id, spain_ffq_id, None, source_id))
+                            "source_id, "
+                            "survey_template_id) "
+                            "VALUES(%s, %s, %s, %s, %s)",
+                            (account_id, spain_ffq_id, None, source_id,
+                             10004))
```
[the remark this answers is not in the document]
charles-cowart: Resolved

microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -655,9 +751,10 @@ def create_vioscreen_id(self, account_id, source_id,
                             "ag_login_id, "
                             "survey_id, "
                             "vioscreen_status, "
-                            "source_id) "
-                            "VALUES(%s, %s, %s, %s)",
-                            (account_id, vioscreen_id, -1, source_id))
+                            "source_id, "
+                            "survey_template_id) "
+                            "VALUES(%s, %s, %s, %s, %s)",
+                            (account_id, vioscreen_id, -1, source_id, 10001))
```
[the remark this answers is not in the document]
charles-cowart: Resolved

microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -465,9 +557,11 @@ def create_polyphenol_ffq_entry(self, account_id, source_id,
...
                             "survey_id, "
                             "vioscreen_status, "
-                            "source_id) "
-                            "VALUES(%s, %s, %s, %s)",
-                            (account_id, polyphenol_ffq_id, None, source_id))
+                            "source_id, "
+                            "survey_template_id) "
+                            "VALUES(%s, %s, %s, %s, %s)",
+                            (account_id, polyphenol_ffq_id, None, source_id,
+                             10003))
```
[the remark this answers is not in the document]
charles-cowart: Resolved

microsetta_private_api/repo/survey_template_repo.py
charles-cowart: Removed. The only caller for this method by its nature only passes valid template ids of local surveys.

microsetta_private_api/api/tests/test_integration.py
charles-cowart: removing from consideration for this PR.

microsetta_private_api/repo/admin_repo.py
```diff
@@ -1156,60 +1164,64 @@ def get_survey_metadata(self, sample_barcode, survey_template_id=None):
...
+
+            # filter out survey_id/survey_template_id pairs that aren't the
+            # optional survey_template_id.
+            template_ids = [x for x in template_ids
+                            if x[1] == survey_template_id]
+
+        metadata_map = survey_ans_repo.build_metadata_map()
+
+        # we now just need a set of template_ids
+        template_ids = list(set([x[1] for x in template_ids]))
```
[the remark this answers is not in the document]
charles-cowart: template_ids in this case is actually a list of (survey_id, survey_template_id) tuples. In the case where we just want a list of template_ids, we need the list comprehension to extract that out. The set is there to remove potential duplicate template ids since there might now be more than one survey of template type N.

microsetta_private_api/repo/tests/test_survey_template_repo.py
charles-cowart: Resolved/

microsetta_private_api/repo/metadata_repo/_repo.py
charles-cowart: Resolved.

microsetta_private_api/repo/tests/test_survey_template_repo.py
charles-cowart: This is used in a couple of different tests. Will leave as is.

microsetta_private_api/repo/survey_template_repo.py
charles-cowart: Resolved.

microsetta_private_api/repo/survey_template_repo.py
charles-cowart: Resolved.

microsetta_private_api/repo/admin_repo.py
```diff
@@ -1156,60 +1164,64 @@ def get_survey_metadata(self, sample_barcode, survey_template_id=None):
...
-            # TODO: This schema is so awkward for this type of query...
-            answers = []
-            for answer_id in answer_ids:
-                if answer_to_template_map[answer_id][0] == survey_template_id:
-                    answers.append(answer_id)
+        st_repo = SurveyTemplateRepo(self._transaction)
 
-            if len(answers) == 0:
+        template_ids = st_repo.get_template_ids_from_survey_ids(a_ids)
+
```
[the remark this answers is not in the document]
charles-cowart: template_ids as a list of tuples is used further below. It wouldn't be good to strip the survey_ids out at this time.

microsetta_private_api/repo/admin_repo.py
```diff
@@ -1156,60 +1164,64 @@ def get_survey_metadata(self, sample_barcode, survey_template_id=None):
...
+        if survey_template_id is not None:
+            # extract just the template_ids for this check.
+            if survey_template_id not in [x[1] for x in template_ids]:
```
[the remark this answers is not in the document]
charles-cowart: Related to above issue. template_ids is needed as a tuple below this line.

microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -1029,11 +1029,56 @@ def has_external_surveys(self, account_id, source_id):
...
+            # to the time the survey was taken.
+            cur.execute("""SELECT b.creation_time
+                                   FROM   ag.source_barcodes_surveys a
+                                          JOIN ag.ag_login_surveys b
+                                            ON a.survey_id = b.survey_id
+                                   WHERE  a.barcode = %s
+                                          AND b.survey_template_id = %s""",
+                        (barcode, survey_template_id,))
+
+            # what if there are multiple template Xs? Should we take the
```
cassidysymons, line 1047: For anything historical, there won't be more than one. Per the way surveys are associated with a source and [this comment](https://github.com/biocore/microsetta-interface/blob/master/microsetta_interface/implementation.py#L1566) explaining the current thought-process, you can safely count on one record.

This will need to be re-thought when we eventually change the metadata process to update, rather than just insert, but it's beyond the scope of this PR.

microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -929,3 +1026,354 @@ def has_external_surveys(self, account_id, source_id):
...
+                                 AND g.retired = false) t
+                 ORDER  BY creation_time ASC,
+                           survey_question_id ASC"""
+
+        with self._transaction.cursor() as cur:
+            # first, confirm account_id is valid.
+            '''
+            cur.execute("""SELECT ag_login_id AS account_id
+                           FROM   ag.ag_login_surveys
+                           WHERE  ag_login_id = %s""", (account_id,))
```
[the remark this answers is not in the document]
charles-cowart: Resolved. I believe this code was part of a commented out check that was removed.

microsetta_private_api/repo/survey_template_repo.py
charles-cowart: Resolved. I believe this code was part of a commented out check that was removed.

microsetta_private_api/repo/survey_template_repo.py
charles-cowart: Resolved. I believe this code was part of a commented out check that was removed.

microsetta_private_api/repo/survey_template_repo.py
charles-cowart: Similar as the previous example. The SELECT * is really just a wrapper for the UNIONed queries inside, which explicitly define the parameters and their order. The enclosing SELECT is just to order them by creation time and survey question id.

microsetta_private_api/repo/survey_template_repo.py
charles-cowart: Removed. Examined the callers and can pretty safely say only the correct types will be sent.

microsetta_private_api/repo/survey_template_repo.py
charles-cowart: Removed. Examined the callers and can pretty safely say only the correct types will be sent.

microsetta_private_api/repo/survey_template_repo.py
charles-cowart: Resolved.

microsetta_private_api/repo/survey_template_repo.py
charles-cowart: Resolved. These were reworked from legacy functions before we included the survey_template_id column in ag_login_surveys table. Both of these were removed and the calling method just makes a simple query.

microsetta_private_api/repo/survey_template_repo.py
charles-cowart: Resolved. These were reworked from legacy functions before we included the survey_template_id column in ag_login_surveys table. Both of these were removed and the calling method just makes a simple query.

microsetta_private_api/repo/tests/test_survey_template_repo.py
charles-cowart: Resolved

microsetta_private_api/repo/tests/test_survey_template_repo.py
charles-cowart: Resolved.

charles-cowart:
@cassidysymons @wasade ostensibly finished! All comments addressed.

wasade:
Thanks, reviewing in a moment

microsetta_private_api/admin/tests/test_admin_api.py
```diff
@@ -815,60 +816,242 @@ def test_metadata_qiita_compatible_invalid(self):
...
         self.assertEqual(200, response.status_code)
         result = json.loads(response.data)
-        self.assertEqual(set(result.keys()), {'000004216', '000004213'})
-        obs = {c.lower() for c in result['000004216']}
-        self.assertNotIn('about_yourself_text', obs)
+        for barcode in result:
+            print(barcode)
+            foo = result[barcode]
+            for bar in foo:
+                print("\t%s" % bar)
```
wasade, line 981:
```suggestion
```

microsetta_private_api/admin/tests/test_admin_api.py
```diff
@@ -815,60 +816,242 @@ def test_metadata_qiita_compatible_invalid(self):
...
```
wasade, line 1051:
```suggestion
```

microsetta_private_api/admin/tests/test_admin_api.py
```diff
@@ -815,60 +816,242 @@ def test_metadata_qiita_compatible_invalid(self):
...
+        self.assertEqual(set(result.keys()), {'000069747'})
+        obs = {c.lower() for c in result['000069747']}
+        self.assertNotIn('birth_month', obs)
```
wasade, line 1055: Can a test be added for when a sample's source has multiple instances of the same survey template, and which demonstrates that the sample is correctly associated to right survey?

wasade: Or is this done elsewhere?

microsetta_private_api/api/tests/test_api.py
```diff
@@ -1780,7 +1820,9 @@ def test_edit_sample_locked(self):
         create_dummy_kit(dummy_acct_id, dummy_source_id,
                          associate_sample=True)
         _ = create_dummy_answered_survey(
-            dummy_acct_id, dummy_source_id)
+            dummy_acct_id,
+            dummy_source_id,
+            survey_template_id=st_repo.DIET_ID,)
```
wasade, line 1825:
```suggestion
            survey_template_id=st_repo.DIET_ID)
```

microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -929,3 +1025,313 @@ def has_external_surveys(self, account_id, source_id):
...
         return False
+
+    def _get_timestamp(self, barcode, survey_template_id):
+        with self._transaction.cursor() as cur:
+            # retrieve the timestamp for the survey being asked for. When we
+            # dynamically generate our results, we will incorporate answers
+            # from other surveys, but we will prioritize responses (or set of
+            # responses in the case of questions w/multiple answers) closest
+            # to the time the survey was taken.
+            cur.execute("""SELECT b.creation_time
```
wasade, line 1036: This will return multiple times if the user has taken the same survey multiple times. A single value is pulled out on like 1047, and I believe that will be an arbitrary time stamp. This function appears to be used later for determining what metadata to associate to a sample, so it seems that using an arbitrary timestamp is incorrect.

microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -929,3 +1025,313 @@ def has_external_surveys(self, account_id, source_id):
...
+                        (barcode, survey_template_id,))
+
+            # what if there are multiple template Xs? Should we take the
+            # latest one? or take the one closest to the source creation
+            # timestamp?
+            result = cur.fetchone()
+            if result is not None:
+                return result[0]
+
+            # if a timestamp could not be found for the survey template id in
```
wasade, line 1051: Does this scenario actually occur?

```sql
[local]:5432 dtmcdonald@ag_test=# select * from ag.ag_login_surveys where creation_time is null;
+-------------+-----------+------------------+-----------+---------------+
| ag_login_id | survey_id | vioscreen_status | source_id | creation_time |
+-------------+-----------+------------------+-----------+---------------+
+-------------+-----------+------------------+-----------+---------------+
(0 rows)

```

microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -929,3 +1025,313 @@ def has_external_surveys(self, account_id, source_id):
...
+                                    ON a.survey_id = b.survey_id
+                                  JOIN ag.source c
+                                    ON b.source_id = c.id
+                           WHERE  a.barcode = %s""", (barcode,))
+
+            result = cur.fetchone()
+            if result is not None:
+                return result[1]
+
+            raise ValueError("A timestamp could not be found for barcode "
```
wasade, line 1068: Let's issue an `ALTER TABLE` on `ag_login_surveys` and enforce that `creation_time` is `NOT NULL`.

microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -929,3 +1025,313 @@ def has_external_surveys(self, account_id, source_id):
...
+        with self._transaction.cursor() as cur:
+            cur.execute(sql, (barcode, survey_template_id,
+                              barcode, survey_template_id,))
+
+            rows = cur.fetchall()
+
+            # create an empty template to fill-in.
+            results = self._generate_empty_survey(survey_template_id)
+
+            # retrieve the timestamp used to prefer one response over another.
```
wasade, line 1141: The called function does not determine what the most temporally similar survey timestamp is to the barcode timestamp.

microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -929,3 +1025,313 @@ def has_external_surveys(self, account_id, source_id):
...
+            target_ts = self._get_timestamp(barcode, survey_template_id)
+
+            # find the timestamp for the response (single, text, string) or
+            # set of responses (multiple) that has the timestamp closest to
+            # the target timestamp target_ts.
+            best_ts = {}
+            for row in rows:
```
wasade, line 1148: Can the tuple be expanded to avoid magic numbers?

microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -929,3 +1025,313 @@ def has_external_surveys(self, account_id, source_id):
...
+                        results[question_id].remove("Unspecified")
+                    except ValueError:
+                        pass
+                elif response_type == 'SINGLE':
+                    results[question_id] = response
+                else:
+                    # STRING and TEXT responses may need their enclosing
+                    # brackets removed. For now, pass them on as is:
+                    # e.g.: ["Free text - <encrypted text>"]
+                    results[question_id] = response
```
wasade, line 1269: I believe this complex logic is done twice, can it be decomposed?

microsetta_private_api/repo/tests/test_survey_template_repo.py
```diff
@@ -504,3 +509,683 @@ def test_has_external_surveys(self):
...
+        with Transaction() as t:
+            str = SurveyTemplateRepo(t)
+            # use data from an old template 1 survey and return a subset of it
+            # in a generated template 10.
+
+            obs, obs_pc = str.migrate_responses(source_id, 10)
+            self.assertDictEqual(obs, self.filled_survey_10a)
+            self.assertAlmostEqual(obs_pc, 0.615, 2)
+
+            # disabled because some other tests rely on an invalid id.
```
wasade, line 624: I'm not sure I understand, is it correct that this test is commented out?

microsetta_private_api/repo/tests/test_survey_template_repo.py
```diff
@@ -504,3 +509,683 @@ def test_has_external_surveys(self):
...
+            # values to the result.
+            obs = rpo.migrate_responses_by_barcode('000001410', 19)
+            self.assertDictEqual(obs, self.filled_survey_19a)
+
+            # the following statements create a new source and submits a
+            # survey. It then submits a survey to the existing source
+            # associated with barcode '000001410'. Neither survey should
+            # affect the results for the final query.
+            print("ACCOUNT ID: %s" % account_id)
+            print("SOURCE ID: %s" % source_id)
```
wasade, line 702:
```suggestion
```

wasade (changes requested):
Thanks Charlie. So I don't see where barcodes are correctly associated with their most temporally similar survey, and I don't see an example of producing correct metadata for a sample when a source has taken the same survey multiple times.

microsetta_private_api/admin/tests/test_admin_api.py
charles-cowart: ty for the catch!

charles-cowart:
> Thanks Charlie. So I don't see where barcodes are correctly associated with their most temporally similar survey...

Hi @wasade Thanks for the review! I have a method called _get_timestamp() that looks for a survey that is an exact match to a barcode and a survey template id. If it finds one, it'll return that to the code I've highlighted below. If for some reason it can't find a matching survey, it'll assume every source has a creation timestamp and return that instead. It's in a separate method so we could easily tweak it based on your expected feedback.

The code below will take the set of all responses across all times and return them to the caller. (Not sure if that's what we really wanted because some answers change over time and some don't.) Where there are multiple responses to the same question, it will select the response closest to the timestamp returned by _get_timestamp():

https://github.com/charles-cowart/microsetta-private-api/blob/eda55696ea443ff5edd21dcd6e4a93260f99536e/microsetta_private_api/repo/survey_template_repo.py#L1141-L1154.

microsetta_private_api/repo/tests/test_survey_template_repo.py
charles-cowart: That is correct. I added a check to migrate_responses() to ensure the source_id is a valid one, but it causes a few of the other tests to fail because they use a mocked source_id ('aaaa-fffff...'). I commented it out but left it here for discussion. I didn't want to try to mock the check for the other tests at that point. I think actually it can be removed because we assume other parameters to functions and methods like these are supplying only valid parameters.

microsetta_private_api/repo/survey_template_repo.py
charles-cowart: Sure I'll take a look.

microsetta_private_api/repo/survey_template_repo.py
charles-cowart: Ah I see what you're saying - no problem, that should work...

microsetta_private_api/repo/survey_template_repo.py
charles-cowart: Ah, gotcha - there is a timestamp for the barcode - it's in barcodes.barcode. You mean that one right?

microsetta_private_api/repo/survey_template_repo.py
charles-cowart: Not in the test data, but I'm not sure what staging and production look like so I was coding a little defensively.

cassidysymons: The query @wasade ran was on Staging, which contains a snapshot of our production database. If we add the NOT NULL constraint as suggested, we can safely assume that the scenario won't occur.

microsetta_private_api/repo/survey_template_repo.py
charles-cowart: Ty, yeah I address that in the comment beginning on L1044. I figured I would get reviewer input on that.

microsetta_private_api/admin/tests/test_admin_api.py
charles-cowart: It's definitely a case I considered when writing the tests for it, so there isn't a specific test for it. I'll go through the modified and new tests and see if I can pick out the examples of doing this and see if you think it tests it well enough. I have a feeling some of those tests will need to be modified after looking at the other comments anyway.

microsetta_private_api/repo/survey_template_repo.py
charles-cowart: That sounds great thanks!

microsetta_private_api/repo/survey_template_repo.py
wasade: No, the timestamp is `ag.ag_kit_barcodes.sample_{date,time}`

microsetta_private_api/repo/survey_template_repo.py
wasade: I actually ran it locally, but just verified it on staging

wasade:
Thanks, @charles-cowart. `_get_timestamp` does not interact with `ag_kit_barcodes` as far as I can tell, so it is not obtaining the sample's timestamp

microsetta_private_api/repo/survey_template_repo.py
charles-cowart: Sounds good! Resolved.

microsetta_private_api/repo/survey_template_repo.py
charles-cowart: @wasade It looks like these columns are mostly NULL for local data at least, is it okay to assume that for our use case at least, we'll always have one? Maybe this is the timestamp you were referring to when you said there was some legacy ids w/out timestamps?

wasade: NULL for unassigned samples, not null for samples assocaited with a source

microsetta_private_api/repo/survey_template_repo.py
charles-cowart: Resolved.

microsetta_private_api/repo/tests/test_survey_template_repo.py
charles-cowart: Removed.

microsetta_private_api/repo/survey_template_repo.py
charles-cowart: Resolved.

microsetta_private_api/repo/survey_template_repo.py
charles-cowart: tentatively resolved.

charles-cowart: Updated to remove survey_template_ids as part of the query. It's not really needed since all timestamps would be the same in any case, and even if there were multiple surveys of the type requested or none, we'd want to center our responses around the barcode's timestamp.

microsetta_private_api/repo/survey_template_repo.py
charles-cowart: Resolved. Replaced with a new query based on barcode timestamp found in ag.ag_kit_barcodes.

microsetta_private_api/repo/survey_template_repo.py
charles-cowart: Resolved. Replaced with a new query based on barcode timestamp found in ag.ag_kit_barcodes.

microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -1028,45 +1028,30 @@ def has_external_surveys(self, account_id, source_id):
...
-
-            raise ValueError("A timestamp could not be found for barcode "
-                             f"{barcode}")
+                                  JOIN ag.ag_kit_barcodes c
+                                    ON a.barcode = c.barcode
+                           WHERE  a.barcode = %s
+                                  AND b.survey_template_id = %s""",
+                        (barcode, survey_template_id))
+
+            # Expect that one and only one result will always be found.
```
cassidysymons: Given that we're changing the system to allow people to re-take surveys, that's not a safe assumption moving forward. I know my earlier comment reflected that it was a safe assumption for old data, but we either need this function to account for multiple records, or one of the upstream functions to do so.

charles-cowart:
I resolved the merge conflict with the idea that I didn't change anything substantial to the vioscreen-related method and hence the changes in master-overhaul superseded them.

charles-cowart:
Reviewing and resolving test errors post merge.

microsetta_private_api/admin/tests/test_admin_api.py
charles-cowart: I would say this case is covered in test_migrate_responses_by_barcode():
https://github.com/biocore/microsetta-private-api/blob/0ebec2c0470aa9189c7e774ce1b54d0f37f1bdcf/microsetta_private_api/repo/tests/test_survey_template_repo.py#L697-L720

We have an existing filled survey of template type 1 and then submit two filled surveys of template type 10. Each survey has their own answer for '111/Birth Month'. The correct answer for retrieving by barcode would be to return the earliest value from template 1 (the one closest to the barcode's timestamp) (June) instead of the later values from the template 10s (February, March).

I could piggy-back a template 10-only value - probably GENDER_v2/502 since it's also a SINGLE, but I don't think it's needed. The first submission could submit "Male" and the second submission could submit "Female" and we would test to see that "Male" was returned (since it would be from a survey closer in time to the barcode timestamp than the one after) (or vice-versa).

microsetta_private_api/repo/survey_template_repo.py
charles-cowart: I think this is okay now, because the query has been changed per Daniel's request to use the sample_date and sample_time columns from ag.ag_kit_barcodes. In this case, there really will be only one timestamp, as this is the one associated with the barcode itself, instead of one of potentially many surveys.

charles-cowart:
@cassidysymons @wasade Hi guys - I think responding to those two open questions makes this tentatively resolved again. All other comments have been resolved. Please let me know what you think. Best, C

microsetta_private_api/admin/tests/test_admin_api.py
cassidysymons: That test isn't accurately exercising the right logic. If you look at how ag.source_barcodes_surveys functions and is populated, it essentially grabs everything from ag.ag_login_surveys at the time that a user finishes claiming a sample. The above test wouldn't exercise that logic, as it's not going to populate ag.source_barcodes_surveys with the new surveys being taken.

charles-cowart: Hmm, I made a point to display the SQL statements being generated and reviewed them during development, so it's a little surprising that something is amiss. I will dig into it first thing in the morning.

cassidysymons:
@wasade When you have a moment, can you please give this a look and see if there are any unresolved issues or further changes you think would be valuable?

wasade (approved):
I think this is okay from a static analysis.

wasade:
@cassidysymons I recommend squash and merge given the number of commits

PR #459, opened 2022-09-19
Move FFQ's from sample to source

SyedAli-789:

PR #491, opened 2022-11-23
Perk Fulfillment

cassidysymons:
WIP - Need to drop in a few email templates and wrap up unit tests.

Not ready for review, uploading to check lint/tests.

### DIFF
--- a/microsetta_private_api/celery_utils.py
+++ b/microsetta_private_api/celery_utils.py
@@ -35,11 +35,31 @@ def init_celery(celery, app):
         },
         "poll_daklapack_orders": {
             "task": "microsetta_private_api.admin.daklapack_polling.poll_dak_orders",  # noqa
-            "schedule":  60 * 60 * 24  # every 24 hours
+            "schedule":  60 * 60 * 4  # every 4 hour
         },
         "update_qiita_metadata": {
             "task": "microsetta_private_api.tasks.update_qiita_metadata",  # noqa
             "schedule":  60 * 60 * 24  # every 24 hours
+        },
+        "pull_fundrazr_transactions": {
+            "task": "microsetta_private_api.util.fundrazr.get_fundrazr_transactions",  # noqa
+            "schedule": 60 * 60  # every hour
+        },
+        "fulfill_new_transactions": {
+            "task": "microsetta_private_api.util.perk_fulfillment.fulfill_new_transactions",  # noqa
+            "schedule": 60 * 60  # every hour
+        },
+        "fulfill_subscriptions": {
+            "task": "microsetta_private_api.util.perk_fulfillment.process_subscription_fulfillments",  # noqa
+            "schedule": 60 * 60 * 24  # every 24 hours
+        },
+        "check_shipping_updates": {
+            "task": "microsetta_private_api.util.perk_fulfillment.check_shipping_updates",  # noqa
+            "schedule": 60 * 60 * 4  # every 4 hours
+        },
+        "perks_without_fulfillment_details": {
+            "task": "microsetta_private_api.util.perk_fulfillment.perks_without_fulfillment_details",  # noqa
+            "schedule": 60 * 60 * 24  # every 24 hours
         },
         # "fetch_ffqs": {
         #     "task": "microsetta_private_api.util.vioscreen.fetch_ffqs",
--- a/microsetta_private_api/db/patches/0108.sql
+++ b/microsetta_private_api/db/patches/0108.sql
@@ -1,0 +1,85 @@
+-- Add a flag to the campaign.fundrazr_transaction_perk table to reflect whether it has been processed
+ALTER TABLE campaign.fundrazr_transaction_perk ADD COLUMN processed BOOLEAN NOT NULL DEFAULT FALSE;
+
+-- Add a flag to the campaign.fundrazr_daklapack_orders table to reflect whether we've sent out a tracking number
+ALTER TABLE campaign.fundrazr_daklapack_orders ADD COLUMN tracking_sent BOOLEAN NOT NULL DEFAULT FALSE;
+
+-- The campaign.fundrazr_perk_to_daklapack_article table hasn't been used yet, so it's safe to drop.
+DROP TABLE campaign.fundrazr_perk_to_daklapack_article;
+
+-- Since our model has changed to include perks that are FFQ-only and subscriptions,
+-- we're going to retool the table to better reflect how perks function.
+-- For subscriptions, we'll utilize the fulfillment_spacing_* columns to control scheduling.
+-- E.g., fulfillment_spacing_number = 3 and fulfillment_spacing_unit = 'months' will schedule quarterly orders.
+-- If we decide to offer perks with multiple kits shipped at once in the future, fulfillment_spacing_number = 0 can be used to reflect this behavior.
+CREATE TYPE FULFILLMENT_SPACING_UNIT AS ENUM ('days', 'months');
+CREATE TABLE campaign.fundrazr_perk_fulfillment_details (
+    perk_id VARCHAR NOT NULL PRIMARY KEY,
+    ffq_quantity INTEGER NOT NULL,
+    kit_quantity INTEGER NOT NULL,
+    dak_article_code VARCHAR, -- Must be nullable, as not all perks include a kit
+    fulfillment_spacing_number INTEGER NOT NULL,
+    fulfillment_spacing_unit FULFILLMENT_SPACING_UNIT,
+    CONSTRAINT fk_perk_to_dak FOREIGN KEY (dak_article_code) REFERENCES barcodes.daklapack_article (dak_article_code)
+);
+
+-- The API will pull down perks automatically, but we need it to exist so we can add the fullfilment info,
+-- so we're just going to insert them here
+INSERT INTO campaign.fundrazr_perk
+    (id, remote_campaign_id, title, price)
+    VALUES ('3QeVd', '4Tqx5', 'Analyze Your Nutrition', 20),
+           ('3QeW6', '4Tqx5', 'Explore Your Microbiome', 180),
+           ('0QeXa', '4Tqx5', 'Follow Your Gut', 720);
+
+-- Insert the fulfillment info for the perks we're offering
+INSERT INTO campaign.fundrazr_perk_fulfillment_details
+    (perk_id, ffq_quantity, kit_quantity, dak_article_code, fulfillment_spacing_number, fulfillment_spacing_unit)
+    VALUES ('3QeVd', 1, 0, NULL, 0, NULL),
+           ('3QeW6', 1, 1, '3510005E', 0, NULL),
+           ('0QeXa', 4, 4, '3510005E', 3, 'months');
+
+-- Both the subscriptions and subscriptions_fulfillment tables will have cancelled flags to create
+-- an audit trail in the event someone contacts us to cancel scheduled shipments.
+-- We're storing the flag at both levels so we can track what portion of a subscription was actually sent.
+-- We say "No refunds" but it would be good to have the granular data on what people received, just in case.
+CREATE TABLE campaign.subscriptions (
+    subscription_id UUID PRIMARY KEY DEFAULT uuid_generate_v4(),
+    account_id UUID, -- Must be nullable in case someone contributes to receive a subscription before creating their account
+    transaction_id VARCHAR NOT NULL,
+    fundrazr_transaction_perk_id UUID NOT NULL,
+    cancelled BOOLEAN NOT NULL DEFAULT FALSE,
+    CONSTRAINT fk_account_id FOREIGN KEY (account_id) REFERENCES ag.account (id),
+    CONSTRAINT fk_transaction_id FOREIGN KEY (transaction_id) REFERENCES campaign.transaction (id),
+    CONSTRAINT fk_ftp_id FOREIGN KEY (fundrazr_transaction_perk_id) REFERENCES campaign.fundrazr_transaction_perk (id)
+);
+
+-- Participants can alter shipment dates, but only once per shipment.
+-- The fulfillment_date_changed column will manage this.
+CREATE TYPE FULFILLMENT_TYPE AS ENUM ('ffq', 'kit');
+CREATE TABLE campaign.subscriptions_fulfillment (
+    fulfillment_id UUID PRIMARY KEY DEFAULT uuid_generate_v4(),
+    subscription_id UUID NOT NULL,
+    fulfillment_type FULFILLMENT_TYPE NOT NULL,
+    dak_article_code VARCHAR, -- Must be nullable, as not all perks include a kit
+    fulfillment_date DATE,
+    fulfillment_date_changed BOOLEAN NOT NULL DEFAULT FALSE,
+    fulfilled BOOLEAN NOT NULL DEFAULT FALSE,
+    cancelled BOOLEAN NOT NULL DEFAULT FALSE,
+    CONSTRAINT fk_subscription_id FOREIGN KEY (subscription_id) REFERENCES campaign.subscriptions (subscription_id),
+    CONSTRAINT fk_dak_article_code FOREIGN KEY (dak_article_code) REFERENCES barcodes.daklapack_article (dak_article_code)
+);
+
+-- FFQs will have a registration code (similar to the old activation code) moving forward,
+-- although it will not be tied to an email address.
+CREATE TABLE campaign.ffq_registration_codes (
+    ffq_registration_code VARCHAR PRIMARY KEY,
+    registration_code_used TIMESTAMP -- Nullable => null = unused code
+)
+
+-- Create a record of the fulfillment of FFQ codes relative to a transaction/perk combination.
+CREATE TABLE campaign.fundrazr_ffq_codes (
+    fundrazr_transaction_perk_id UUID NOT NULL,
+    ffq_registration_code VARCHAR NOT NULL,
+    CONSTRAINT fk_ftp_id FOREIGN KEY (fundrazr_transaction_perk_id) REFERENCES campaign.fundrazr_transaction_perk (id),
+    CONSTRAINT fk_ffq_code FOREIGN KEY (ffq_registration_code) REFERENCES campaign.ffq_registration_codes (ffq_registration_code)
+)
--- a/microsetta_private_api/model/subscription.py
+++ b/microsetta_private_api/model/subscription.py
@@ -1,0 +1,21 @@
+from microsetta_private_api.model.model_base import ModelBase
+
+
+class Subscription(ModelBase):
+    def __init__(self, **kwargs):
+        # subscription_id won't exist yet on new subscriptions
+        self.subscription_id = kwargs.get('subscription_id')
+
+        # absolute minimum fields required for a subscription
+        self.transaction_id = kwargs['transaction_id']
+
+        # remaining fields are either optional or auto-created later
+        self.account_id = kwargs.get('account_id')
+        self.cancelled = kwargs.get('cancelled', False)
+
+    def to_api(self):
+        return self.__dict__.copy()
+
+    @classmethod
+    def from_dict(cls, values_dict):
+        return cls(**values_dict)
--- a/microsetta_private_api/repo/perk_fulfillment_repo.py
+++ b/microsetta_private_api/repo/perk_fulfillment_repo.py
@@ -1,0 +1,527 @@
+import pycountry
+from datetime import datetime
+from dateutil.relativedelta import relativedelta
+
+from microsetta_private_api.repo.base_repo import BaseRepo
+from microsetta_private_api.exceptions import RepoException
+from microsetta_private_api.model.subscription import Subscription
+from microsetta_private_api.admin.admin_impl import\
+    create_daklapack_order_internal
+from microsetta_private_api.model.daklapack_order import FEDEX_PROVIDER,\
+    FEDEX_2DAY_SHIPPING
+from microsetta_private_api.model.activation_code import ActivationCode
+
+
+class PerkFulfillmentRepo(BaseRepo):
+    def __init__(self, transaction):
+        super().__init__(transaction)
+
+    def _row_to_subscription(self, r):
+        return Subscription(r['subscription_id'], r['transaction_id'],
+                            r['account_id'], r['cancelled'],)
+
+    def find_perks_without_fulfillment_details(self):
+        """If a perk isn't represented in our system at all, the transaction
+        won't be imported. However, we need to have fulfillment details to
+        act on a perk being selected. This finds any perks for which we
+        haven't established fulfillment details and returns content for an
+        alert email.
+        """
+        with self._transaction.dict_cursor() as cur:
+            cur.execute(
+                "SELECT transaction_id, perk_id "
+                "FROM campaign.fundrazr_transaction_perk "
+                "WHERE perk_id NOT IN ("
+                "    SELECT DISTINCT(perk_id) "
+                "    FROM campaign.fundrazr_perk_fulfillment_details"
+                ")"
+            )
+            rows = cur.fetchall()
+            if rows is None:
+                return None
+            else:
+                return [dict(r) for r in rows]
+
+    def process_pending_fulfillments(self):
+        """Find all purchased perks that have yet to be processed
+        and fulfill them"""
+        error_report = []
+
+        with self._transaction.dict_cursor() as cur:
+            cur.execute(
+                "SELECT ftp.id ftp_id, ftp.transaction_id, ftp.perk_id, "
+                "ftp.quantity, ft.payer_email, fpfd.ffq_quantity, "
+                "fpfd.kit_quantity, fpfd.dak_article_code, "
+                "fpfd.fulfillment_spacing_number, "
+                "fpfd.fulfillment_spacing_unit, iu.first_name, iu.last_name, "
+                "iu.phone, iu.address_1, iu.address_2, iu.city, iu.state, "
+                "iu.postal_code, iu.country, iu.campaign_id "
+                "FROM campaign.fundrazr_transaction_perk ftp "
+                "INNER JOIN campaign.transaction ft "
+                "ON ftp.transaction_id = ft.id "
+                "INNER JOIN campaign.fundrazr_perk_fulfillment_details fpfd "
+                "ON ftp.perk_id = fpfd.perk_id "
+                "INNER JOIN campaign.interested_users iu "
+                "ON ft.interested_user_id = iu.interested_user_id "
+                "AND iu.address_valid = true "
+                "WHERE ftp.processed = false"
+            )
+            rows = cur.fetchall()
+
+            for row in rows:
+                if (row['ffq_quantity'] > 1 or row['kit_quantity'] > 1) and \
+                        row['fulfillment_spacing_number'] > 0:
+                    subscription_id = \
+                        self._create_subscription(row['payer_email'],
+                                                  row['transaction_id'],
+                                                  row['ftp_id'])
+
+                for x in range(row['ffq_quantity']):
+                    if x > 0 and row['fulfillment_spacing_number'] > 0:
+                        fulfillment_date =\
+                            self._future_fulfillment_date(
+                                row['fulfillment_spacing_number'],
+                                row['fulfillment_spacing_unit'],
+                                x
+                            )
+                        self._schedule_ffq(subscription_id, fulfillment_date,
+                                           False)
+                    else:
+                        self._fulfill_ffq(row['ftp_id'], row['first_name'],
+                                          row['payer_email'])
+                        if row['ffq_quantity'] > 0 and\
+                                row['fulfillment_spacing_number'] > 0:
+                            # If this is the first FFQ of a subscription,
+                            # we mark it as both scheduled and fulfilled
+                            cur_date = datetime.now()
+                            cur_date = cur_date.strftime("%Y-%m-%d")
+                            self._schedule_ffq(subscription_id, cur_date,
+                                               True)
+
+                for x in range(row['kit_quantity']):
+                    if x > 0 and row['fulfillment_spacing_number'] > 0:
+                        fulfillment_date =\
+                            self._future_fulfillment_date(
+                                row['fulfillment_spacing_number'],
+                                row['fulfillment_spacing_unit'],
+                                x
+                            )
+                        self._schedule_kit(subscription_id,
+                                           fulfillment_date,
+                                           row['dak_article_code'],
+                                           False)
+                    else:
+                        country = pycountry.countries.get(alpha_2=
+                                                          row['country'])
+                        country_name = country.name
+
+                        projects =\
+                            self._campaign_id_to_projects(row['campaign_id'])
+                        address_dict = {
+                            "firstName": row['first_name'],
+                            "lastName": row['last_name'],
+                            "address1": row['address_1'],
+                            "insertion": "",
+                            "address2": row['address_2'],
+                            "postalCode": row['postal_code'],
+                            "city": row['city'],
+                            "state": row['state'],
+                            "country": country_name,
+                            "countryCode": row['country'],
+                            "phone": row['phone']
+                        }
+                        status, return_val =\
+                           self._fulfill_kit(row['ftp_id'],
+                                             projects,
+                                             row['dak_article_code'],
+                                             1,
+                                             address_dict)
+                        if not status:
+                            # Daklapack order failed, let the error percolate
+                            error_report.append(return_val)
+
+                        if row['kit_quantity'] > 0 and\
+                                row['fulfillment_spacing_number'] > 0:
+                            # If this is the first kit of a subscription,
+                            # we mark it as both scheduled and fulfilled
+                            cur_date = datetime.now()
+                            cur_date = cur_date.strftime("%Y-%m-%d")
+                            self._schedule_kit(subscription_id,
+                                               cur_date,
+                                               row['dak_article_code'],
+                                               True)
+                cur.execute(
+                    "UPDATE campaign.fundrazr_transaction_perk "
+                    "SET processed = true "
+                    "WHERE id = %s",
+                    (row['ftp_id'], )
+                )
+
+        return error_report
+
+    def process_subscription_fulfillments(self):
+        error_report = []
+
+        with self._transaction.dict_cursor() as cur:
+            cur.execute(
+                "SELECT sf.fulfillment_id, sf.fulfillment_type, "
+                "sf.dak_article_code, ftp.id ftp_id, ft.payer_email, "
+                "iu.first_name iu_first_name, iu.last_name iu_last_name, "
+                "iu.phone iu_phone, iu.address_1 iu_address_1, "
+                "iu.address_2 iu_address_2, iu.city iu_city, "
+                "iu.state iu_state, iu.postal_code iu_postal_code, "
+                "iu.country iu_country, iu.campaign_id, s.account_id, "
+                "a.email a_email, a.first_name a_first_name, "
+                "a.last_name a_last_name, a.street a_address_1, "
+                "a.city a_city, a.state a_state, a.post_code a_postal_code, "
+                "a.country_code a_country "
+                "FROM campaigns.subscriptions_fulfillment sf "
+                "INNER JOIN campaigns.subscriptions s "
+                "ON sf.subscription_id = s.subscription_id "
+                "INNER JOIN campaigns.fundrazr_transaction_perk ftp "
+                "ON s.fundrazr_transaction_perk_id = ftp.id "
+                "INNER JOIN campaigns.transaction ft "
+                "ON ftp.transaction_id = ft.id "
+                "INNER JOIN campaign.interested_users iu "
+                "ON ft.interested_user_id = iu.interested_user_id "
+                "LEFT JOIN ag.account a"
+                "ON s.account_id = a.id"
+                "WHERE sf.fulfilled = false AND sf.cancelled = FALSE "
+                "AND sf.fulfillment_date <= CURRENT_DATE"
+            )
+            rows = cur.fetchall()
+            for row in rows:
+                if row['fulfillment_type'] == "ffq":
+                    # If an account is linked to the subscription, we use
+                    # that account's first name and email
+                    if row['account_id']:
+                        self._fulfill_ffq(row['ftp_id'], row['a_first_name'],
+                                          row['a_email'])
+                    # If no account, fall back to original Fundrazr data
+                    else:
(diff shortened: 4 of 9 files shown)


microsetta_private_api/repo/perk_fulfillment_repo.py
```diff
@@ -0,0 +1,591 @@
...
+            )
+            rows = cur.fetchall()
+
+            for row in rows:
+                if (row['ffq_quantity'] > 1 or row['kit_quantity'] > 1) and \
+                        row['fulfillment_spacing_number'] > 0:
+                    subscription_id = \
+                        self._create_subscription(row['payer_email'],
+                                                  row['transaction_id'],
+                                                  row['ftp_id'])
```
wasade, line 80:
```suggestion
                if has_subscription(row): 
                    subscription_id = \
                        self._create_subscription(row['payer_email'],
                                                  row['transaction_id'],
                                                  row['ftp_id'])
                else:
                    subscription_id = None 
```

...and add a method elsewhere:

```python
def has_subscription(d):
    return (d['ffq_quantity'] > 1 or d['kit_quantity'] > 1) and \
           (d['fulfillment_spacing_number'] > 0)

File: microsetta_private_api/repo/perk_fulfillment_repo.py [hunk 2]
```diff
@@ -0,0 +1,591 @@
...
+                            send_email(
+                                row['payer_email'],
+                                template,
+                                {
+                                    "first_name": row['first_name'],
+                                    "registration_code": registration_code,
+                                    "interface_endpoint":
+                                        SERVER_CONFIG["interface_endpoint"]
+                                },
+                                EN_US
```
wasade, line 115: Does this force the email to be in english?

wasade: ...granted, I guess we don't have anything right now to make a determination one way or another so I makes sense to punt

microsetta_private_api/repo/perk_fulfillment_repo.py
```diff
@@ -0,0 +1,591 @@
...
+
+            for row in rows:
+                if (row['ffq_quantity'] > 1 or row['kit_quantity'] > 1) and \
+                        row['fulfillment_spacing_number'] > 0:
+                    subscription_id = \
+                        self._create_subscription(row['payer_email'],
+                                                  row['transaction_id'],
+                                                  row['ftp_id'])
+
+                for x in range(row['ffq_quantity']):
```
wasade, line 82: This for loop is pretty complex. It should simplify if `range(1, row['ffq_quantity'])` is used, and the contained `else` block is performed outside of the forloop. This would reduce the complexity of the forloop, and simplify the contained conditional

wasade: ...it would nice for readability if these blocks were decomposed to functions, but not critical.

microsetta_private_api/repo/perk_fulfillment_repo.py
```diff
@@ -0,0 +1,591 @@
...
+                                    "interface_endpoint":
+                                        SERVER_CONFIG["interface_endpoint"]
+                                },
+                                EN_US
+                            )
+                        except:  # noqa
+                            # try our best to email
+                            pass
+
+                        if row['ffq_quantity'] > 0 and\
```
wasade, line 121: I dont believe this conditional actually needs to be in the forloop as it does not depend on `x`

microsetta_private_api/repo/perk_fulfillment_repo.py
```diff
@@ -0,0 +1,591 @@
...
+                                row['fulfillment_spacing_number'] > 0:
+                            # If this is the first FFQ of a subscription,
+                            # we mark it as both scheduled and fulfilled
+                            cur_date = datetime.now()
+                            cur_date = cur_date.strftime("%Y-%m-%d")
+                            self._schedule_ffq(subscription_id, cur_date,
+                                               True)
+
+                for x in range(row['kit_quantity']):
```
wasade, line 130: Qualitatively the same comments with this forloop

microsetta_private_api/repo/perk_fulfillment_repo.py
```diff
@@ -0,0 +1,591 @@
...
+                "INNER JOIN campaign.fundrazr_perk_fulfillment_details fpfd "
+                "ON ftp.perk_id = fpfd.perk_id "
+                "INNER JOIN campaign.interested_users iu "
+                "ON ft.interested_user_id = iu.interested_user_id "
+                "AND iu.address_valid = true "
+                "WHERE ftp.processed = false"
+            )
+            rows = cur.fetchall()
+
+            for row in rows:
```
wasade, line 74: I *think* we actually want to operate within a transaction per order, committing along the way, so that we can safely rollback without one order affecting another. As this is written, if one order causes an exception, all orders being evaluated will rollback (but we cannot rollback sent emails and communications to daklapack). All orders would remain affected until the problematic data, assuming easily traced, is fixed. 

What would be necessary I believe is a lock on the `fundrazr_transaction_perk` table to avoid the possibility of multiple processes attempting to `process_pending_fulfillments`. Once set, a single transcation can be created to obtain `rows`, and then individual transactions per row can be created.

microsetta_private_api/repo/perk_fulfillment_repo.py
```diff
@@ -0,0 +1,591 @@
...
+                        status, return_val = self._fulfill_kit(
+                            row['ftp_id'],
+                            projects,
+                            row['dak_article_code'],
+                            1,
+                            address_dict
+                        )
+                        if not status:
+                            # Daklapack order failed, let the error percolate
+                            error_report.append(return_val)
```
wasade, line 172: Is it correct to still mark the order as processed?

microsetta_private_api/repo/perk_fulfillment_repo.py
```diff
@@ -0,0 +1,591 @@
...
+
+                        if row['kit_quantity'] > 0 and\
+                                row['fulfillment_spacing_number'] > 0:
```
wasade, line 175: similar to above, it would be nice for this to be `if is_first_kit_in_subscription(...)`

microsetta_private_api/repo/perk_fulfillment_repo.py
wasade: ...similar to above, would be nice if this is `if is_first_ffq_in_subscription(...)`

microsetta_private_api/repo/perk_fulfillment_repo.py
wasade: and also, I don't believe this statemetn depends on the loop variable

microsetta_private_api/repo/perk_fulfillment_repo.py
```diff
@@ -0,0 +1,591 @@
...
+                "ON sf.subscription_id = s.subscription_id "
+                "INNER JOIN campaigns.fundrazr_transaction_perk ftp "
+                "ON s.fundrazr_transaction_perk_id = ftp.id "
+                "INNER JOIN campaigns.transaction ft "
+                "ON ftp.transaction_id = ft.id "
+                "INNER JOIN campaign.interested_users iu "
+                "ON ft.interested_user_id = iu.interested_user_id "
+                "LEFT JOIN ag.account a"
+                "ON s.account_id = a.id"
+                "WHERE sf.fulfilled = false AND sf.cancelled = FALSE "
```
wasade, line 220:
```suggestion
                "WHERE sf.fulfilled = FALSE AND sf.cancelled = FALSE "
```

microsetta_private_api/repo/perk_fulfillment_repo.py
```diff
@@ -0,0 +1,591 @@
...
+                "AND sf.fulfillment_date <= CURRENT_DATE"
+            )
+            rows = cur.fetchall()
+            for row in rows:
+                if row['fulfillment_type'] == "ffq":
+                    registration_code = self._fulfill_ffq(row['ftp_id'])
+
+                    # If an account is linked to the subscription, we use
+                    # that account's first name and email
+                    if row['account_id']:
```
wasade, line 230:
```suggestion
                    if row['account_id'] is not None:
```

microsetta_private_api/repo/perk_fulfillment_repo.py
```diff
@@ -0,0 +1,591 @@
...
+                        )
+                    except:  # noqa
+                        # try our best to email
+                        pass
+
+                elif row['fulfillment_type'] == "kit":
+                    projects = \
+                        self._campaign_id_to_projects(row['campaign_id'])
+
+                    if row['account_id']:
```
wasade, line 258:
```suggestion
                    if row['account_id'] is not None:
```

microsetta_private_api/repo/perk_fulfillment_repo.py
```diff
@@ -0,0 +1,591 @@
...
+
+                    status, return_val = \
+                        self._fulfill_kit(row['ftp_id'],
+                                          projects,
+                                          row['dak_article_code'],
+                                          1,
+                                          address_dict)
+                    if not status:
+                        # Daklapack order failed, let the error percolate
+                        error_report.append(return_val)
```
wasade, line 305: Should the order be considered fulfilled if this error happens?

microsetta_private_api/repo/perk_fulfillment_repo.py
```diff
@@ -0,0 +1,591 @@
...
+                "ON ftp.transaction_id = ft.id "
+                "INNER JOIN campaign.interested_users iu "
+                "ON ft.interested_user_id = iu.interested_user_id "
+                "LEFT JOIN ag.account a"
+                "ON s.account_id = a.id"
+                "WHERE sf.fulfilled = false AND sf.cancelled = FALSE "
+                "AND sf.fulfillment_date <= CURRENT_DATE"
+            )
+            rows = cur.fetchall()
+            for row in rows:
```
wasade, line 224: Like above, and for the same reason, I think this needs to be expressed with a transaction per row otherwise an unknown error condition with one row affects all others.

microsetta_private_api/repo/perk_fulfillment_repo.py
```diff
@@ -0,0 +1,591 @@
...
+                                          1,
+                                          address_dict)
+                    if not status:
+                        # Daklapack order failed, let the error percolate
+                        error_report.append(return_val)
+                else:
+                    error_report.append(
+                        f"Subscription fulfillment {row['fulfillment_id']} "
+                        f"contains malformed fulfillment_type "
+                        f"{row['fulfillmnet_type']}"
```
wasade, line 310: This would have resulted in a `KeyError` causing all order fulfillments to rollback. `fulfillmnet_type` -> `fulfillment_type`

microsetta_private_api/repo/perk_fulfillment_repo.py
```diff
@@ -0,0 +1,591 @@
...
+                "ON do.dak_order_id = dotk.dak_order_id "
+                "INNER JOIN barcodes.kit kit "
+                "ON dotk.kit_uuid = kit.kit_uuid "
+                "INNER JOIN campaign.fundrazr_transaction_perk ftp "
+                "ON fdo.fundrazr_transaction_perk_id = ftp.id "
+                "INNER JOIN campaign.transaction t"
+                "ON ftp.transaction_id = t.id"
+            )
+            rows = cur.fetchall()
+            for r in rows:
```
wasade, line 344: I _think_ this one does not need to be a transaction per row as the critical code is wrapped, and the forloop is not very complex

microsetta_private_api/repo/perk_fulfillment_repo.py
```diff
@@ -0,0 +1,591 @@
...
+                        },
+                        EN_US
+                    )
+                except:  # noqa
+                    # try our best to email
+                    pass
+
+                cur.execute(
+                    "UPDATE campaign.fundrazr_daklapack_orders "
+                    "SET tracking_sent = true "
```
wasade, line 361: Shouldn't we *not* do this if we failed to send an email?

microsetta_private_api/repo/perk_fulfillment_repo.py
wasade: However, it would create confusion if `tracking_sent = false`, when we have actually sent tracking information out, so this may also need to be separate transactions. Confusing for admins, and possibly means we send repeated emails to people

microsetta_private_api/db/patches/0108.sql
```diff
@@ -0,0 +1,85 @@
...
+    fulfillment_spacing_number INTEGER NOT NULL,
+    fulfillment_spacing_unit FULFILLMENT_SPACING_UNIT,
+    CONSTRAINT fk_perk_to_dak FOREIGN KEY (dak_article_code) REFERENCES barcodes.daklapack_article (dak_article_code)
+);
+
+-- The API will pull down perks automatically, but we need it to exist so we can add the fullfilment info,
+-- so we're just going to insert them here
+INSERT INTO campaign.fundrazr_perk
+    (id, remote_campaign_id, title, price)
+    VALUES ('3QeVd', '4Tqx5', 'Analyze Your Nutrition', 20),
```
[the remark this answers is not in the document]
cassidysymons: So their production and staging systems have the same content/perks for our campaigns, but those perks and campaigns are attached to different IDs?

wasade:
Staging doesn't mirror production at all iirc

microsetta_private_api/db/patches/0108.sql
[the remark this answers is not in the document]
wasade: Various comments in the client tests elude to the discrepancies (https://github.com/biocore/microsetta-private-api/blob/master/microsetta_private_api/client/tests/test_fundrazr.py)

microsetta_private_api/db/patches/0108.sql
```diff
@@ -0,0 +1,85 @@
...
+    fulfillment_date_changed BOOLEAN NOT NULL DEFAULT FALSE,
+    fulfilled BOOLEAN NOT NULL DEFAULT FALSE,
+    cancelled BOOLEAN NOT NULL DEFAULT FALSE,
+    CONSTRAINT fk_subscription_id FOREIGN KEY (subscription_id) REFERENCES campaign.subscriptions (subscription_id),
+    CONSTRAINT fk_dak_article_code FOREIGN KEY (dak_article_code) REFERENCES barcodes.daklapack_article (dak_article_code)
+);
+
+-- FFQs will have a registration code (similar to the old activation code) moving forward,
+-- although it will not be tied to an email address.
+CREATE TABLE campaign.ffq_registration_codes (
```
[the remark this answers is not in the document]
cassidysymons: The association between a source and a used registration code will be logged in a separate table. The groundwork for that is in #459 and I'll be adding appropriate foreign key constraints in a separate PR, once both this and 459 are merged.

The campaign.fundrazr_ffq_codes table offers an audit trail of what was created and disseminated in the event that a participant claims they didn't receive their code. An admin interface for that is planned, but deferred until later as it's not an absolute necessity.

microsetta_private_api/repo/perk_fulfillment_repo.py
cassidysymons: It does, and follows the precedent from campaign_repo.py ([here](https://github.com/biocore/microsetta-private-api/blob/master/microsetta_private_api/repo/campaign_repo.py#L472)).

microsetta_private_api/db/patches/0108.sql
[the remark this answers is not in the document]
cassidysymons: After poking at their staging system system, that whole situation makes a lot more sense - it's not just an API staging server, it's a staging version of their entire platform. This actually would've been valuable for Kat to have access to in developing the content for the new perks. AFAIK, she was unaware of it.

That tangent aside, I don't think the inconsistency between staging and production means we shouldn't insert the IDs into our database. The perk->fulfillment details need to be explicitly expressed in some form, and the perks need to be there to fulfill the foreign key constraint.

Long term, I think it would be valuable to add a flag for production/staging to these tables so we can internally track the meaning of IDs and avoid the (admittedly low) potential for ID collisions between the two platforms (I have no idea if it's possible, but in the absence of explicit information to the contrary, I'd assume it is possible).

Short term, I don't think it's strictly necessary to address that prior to relaunch. The Fundrazr code has been tested, and unless they've changed their API, we don't need to interact with their staging system to prepare for relaunch. If they have changed their API, we'll need to deal with a whole set of other issues.

microsetta_private_api/repo/perk_fulfillment_repo.py
cassidysymons: I think the pragmatic answer here is "yes," as each record in campaign.fundrazr_transaction_perk can correlate to more than a single fulfillment item. For example, a subscription record in that table would mean we're going to immediately send an FFQ code, immediately dispatch a kit, and schedule three more of each. If the kit shipment failed and we left the row in the table marked as processed=false, the next time the script ran, we'd repeat the cycle of re-fulfilling each item even though only one specific item is still in need of fulfillment.

The defensive approach here seems to be mark the campaign.ftp record as processed and investigate/rectify fulfillment problems based on the error log that goes to the pester email. It creates manual labor for whoever is receiving the pester emails at any given time (me at the time of writing, but it's unsafe to assume that will always be the case), but erring on the side of caution seems prudent. The alternative is to re-structure the relationship between some of the perk fulfillment tables to eliminate the one-to-many relationship, but since an order failure is _very_ likely to be something we'd want to investigate anyway, I'd argue that the current approach is suitable.

microsetta_private_api/repo/perk_fulfillment_repo.py
cassidysymons: For subscription fulfillment, no. Each record in that table is a single item - either a kit or an FFQ - so there's a 1:1 relationship and we can safely re-try failed orders. Updated code will reflect this.

microsetta_private_api/repo/perk_fulfillment_repo.py
cassidysymons: I definitely see the value in creating a new transaction for each row, but the idea of combining nested transactions with a table locked at the outer transaction seems problematic. If we lock the table for the query that obtains rows (in the main transaction), then the inner transaction for each row won't be able to update the table to reflect an accurate status.

Unless I'm missing something, it seems safer to query for all matching rows, loop through them, create the row-level transaction, lock the table at the row-level transaction level, run a select to verify that the processed column hasn't changed, then go through the rest of the fulfillment code. This way, if multiple processes are running process_pending_fulfillments(), they might have shared results from the outer transaction's select query, but any given row will be within a locked transaction to prevent the same fundrazr_transaction_perk from processing twice.

microsetta_private_api/repo/perk_fulfillment_repo.py
cassidysymons: Agreed, code changes pending the discussion on the earlier comment above.

cassidysymons:
@wasade I believe I've either resolved or responded to each comment you left on Wednesday. If I resolved a comment, it means I have updated code ready to push up to address it. Would you prefer I push those updates before you resume the code review? Or do you want to finish looking through the code you haven't reviewed before I push updates?

microsetta_private_api/db/patches/0108.sql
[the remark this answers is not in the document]
wasade: Sounds good, thanks!

microsetta_private_api/repo/perk_fulfillment_repo.py
wasade: Great, thought so

microsetta_private_api/repo/perk_fulfillment_repo.py
wasade: That is an excellent point, good catch.

microsetta_private_api/repo/perk_fulfillment_repo.py
wasade: Thanks. I believe then that the `error_report` should contain additional provenance information so that an administrator can easily examine the specific problematic database state. I worry that with the present detail, it will be unclear what records are associated to a `return_val`

microsetta_private_api/repo/perk_fulfillment_repo.py
wasade: Great, thanks

microsetta_private_api/db/patches/0108.sql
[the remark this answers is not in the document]
cassidysymons: Can you please add me as an admin on our Fundrazr staging account so I can set up appropriate perks to test?

microsetta_private_api/repo/perk_fulfillment_repo.py
```diff
@@ -0,0 +1,601 @@
...
+        """Cancels a subscription and associated future fulfillments
+
+        Parameters
+        ----------
+        subscription_id : str
+            A subscription ID
+
+        Returns
+        -------
+        True
```
wasade, line 553: I believe other areas of the codebase tend to use `rowcount` for similar scenarios, which allows the callee to make a decision about what to bubble out of the API. 

Note that the current behavior is to return `None` not `True`

microsetta_private_api/repo/perk_fulfillment_repo.py
```diff
@@ -0,0 +1,601 @@
...
+            project_rows = cur.fetchall()
+            project_ids = [project['project_id'] for project in project_rows]
+            return project_ids
+
+    def _future_fulfillment_date(self, spacing_number, spacing_unit,
+                                 multiplier):
+        cur_date = datetime.now()
+        if spacing_unit == "days":
+            new_date = cur_date + relativedelta(
+                days=+(spacing_number*multiplier)
```
wasade, line 587: What's the purpose of the `+`? It doesn't ensure the value is positive:

```python
>>> +(-1)
-1
>>>
```

If that is the intention, I recommend `abs`

microsetta_private_api/repo/perk_fulfillment_repo.py
```diff
@@ -0,0 +1,601 @@
...
+                (campaign_id,)
+            )
+            project_rows = cur.fetchall()
+            project_ids = [project['project_id'] for project in project_rows]
+            return project_ids
+
+    def _future_fulfillment_date(self, spacing_number, spacing_unit,
+                                 multiplier):
+        cur_date = datetime.now()
+        if spacing_unit == "days":
```
wasade, line 585:
```suggestion
        spacing = spacing_number * multiplier)
        if spacing_unit == "days":
```

microsetta_private_api/repo/perk_fulfillment_repo.py
wasade, line 587:
```suggestion
                days=spacing
```

microsetta_private_api/repo/perk_fulfillment_repo.py
```diff
@@ -0,0 +1,601 @@
...
+            new_date = cur_date + relativedelta(
+                days=+(spacing_number*multiplier)
+            )
+        elif spacing_unit == "months":
+            new_date = cur_date + relativedelta(
+                months=+(spacing_number*multiplier)
```
wasade, line 591:
```suggestion
                months=spacing
```

microsetta_private_api/repo/tests/test_perk_fulfillment_repo.py
```diff
@@ -0,0 +1,538 @@
...
+        }
+
+        with Transaction() as t:
+            ut = UserTransaction(t)
+            ut.add_transaction(TRANSACTION_ONE_KIT)
+            pfr = PerkFulfillmentRepo(t)
+            res = pfr.process_pending_fulfillments()
+
+            # res is a list of errors, which should be 1
+            self.assertNotEqual(len(res), 0)
```
wasade, line 312: Based on the comment, I think the following change makes sense

```suggestion
            self.assertEqual(len(res), 1)
```

microsetta_private_api/templates/email/kit_tracking_number.jinja2
```diff
@@ -0,0 +1,249 @@
...
+<p class="c8 c9">
+    <span class="c0">{{ _("Item: Collection Kit") }}</span><br />
+    <span class="c0">{{ _("Tracking number:") }} {{ tracking_number }}</span><br />
+</p>
+<p class="c8 c9">
+    <span class="c0">&nbsp;</span>
+</p>
+<p class="c8 c9">
+    <span class="c0">
+        {{ _("If you have any questions, please visit our <a href="https://microsetta.ucsd.edu/participant-resources/faqs/">FAQs</a> and <a href="https://microsetta.ucsd.edu/participant-resources/instructions/">instructions</a> pages, or you can email us at ") }}<a href="mailto:microsetta@ucsd.edu">microsetta@ucsd.edu</a>.<br />
```
wasade, line 238: This won't render as expected given the `"`s in the `a href`

microsetta_private_api/templates/email/subscription_ffq_code.jinja2
```diff
@@ -0,0 +1,260 @@
...
+    </style>
+</head>
+<body class="c10">
+<p class="c1 c5"><span class="c0"></span></p>
+<p class="c1">
+    <span class="c0">{{ _("Dear") }} {{ first_name }},</span>
+</p>
+<p class="c1"><span class="c0">&nbsp;</span></p>
+<p class="c8 c9">
+    <span class="c0">{{ _("Thank you for helping to drive research to discover more about the human microbiome. Once the next kit for your ‘Follow Your Gut’ ships, we will send you an email with a tracking number so you will know when it’s on the way.") }}</span>
```
wasade, line 224: Use of non-standard `'` marks

microsetta_private_api/templates/email/subscription_ffq_code.jinja2
```diff
@@ -0,0 +1,260 @@
...
+    <span class="c0"></span>
+</p>
+<p class="c8 c9">
+    <span class="c0">{{ _("Item: Collection Kit") }}</span>
+</p>
+<p class="c2">
+    <span class="c0"></span>
+</p>
+<p class="c8 c9">
+    <span class="c0">{{ _("<strong>What’s next</strong>: Your new food frequency questionnaire will be available to complete when you <a href="{{ interface_endpoint }}">login to your account</a>. Visit the ‘My Nutrition’ page and enter the following:") }}</span>
```
wasade, line 236: Use of a non-standard `'` mark

wasade: Also this won't render as expected given the `"` marks in the `a href`

microsetta_private_api/templates/email/subscription_ffq_code.jinja2
```diff
@@ -0,0 +1,260 @@
...
+</p>
+<p class="c8 c9">
+    <span class="c0">{{ _("Registration code:") }} <strong>{{ registration_code }}</strong></span>
+</p>
+<p class="c8 c9">
+    <span class="c0">&nbsp;</span>
+</p>
+<p class="c8 c9">
+    <span class="c0">
+        {{ _("Thank you for your participation and we hope you enjoy learning more about your gut microbiome! If you have any questions, please visit our <a href="https://microsetta.ucsd.edu/participant-resources/faqs/">FAQs</a> and <a href="https://microsetta.ucsd.edu/participant-resources/instructions/">instructions</a> pages, or you can email us at ") }}<a href="mailto:microsetta@ucsd.edu">microsetta@ucsd.edu</a>.<br />
```
wasade, line 249: Same here

microsetta_private_api/templates/email/thank_you_no_kit.jinja2
```diff
@@ -0,0 +1,266 @@
...
+</p>
+<p class="c1"><span class="c0">&nbsp;</span></p>
+<p class="c8 c9">
+    <span class="c0">{{ _("Thank you for your contribution to The Microsetta Initiative and for helping to drive research to discover more about the human microbiome.") }}</span>
+</p>
+<p class="c2">
+    <span class="c0"></span>
+</p>
+<p class="c8 c9">
+    <span class="c0">{{ _("<strong>What’s next</strong>: If you’re joining us for the first time visit the TMI website to create your account, profile, and fill in the initial forms & surveys. <strong><i>Your email address is linked to this information and must be the same email used to set up your account</i></strong>.") }}</span>
```
wasade, line 230: Non-standard `'`

microsetta_private_api/templates/email/thank_you_no_kit.jinja2
```diff
@@ -0,0 +1,266 @@
...
+    <span class="c0">&nbsp;</span>
+</p>
+<p class="c8 c9">
+    <span class="c0"><a href="{{ interface_endpoint }}"><strong>[{{ _("Create Account or Login") }}</strong></a></span>
+</p>
+<p class="c8 c9">
+    <span class="c0">&nbsp;</span>
+</p>
+<p class="c1">
+    <span class="c0">{{ _("Once completed you will be ready to take your food frequency questionnaire. Visit the ‘My Nutrition’ page and enter the following:") }}</span>
```
wasade, line 242: Same, as well as use of `"`

microsetta_private_api/templates/email/thank_you_no_kit.jinja2
```diff
@@ -0,0 +1,266 @@
...
+</p>
+<p class="c8 c9">
+    <span class="c0">{{ _("Registration code:") }} <strong>{{ registration_code }}</strong></span>
+</p>
+<p class="c8 c9">
+    <span class="c0">&nbsp;</span>
+</p>
+<p class="c8 c9">
+    <span class="c0">
+        {{ _("Thank you for your participation! If you have any questions, please visit our <a href="https://microsetta.ucsd.edu/participant-resources/faqs/">FAQs</a> and <a href="https://microsetta.ucsd.edu/participant-resources/instructions/">instructions</a> pages, or you can email us at ") }}<a href="mailto:microsetta@ucsd.edu">microsetta@ucsd.edu</a>.<br />
```
wasade, line 255: Same here

microsetta_private_api/templates/email/thank_you_with_kit.jinja2
```diff
@@ -0,0 +1,268 @@
...
+    </style>
+</head>
+<body class="c10">
+<p class="c1 c5"><span class="c0"></span></p>
+<p class="c1">
+    <span class="c0">{{ _("Dear") }} {{ first_name }},</span>
+</p>
+<p class="c1"><span class="c0">&nbsp;</span></p>
+<p class="c8 c9">
+    <span class="c0">{{ _("Thank you for helping to drive research to discover more about the human microbiome. Once your collection kit ships, we will send you an email with a tracking number so you will know when it’s on the way.") }}</span>
```
wasade, line 224: I think non standard `'`

microsetta_private_api/templates/email/thank_you_with_kit.jinja2
```diff
@@ -0,0 +1,268 @@
...
+</p>
+<p class="c2">
+    <span class="c0"></span>
+</p>
+<p class="c8 c9">
+    <span class="c0">{{ _("<strong>What’s next</strong>: If you’re joining us for the first time visit the TMI website to create your account, profile, and fill in the initial forms & surveys. <strong><i>Your email address is linked to this information and must be the same email used to set up your account</i></strong>.") }}</span>
```
wasade, line 230: Same

microsetta_private_api/templates/email/thank_you_with_kit.jinja2
```diff
@@ -0,0 +1,268 @@
...
+    <span class="c0">&nbsp;</span>
+</p>
+<p class="c8 c9">
+    <span class="c0"><a href="{{ interface_endpoint }}"><strong>[{{ _("Create Account or Login") }}</strong></a></span>
+</p>
+<p class="c8 c9">
+    <span class="c0">&nbsp;</span>
+</p>
+<p class="c1">
+    <span class="c0">{{ _("Once your profile is complete you will be ready to take your food frequency questionnaire. Visit the ‘My Nutrition’ page and enter the following:") }}</span>
```
wasade, line 242: Same

microsetta_private_api/util/fundrazr.py
```diff
@@ -20,9 +22,12 @@ def _get_load(t):
         latest = tr.most_recent_transaction(transaction_source=tr.TRN_TYPE_FUNDRAZR,  # noqa
                                             include_anonymous=True)
 
-        # if we do not have any transactions, we don't have anything recent
+        # We've decided that there's no reason to copy all of the old
```
wasade, line 25: I disagree, however my concern is non-blocking.

wasade (changes requested):
Please make sure to render the template emails. There are many instances where quote marks cancel out, and less important but tentatively problematic are the use of non-standard quote marks.

microsetta_private_api/repo/perk_fulfillment_repo.py
cassidysymons: The + is following the recommended usage of the relative keyword arguments from the [dateutil docs](https://dateutil.readthedocs.io/en/stable/relativedelta.html). That said, adding an abs() around the spacing_number variable is a good call to ensure all scheduling is in the future.

wasade: So weird, it seems like it's a noop?

```python
>>> +1
1
>>> +(-1)
-1
>>>
```

microsetta_private_api/util/fundrazr.py
cassidysymons: Reverting.

microsetta_private_api/celery_utils.py
```diff
@@ -29,17 +29,37 @@ def __call__(self, *args, **kwargs):
...
+        #    "schedule": 60 * 60  # every hour
+        # },
+        # "fulfill_new_transactions": {
+        #    "task": "microsetta_private_api.util.perk_fulfillment.fulfill_new_transactions",  # noqa
+        #    "schedule": 60 * 60  # every hour
+        # },
+        # "fulfill_subscriptions": {
+        #    "task": "microsetta_private_api.util.perk_fulfillment.process_subscription_fulfillments",  # noqa
+        #    "schedule": 60 * 60 * 24  # every 24 hours
+        # },
```
wasade, line 55: Shouldn't these be uncommented?

microsetta_private_api/celery_worker.py
```diff
@@ -2,10 +2,17 @@
...
 
 # Run any celery tasks that require initialization on worker start
 refresh_headers.delay()  # Initialize the vioscreen task with a token
 poll_dak_orders.delay()  # check for orders
-update_qiita_metadata.delay()  # run Qiita metadata push
+# get_fundrazr_transactions.delay()  # check for new transactions
+check_shipping_updates.delay()  # check for tracking updates
+# update_qiita_metadata.delay()  # run Qiita metadata push
+
+# Disabling Qiita metadata push until we have survey changes in place and
```
wasade, line 17: These were merged though right?

microsetta_private_api/repo/perk_fulfillment_repo.py
```diff
@@ -0,0 +1,779 @@
...
+                    subscription_id = None
+
+                # If there are any FFQs attached to the perk, immediately
+                # fulfill the first one
+                if row['ffq_quantity'] > 0:
+                    # If the perk is a kit or subscription, send thank you
+                    # email with kit content. Otherwise, send thank you
+                    # for FFQ only
+                    if row['kit_quantity'] > 0:
+                        template = "thank_you_with_kit"
```
wasade, line 115: It seems like this should refer to an enum value from `email_templates.py`?

microsetta_private_api/repo/perk_fulfillment_repo.py
```diff
@@ -0,0 +1,779 @@
...
+            template = "kit_tracking_number"
+            for r in rows:
+                try:
+                    email_args = {
+                        "first_name": r['payer_first_name'],
+                        "tracking_number": r['outbound_fedex_tracking']
+                    }
+
+                    email_address = "csymons@eng.ucsd.edu"
+                    # email_address = r['payer_email']
```
wasade, line 349: This should probably be uncommented :)

microsetta_private_api/repo/perk_fulfillment_repo.py
wasade: ...I guess not maybe. just seems surprising this is a string rather than a controlled value

microsetta_private_api/repo/perk_fulfillment_repo.py
```diff
@@ -0,0 +1,779 @@
...
+            try:
+                email_args = {
+                    "first_name": first_name,
+                    "registration_code": code,
+                    "interface_endpoint": SERVER_CONFIG["interface_endpoint"]
+                }
+
+                send_email(
+                    email,
+                    template,
```
wasade, line 498: ...I guess `send_email` is expecting a string here, huh

microsetta_private_api/model/subscription.py
```diff
@@ -0,0 +1,23 @@
+from microsetta_private_api.model.model_base import ModelBase
+
+FULFILLMENT_ACCOUNT_ID = "000fc4cd-8fa4-db8b-e050-8a800c5d81b7"
```
wasade, line 3: I think this should move to `test_perk_fulfillment_repo.py` based on its use

wasade: Or maybe I'm misunderstanding how this is used -- i only saw it used in that script but I may be overlooked something

wasade (commented):
Just a few what I think are small items and questions

microsetta_private_api/celery_worker.py
cassidysymons: The TMI-side changes are in place, we still need to add the new fields to Qiita, which will be done within the next week.

microsetta_private_api/repo/perk_fulfillment_repo.py
cassidysymons: The template argument for send_email is expressed as a string rather than a controlled value throughout the code base. I agree in theory that it would be nice to refine that - and bake the event logging into the email sending, rather than have each section of the code base handle the event logging on its own - but both of those are beyond the scope of this PR.

microsetta_private_api/model/subscription.py
cassidysymons: It's also used in admin_impl.py. I can move it into server_config.json if we'd prefer, but it is used in multiple places.

microsetta_private_api/celery_worker.py
wasade: Thanks!

microsetta_private_api/model/subscription.py
wasade: Okay, ya I think it would be better for this to be in server config. Thanks!

wasade (approved):

Issue #492, opened 2022-11-23
Account deletion not possible when a user takes a polyphenol FFQ


wasade:
From the logs. It appears that the polyphenol FFQ table may not be cleaned up correctly prior to account delete.

```
[2022-11-22 13:06:38 -0800] [7400] [ERROR] Exception on /api/accounts/b499417a-0840-44da-b269-d243f6eb02ae [DELETE]
Traceback (most recent call last):
  File "/Users/api_user/miniconda3/envs/microsetta-private-api/lib/python3.8/site-packages/flask/app.py", line 2446, in wsgi_app
    response = self.full_dispatch_request()
  File "/Users/api_user/miniconda3/envs/microsetta-private-api/lib/python3.8/site-packages/flask/app.py", line 1951, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/Users/api_user/miniconda3/envs/microsetta-private-api/lib/python3.8/site-packages/flask/app.py", line 1820, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/Users/api_user/miniconda3/envs/microsetta-private-api/lib/python3.8/site-packages/flask/_compat.py", line 39, in reraise
    raise value
  File "/Users/api_user/miniconda3/envs/microsetta-private-api/lib/python3.8/site-packages/flask/app.py", line 1949, in full_dispatch_request
    rv = self.dispatch_request()
  File "/Users/api_user/miniconda3/envs/microsetta-private-api/lib/python3.8/site-packages/flask/app.py", line 1935, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/Users/api_user/miniconda3/envs/microsetta-private-api/lib/python3.8/site-packages/connexion/decorators/decorator.py", line 48, in wrapper
    response = function(request)
  File "/Users/api_user/miniconda3/envs/microsetta-private-api/lib/python3.8/site-packages/connexion/decorators/security.py", line 327, in wrapper
    return function(request)
  File "/Users/api_user/miniconda3/envs/microsetta-private-api/lib/python3.8/site-packages/connexion/decorators/uri_parsing.py", line 144, in wrapper
    response = function(request)
  File "/Users/api_user/miniconda3/envs/microsetta-private-api/lib/python3.8/site-packages/connexion/decorators/validation.py", line 384, in wrapper
    return function(request)
[...]

PR #495, opened 2022-11-29
Fix account scrub bug

cassidysymons:
#492


cassidysymons:
Leaving this for the night - the fact that

PR #497, opened 2022-12-07
Functionality to allow interested users to opt out

cassidysymons:
For THDMI Japan, we'll be emailing people who signed up to join the cohort, giving them the opportunity to opt out if they've decided they don't want to receive a kit. This PR exposes new API endpoints and implements corresponding repo functionality to reflect a given interested_user deciding they no longer want to participate, using the confirm_consent column.


wasade (changes requested):
Very minor suggested edits, thanks!

PR #498, opened 2022-12-12
Resolve external survey account deletion bug.

cassidysymons:
https://github.com/biocore/microsetta-private-api/issues/492

This PR addresses a few related issues:
1) Fixes stale state for an account that we need to delete.
2) Changes the user-facing source deletion code to scrub, rather than delete, if they've taken external surveys. This is then leveraged by the admin-facing account deletion code to detect the external surveys that could be left dangling before this change.
3) Implemented a 422 response scenario for delete_source(). Prior to this code, it was declared as a possibility in the yaml file, but wasn't actually implemented in the API.

### DIFF
--- a/microsetta_private_api/api/_source.py
+++ b/microsetta_private_api/api/_source.py
@@ -4,11 +4,14 @@ from datetime import date
 from flask import jsonify
 
 from microsetta_private_api.api._account import _validate_account_access
-from microsetta_private_api.api.literals import SRC_NOT_FOUND_MSG
+from microsetta_private_api.api.literals import SRC_NOT_FOUND_MSG,\
+    SRC_NO_DELETE_MSG
 from microsetta_private_api.exceptions import RepoException
 from microsetta_private_api.model.source import Source, HumanInfo, NonHumanInfo
 from microsetta_private_api.repo.source_repo import SourceRepo
 from microsetta_private_api.repo.survey_answers_repo import SurveyAnswersRepo
+from microsetta_private_api.repo.survey_template_repo import SurveyTemplateRepo
+from microsetta_private_api.repo.sample_repo import SampleRepo
 from microsetta_private_api.repo.transaction import Transaction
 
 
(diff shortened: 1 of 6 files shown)


cassidysymons:
@wasade When you have a moment, can you please review this PR?

microsetta_private_api/api/_source.py
```diff
@@ -109,18 +112,44 @@ def delete_source(account_id, source_id, token_info):
...
+            )
+
+            if has_external:
+                for survey_id in answers:
+                    survey_answers_repo.scrub(
+                        account_id,
+                        source_id,
+                        survey_id
+                    )
+                if not source_repo.scrub(account_id, source_id):
```
wasade, line 140: Should this check happen on line 124 above the call to  `has_external_surveys`?

microsetta_private_api/api/_source.py
```diff
@@ -109,18 +112,44 @@ def delete_source(account_id, source_id, token_info):
...
+                    return jsonify(code=404, message=SRC_NOT_FOUND_MSG), 404
```
wasade, line 141: So I think we need to be careful with short circuit returns. If this return is hit, the calls to `scrub` above will not be committed

microsetta_private_api/api/_source.py
```diff
@@ -109,18 +112,44 @@ def delete_source(account_id, source_id, token_info):
...
+            else:
+                for survey_id in answers:
```
wasade, line 143:
```suggestion
                # If we reach this point, then the user does not have external surveys
                # nor do they have samples associated. Therefore it is safe to remove
                # their surveys and source entirely 
                for survey_id in answers:
```

wasade (commented):
So I _think_ there is a concern w/ the early returns and commit. Comment inline, if you agree I think the change is minor to make

microsetta_private_api/api/_source.py
cassidysymons: It's less a check and more of an in-line verification that the scrub function actually works, modeled after [the existing functionality to delete a source](https://github.com/biocore/microsetta-private-api/blob/master/microsetta_private_api/api/_source.py#L119).

From that standpoint (addressing the other comment), I think it makes sense to _not_ commit the scrub calls on the surveys, as we haven't successfully scrubbed the source.

We could avoid all of that by putting a check on line 124 that verifies the source meets the conditions to be deleted/scrubbed (i.e., the source exists and there's only a single record in the ag.source table with that ID). In that case, the scrub function could still fail if it's not a human source, but that would only happen if someone had somehow associated either a sample or external survey with a non-human source. And if that happened, I'd want it to show up in the exception logs to investigate because that would be rather concerning.

wasade: Sounds good. I'm +1 on path of least resistance which I think is to retain the current code?

cassidysymons: Yep, code as written would be the simplest path. If you don't see any other issues here, I'm going to merge this, https://github.com/biocore/microsetta-private-api/pull/497, and https://github.com/biocore/microsetta-interface/pull/264 and get them up on staging for testing.

wasade: Sounds good lets do this

PR #500, opened 2023-01-09
Fixes error in microsetta-interface integration test

charles-cowart:
Fixes an error found when running microsetta-interface integration tests.
https://github.com/biocore/microsetta-interface/pull/267

### DIFF
--- a/microsetta_private_api/api/_survey.py
+++ b/microsetta_private_api/api/_survey.py
@@ -291,7 +291,7 @@ def read_answered_surveys(account_id, source_id, language_tag, token_info):
             if template_id is None:
                 continue
             o = survey_template_repo.get_survey_template_link_info(template_id)
-            api_objs.append(o.to_api(ans, status))
+            api_objs.append(o.to_api(ans, status, None))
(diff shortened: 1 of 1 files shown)


charles-cowart:
@cassidysymons This small fix is needed to support the microsetta-interface integration test fix.

cassidysymons:
Is this a functional change just to resolve issues with tests? If so, that seems like an issue with the tests, rather than the code itself. Can you explain why this change is necessary?

charles-cowart:
This isn't a functional change. We added percentage_completed to the SurveyTemplateLinkInfo model class not long ago, per @wasade's suggestion to add it to the model. It's to_api() previously only required survey_answers_id and survey_status, but we added percentage_completed as an additional parameter to add and initialize that key. Depending on when the SurveyTemplateLinkInfo object is created, percentage_completed is set to None for a new object or set to a proper value in read_survey_template().

The additional parameter is needed or else the REST API call pointing to this method will fail.

cassidysymons:
Gotcha, re-read the code and it makes sense now. Thanks.

PR #512, opened 2023-03-24
bulk_scan barcode

raginirai553:


microsetta_private_api/db/patches/0115.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE barcodes.rack_samples (
```
cassidysymons, line 1: I would recommend adding a scan_id column to this table facilitate tracking the current state of a rack's contents. A scan_id would be constant for all samples contained in a given request to the /bulk_scan path on microsetta-admin.

microsetta_private_api/admin/tests/test_admin_repo.py
```diff
@@ -1324,6 +1324,36 @@ def test_set_kit_uuids_for_dak_order(self):
...
+
+        uuid = ""
+        with Transaction() as t:
+            barcode = '000001024'
+
+            admin_repo = AdminRepo(t)
+            uuid = admin_repo.map_to_rack(barcode, scan_info)
+
+        self.assertTrue(len(uuid) > 0)
+
```
cassidysymons, line 1356: Once you add the scan_id to the table, please add unit tests to confirm that get_rack_samples only returns the samples contained in the most recent scan.

microsetta_private_api/admin/admin_impl.py
```diff
@@ -917,3 +917,34 @@ def get_vioscreen_sample_to_user(token_info):
         st_repo = SurveyTemplateRepo(t)
         data = st_repo.get_vioscreen_sample_to_user()
     return jsonify(data), 200
+
+
+def map_to_rack(token_info, sample_barcode, body):
+    validate_admin_access(token_info)
+
+    print(str(body))
```
cassidysymons, line 925: Please make sure to remove debugging info from code.

microsetta_private_api/admin/admin_impl.py
```diff
@@ -917,3 +917,34 @@ def get_vioscreen_sample_to_user(token_info):
...
+        admin_repo = AdminRepo(t)
+        scan_id = admin_repo.map_to_rack(sample_barcode, body)
+        t.commit()
+
+    response = jsonify({"scan_id": scan_id})
+    response.status_code = 201
+    return response
+
+
+def get_sample_from_rack(token_info, rack_id):
```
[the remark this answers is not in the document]
raginirai553: Resolved the issue!

microsetta_private_api/db/patches/0115.sql
raginirai553: I modified the query to fetch most recent scanned samples using distinct clause on sample_id, sorted by date_time in descending order. Further, the filtration is done using rack_id in where clause of the said select query. Do we still need to make this change?

cassidysymons: That modification wouldn't work, as it won't capture the fact that samples can be removed from a rack without being replaced. For example:
1. You scan rack XYZ and it has samples A, B, and C on it.
2. A half hour later, you remove sample B, then scan the rack again - our system would still think it has all three samples.

So yes, we still need to add a scan_id to manage the state of a rack at any given scan.

cassidysymons:
@raginirai553 We just merged a big set of changes into master-overhaul. Can you please back-out the changes to `0114.sql` and rename `0115.sql` to `0116.sql` and update your working branch? Thanks!

PR #514, opened 2023-03-28
Pin flask-babel

cassidysymons:
@wasade I had pinned flask-babel in master-overhaul when this issue popped up in January, but not master. This should be a formality but can you merge this so that workflow can re-run on my other PR? (Going to open one on Interface momentarily as well)


wasade (approved):

wasade:
Good to merge once passing

PR #525, opened 2023-08-09
Overhaul Adjustments + Bugfix

cassidysymons:
This PR addresses three issues:
1) It changes our psycopg2 pool from `SimpleConnectionPool` to `ThreadedConnectionPool`, which is necessary as we'll be running Celery in threaded mode, rather than prefork.
2) It adds a `t.commit()` that was missing from the `check_shipping_updates()` task in util/perk_fulfillment.py.
3) It switches all perk fulfillment communication from the `payer_email` that Fundrazr returns to the `contact_email`, which is stored as` interested_users.email`.

### DIFF
--- a/microsetta_private_api/repo/transaction.py
+++ b/microsetta_private_api/repo/transaction.py
@@ -6,9 +6,9 @@ from psycopg2 import sql
 
 
 class Transaction:
-    # Note: SimpleConnectionPool works only for single threaded applications
-    #  Should we make the server multi threaded, we must switch to a
-    #  ThreadedConnectionPool
+    # Note: Using ThreadedConnectionPool as we've switched Celery to threaded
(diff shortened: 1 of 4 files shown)


microsetta_private_api/repo/transaction.py
```diff
@@ -9,7 +9,7 @@ class Transaction:
     # Note: SimpleConnectionPool works only for single threaded applications
     #  Should we make the server multi threaded, we must switch to a
```
wasade, line 10: :/

wasade (approved):

microsetta_private_api/repo/transaction.py
cassidysymons: Good catch, I'll update that comment before I merge. Thanks for the quick review.

PR #527, opened 2023-08-18
Limit Unfinished Vioscreen Sessions

cassidysymons:
This PR limits the `VioscreenSessionRepo.get_unfinished_sessions()` function to the past 30 days, relative either to the creation date (for criteria 1 within the function) or the last modified date (relative to criteria 2/3).

The impact on staging is that the number of unfinished sessions drops from 4344 to 3. Based on this substantial change, I'm going to defer on adding any `sleep` calls between sessions until/unless Viocare requests further changes.

Note: This PR is relative to overhaul, as I'd like to observe it running on staging for 48 hours before pushing an equivalent change to master.

### DIFF
--- a/ci/conda_requirements.txt
+++ b/ci/conda_requirements.txt
@@ -3,4 +3,4 @@ psycopg2
 flask
 natsort
 pycryptodome
-pandas
+pandas < 2.0.0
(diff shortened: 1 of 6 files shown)


wasade (approved):
That's super clean

cassidysymons:
> That's super clean

Thanks. Fixing unit tests before merging. Long-term, it would be good to adjust the unit tests to create new Vioscreen registry entries & sessions rather than depending on stable ones in the database, but that's not something I really want to tackle at the moment.

PR #528, opened 2023-08-22
Overhaul Adjustments Based on Testing

cassidysymons:
Notable changes:
- Expose the address error that Melissa Data returns to the user when they're correcting their address
- Survey question adjustments
- Rename the perk fulfillment account to appear more descriptively on shipping labels

### DIFF
--- a/microsetta_private_api/api/_interested_user.py
+++ b/microsetta_private_api/api/_interested_user.py
@@ -121,7 +121,7 @@ def get_interested_user_address_update(interested_user_id, email):
                         interested_user.country
                     )
 
-                    if isinstance(melissa_result, bool):
+                    if melissa_result is False:
                         # We shouldn't reach this point, but it means the
                         # address wasn't actually checked
                         return jsonify(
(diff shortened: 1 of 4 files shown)


microsetta_private_api/api/_interested_user.py
```diff
@@ -95,15 +96,46 @@ def get_interested_user_address_update(interested_user_id, email):
...
+                    melissa_repo = MelissaRepo(t)
+                    melissa_result = melissa_repo.check_duplicate(
+                        interested_user.address_1,
+                        interested_user.address_2,
+                        interested_user.address_3,
+                        interested_user.postal_code,
+                        interested_user.country
+                    )
+
+                    if isinstance(melissa_result, bool):
```
wasade, line 124:
```suggestion
                    if melissa_result is False:
```

The docstring and method only define `False` and at the moment it doesn't seem like there is a way to return `True`

wasade (approved):
Just a minor comment

PR #535, opened 2023-09-11
Addressing Feedback from Production Testing

cassidysymons:
- Change association between Fundrazr campaign and internal campaign to be TMI rather than AGP
- Change shipping speed from 2-day to Ground
- Change Rob's phone number in consent documents
- Survey question adjustments
- Fix height/weight unit layout on small devices
- Add plumbing for edge cases of Vioscreen sessions with neither a sample_id nor registration_code

### DIFF
--- a/microsetta_private_api/util/tests/test_google_geocoding.py
+++ b/microsetta_private_api/util/tests/test_google_geocoding.py
@@ -62,7 +62,7 @@ UCSD_GEOCODING_RESULTS = {
          "formatted_address": "9500 Gilman Dr, La Jolla, CA 92093, USA",
          "geometry": {
             "location": {
-               "lat": 32.8798977,
+               "lat": 32.8798916,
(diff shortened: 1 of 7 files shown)


microsetta_private_api/util/tests/test_google_geocoding.py
```diff
@@ -211,8 +211,8 @@ def test_geocode_address(self):
         obs_lat, obs_long, obs_state, obs_country, obs_error =\
            geocode_address(UCSD_ADDRESS)
 
-        self.assertEqual(obs_lat, 32.8798916)
-        self.assertEqual(obs_long, -117.2363115)
+        self.assertEqual(obs_lat, 32.8798977)
```
wasade, line 214: UCSD hasn't moved...? If there is flux at sig digits maybe test on round?

wasade (approved):

PR #536, opened 2023-09-19
Add Toggle for Perk Fulfillment to Settings

cassidysymons:
This pull request adds a new column to `ag.settings` - `perk_fulfillment_active`. When this column is True, the Celery jobs for fulfilling transactions will run as normal. When it's False, the jobs will abort immediately after checking the `perk_fulfillment_active` setting.

I've allowed the jobs that check for shipping updates and perks without fulfillment details to bypass the check, as there's no harm in allowing them to run even when perk fulfillment is off (and in some cases would actually be beneficial).


wasade:
Seems reasonable in the interests of time. If we end up toggling this more than once, then we may want to move to an API call to enable/disable, and expose on admin dashboard. But I think fine to defer that

PR #538, opened 2023-09-25
Address Verification Adjustments

cassidysymons:
The soft launch revealed some edge cases in both the address verification and address adjustment pieces that needed adjustment, along with supporting pieces that needed improvement:
- We've added AV14 as a good code, provided there are no other errors
- Address_3 has been removed from the duplicate checking function, as it was added for Spain and is not relevant to any planned or foreseeable functionality
- The way in which address_2 is handled for duplicate checking has been refined, as None/null values were causing certain records to be ignored erroneously
- Unit tests for MelissaRepo did not exist but have been added
- The send_email event for invalid addresses was not being logged, but is now

### DIFF
--- a/microsetta_private_api/repo/melissa_repo.py
+++ b/microsetta_private_api/repo/melissa_repo.py
@@ -71,7 +71,7 @@ class MelissaRepo(BaseRepo):
             # psycopg won't automatically handle = NULL vs. IS NULL
             sql = """SELECT * FROM campaign.melissa_address_queries
                                 WHERE (source_address_1 = %s
-                                AND source_{0} 
+                                AND source_{0}
                                 AND source_postal = %s
                                 AND source_country = %s
                                 AND result_processed = true)
--- a/microsetta_private_api/repo/melissa_repo.py
+++ b/microsetta_private_api/repo/melissa_repo.py
@@ -67,19 +67,36 @@ class MelissaRepo(BaseRepo):
         Full table row is record is a duplicate
         """
         with self._transaction.dict_cursor() as cur:
-            cur.execute("""SELECT * FROM campaign.melissa_address_queries
-                            WHERE (source_address_1 = %s
-                            AND source_address_2 = %s
-                            AND source_postal = %s
-                            AND source_country = %s
-                            AND result_processed = true)
-                            OR (result_address_1 = %s
-                            AND result_address_2 = %s
-                            AND result_postal = %s
-                            AND result_country = %s
-                            AND result_processed = true)""",
-                        (address_1, address_2, postal, country,
-                         address_1, address_2, postal, country))
+            # We need to gracefully handle a null/None value for address_2 as
+            # psycopg won't automatically handle = NULL vs. IS NULL
+            if address_2 is None:
+                cur.execute("""SELECT * FROM campaign.melissa_address_queries
+                                WHERE (source_address_1 = %s
+                                AND source_address_2 IS NULL
+                                AND source_postal = %s
+                                AND source_country = %s
+                                AND result_processed = true)
+                                OR (result_address_1 = %s
+                                AND result_address_2 IS NULL
+                                AND result_postal = %s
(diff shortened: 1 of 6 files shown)


microsetta_private_api/repo/melissa_repo.py
```diff
@@ -69,21 +67,36 @@ def check_duplicate(self, address_1, address_2, address_3, postal,
...
+                                AND source_postal = %s
+                                AND source_country = %s
+                                AND result_processed = true)
+                                OR (result_address_1 = %s
+                                AND result_address_2 = %s
+                                AND result_postal = %s
+                                AND result_country = %s
+                                AND result_processed = true)""",
+                            (address_1, address_2, postal, country,
+                             address_1, address_2, postal, country))
```
wasade, line 99:
```suggestion
            sql = """SELECT * FROM campaign.melissa_address_queries
                                WHERE (source_address_1 = %s
                                AND source_{0} 
                                AND source_postal = %s
                                AND source_country = %s
                                AND result_processed = true)
                                OR (result_address_1 = %s
                                AND result_{0}
                                AND result_postal = %s
                                AND result_country = %s
                                AND result_processed = true)"""
            if address_2 is None:
                address_2_is_null = 'address_2 IS NULL'
                sql = sql.format(address_2_is_null)
                arguments = (address_1, postal, country,
                             address_1, postal, country)
            else:
                address_2_is_not_null = 'address_2 = %s'
                sql = sql.format(address_2_is_not_null)
                arguments = (address_1, address_2, postal, country,
                             address_1, address_2, postal, country)
            cur.execute(sql, arguments)
```

The above would reduce some of the replication, and ensures the overall structure of the SQL query is consistent between branches. The `format` with the SQL query is safe in this context as the strings are effectively static. `arguments` could probably simplified further but I dont think it matters as much.

microsetta_private_api/util/melissa.py
```diff
@@ -97,13 +106,22 @@ def verify_address(address_1, address_2=None, address_3=None, city=None,
...
+                    if code[0:2] == "AE":
+                        r_errors_present = True
+                    if code in GOOD_CODES_NO_ERROR:
+                        r_good_conditional = True
                     if code in GOOD_CODES:
                         r_good = True
                         break
 
+                if not r_good and r_good_conditional and not r_errors_present:
+                    r_good = True
```
wasade, line 123: This statement reads: if not good ... then say we're good. Is that correct? If so then that's a bit confusing :)

If `r_good_conditional` is also set in the `if code in GOOD_CODES`, then I think this could be changed too:

```python
if r_good_conditional:
    if not r_errors_present:
        r_good = True
```

cassidysymons: It's essentially an alternative path to reaching a state of `r_good = True`. It could be simplified to just `if r_good_conditional and not r_errors_present:` but that has the potential to redundantly set `r_good = True` (which doesn't hurt anything, just feels a bit silly). I don't have strong feelings on the stylistic approach, I just wrote it to be abundantly clear that it's a conditional path to reaching a state of `r_good = True` since we're treating the code in `GOOD_CODES_NO_ERROR` differently than the ones in `GOOD_CODES`.

wasade: Makes sense. But my worry is looking back at this code a year from now, when something urgent is going on, where it's valuable to minimize the amount of time it takes to understand a conditional. If the concern is a redundant setting on `r_good` then that may be expressable as a comment. Sound reasonable?

cassidysymons: Definitely, will adjust in a few minutes.

cassidysymons: Adjusted and passing Lint now, let me know if there's anything else you'd like me to address before merging. Thanks!

wasade: Thanks! I'm +1

PR #540, opened 2023-10-03
Fix Perk Fulfillment for Orders with Quantity > 1

cassidysymons:


microsetta_private_api/repo/perk_fulfillment_repo.py
```diff
@@ -97,94 +97,70 @@ def process_pending_fulfillment(self, ftp_id):
...
-                        self._schedule_ffq(
-                            subscription_id,
-                            fulfillment_date,
-                            False
-                        )
-                    else:
                         error_info = self._fulfill_ffq(
                             row['ftp_id'],
-                            row['kit_quantity'],
+                            template,
```
wasade, line 122: It is technically possible to take a path through the code where dereferencing `template` would trigger a `NameError`. Can a default be set outside the loop?

microsetta_private_api/repo/perk_fulfillment_repo.py
```diff
@@ -97,94 +97,70 @@ def process_pending_fulfillment(self, ftp_id):
...
+                            fulfillment_date =\
+                                self._future_fulfillment_date(
+                                    row['fulfillment_spacing_number'],
+                                    row['fulfillment_spacing_unit'],
+                                    x
+                                )
+                            self._schedule_ffq(
+                                subscription_id,
+                                fulfillment_date,
+                                False
```
wasade, line 146: Could the associated kwarg be used here? e.g., `foo=False`

microsetta_private_api/repo/perk_fulfillment_repo.py
```diff
@@ -197,6 +173,31 @@ def process_pending_fulfillment(self, ftp_id):
...
+                            fulfillment_date =\
+                                self._future_fulfillment_date(
+                                    row['fulfillment_spacing_number'],
+                                    row['fulfillment_spacing_unit'],
+                                    x
+                                )
+                            self._schedule_kit(subscription_id,
+                                               fulfillment_date,
+                                               row['dak_article_code'],
+                                               False)
```
wasade, line 187: Same here

microsetta_private_api/repo/perk_fulfillment_repo.py
```diff
@@ -197,6 +173,31 @@ def process_pending_fulfillment(self, ftp_id):
...
+                        else:
+                            status, return_val = self._fulfill_kit(
+                                row,
+                                1,
```
wasade, line 191: a comment following this would be nice, or setting a static variable outside the loop

microsetta_private_api/repo/perk_fulfillment_repo.py
cassidysymons: I'm not quite following here, and I think the way that GitHub's interface is calculating the changes and displaying them might be causing an issue. Here's the code in question, with context:
```python
                    # If there are any FFQs attached to the perk, immediately
                    # fulfill the first one
                    if row['ffq_quantity'] > 0:
                        # If the perk is a kit or subscription, send thank you
                        # email with kit content. Otherwise, send thank you
                        # for FFQ only
                        if row['kit_quantity'] > 0:
                            template = "thank_you_with_kit"
                        else:
                            template = "thank_you_no_kit"

                        error_info = self._fulfill_ffq(
                            row['ftp_id'],
                            template,
                            row['email'],
                            row['first_name'],
                            subscription_id
                        )
``` 

So the `if` statement sets it to one value or the other immediately before it's used in `self._fulfill_ffq()`. Does that same appropriate or am I missing a path where it can be de-referenced?

microsetta_private_api/repo/perk_fulfillment_repo.py
cassidysymons: Will be adjusted in next commit.

microsetta_private_api/repo/perk_fulfillment_repo.py
cassidysymons: Will be adjusted in next commit.

microsetta_private_api/repo/perk_fulfillment_repo.py
cassidysymons: Will be adjusted in next commit.

microsetta_private_api/repo/perk_fulfillment_repo.py
wasade: you're right, sorry, read through too quickly

cassidysymons:
@wasade I rebuilt the database on staging from last night's backup and deployed this patch to confirm it works as intended. It runs without error and the states of `campaign.transaction` and `campaign.fundrazr_transaction_perk` both look correct to me.

wasade:
Excellent! Im +1 if green

PR #544, opened 2023-11-15
Add Language Tag/Locale to Reconsent Check

cassidysymons:
When the reconsent process was originally set up, it was assumed that whenever an update to consent agreements was necessary, it would be done for all languages at once. However, this was not the case in reality, as the Japanese documents were updated after the fact.

Based on this reality, the code that checks whether a user needs to reconsent needed to be made aware of the user's locale. This PR contains the code that does that, as well as updated unit tests to observe that it works as intended. No corresponding updates to microsetta-interface are necessary.

### DIFF
--- a/microsetta_private_api/api/_consent.py
+++ b/microsetta_private_api/api/_consent.py
@@ -44,17 +44,21 @@ def get_consent_doc(account_id, consent_id, token_info):
         return jsonify(code=404, message=CONSENT_DOC_NOT_FOUND_MSG), 404
 
 
-def check_consent_signature(account_id, source_id, consent_type, language_tag, token_info):
+def check_consent_signature(
+        account_id, source_id, consent_type, language_tag, token_info
+):
     _validate_account_access(token_info, account_id)
 
     with Transaction() as t:
         consent_repo = ConsentRepo(t)
-        res = consent_repo.is_consent_required(source_id, consent_type, language_tag)
+        res = consent_repo.is_consent_required(
+            source_id, consent_type, language_tag
+        )
 
     return jsonify({"result": res}), 200
 
 
-def sign_consent_doc(account_id, source_id, consent_type, body, token_info):
+def sign_consent_doc(account_id, source_id, body, token_info):
     _validate_account_access(token_info, account_id)
 
     with Transaction() as t:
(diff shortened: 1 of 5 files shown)


microsetta_private_api/api/_consent.py
```diff
@@ -44,12 +44,16 @@ def get_consent_doc(account_id, consent_id, token_info):
         return jsonify(code=404, message=CONSENT_DOC_NOT_FOUND_MSG), 404
 
 
-def check_consent_signature(account_id, source_id, consent_type, token_info):
+def check_consent_signature(
```
wasade, line 47: why not pull the language tag from the users account entry?

wasade: ...actually, it seems like a bad idea to allow the user to alter their language tag to sign a consent which may not be relevant to them?

wasade (commented):
The locale of the signed consent is implicitly stored based on the documents they sign, but it seems like a user can "game" the system if they just change the locale up either on the api call or in their account details. Is it possible for the same user to sign multiple versions of the same consent by changing their locale, and if so, does that have consequences?

cassidysymons:
@wasade Given the fact that we've limited the ja_JP locale to Japanese residents, and the English/Spanish consents are equivalent in content, do you still have concerns about this from a conceptual standpoint?

wasade:
Equivalent right now but not assured in the future. The concern w/ "Is it possible for the same user to sign multiple versions of the same consent by changing their locale, and if so, does that have consequences?" does I think still hold

cassidysymons:
It would be possible for someone to sign both the English and Spanish documents (either/both Spanish locales), yes. As far as consequences, @sejsong, are you aware of any IRB issues if a user signs multiple translations of the same consent document?

wasade:
It would be good for the code to verify document versions explicitly, rather than implicitly

sejsong:
No issues since the English and Spanish versions have exactly the same content.

wasade:
I'm not concerned about today, I'm concerned about the future

cassidysymons:
Adding explicit versioning to the consent documents wouldn't be a huge
task. Considering all of the consequences of different locales being on
different versions would be a much bigger conversation. Would you agree
that refactoring the current code to explicit versioning but punting on
dealing with all of the consequences is acceptable for now?

wasade:
Sounds good thanks!

cassidysymons:
Closing this PR as it's been replaced by https://github.com/biocore/microsetta-private-api/pull/548

PR #548, opened 2023-11-29
Explicit Consent Versioning

cassidysymons:
This PR implements explicit consent versioning and refactors the `is_consent_required` function accordingly.

Importantly, we're still operating under the assumption that all locales and age groups will be operating on the same version of the consent documents. If that changes at some point in the future, it will likely require significant changes to microsetta-private-api, microsetta-interface, and microsetta-admin.

### DIFF
--- a/microsetta_private_api/repo/consent_repo.py
+++ b/microsetta_private_api/repo/consent_repo.py
@@ -134,20 +134,20 @@ class ConsentRepo(BaseRepo):
         bool
             True if the user needs to reconsent, False otherwise
         """
+        if age_range == "18-plus":
+            consent_type = "adult_" + consent_type
+        elif age_range == "13-17":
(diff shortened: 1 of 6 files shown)


microsetta_private_api/repo/consent_repo.py
```diff
@@ -113,36 +117,60 @@ def get_consent_document(self, consent_id):
...
+            The source's age range
+        consent_type : str
+            data or biospecimen
+
+        Returns
+        -------
+        bool
+            True if the user needs to reconsent, False otherwise
+        """
         with self._transaction.dict_cursor() as cur:
```
wasade, line 137:
```suggestion
```

wasade (approved):
Just a minor suggested change

PR #551, opened 2023-12-01
Activate Subscription Fulfillment

cassidysymons:
Since we're not going to allow users to alter their subscriptions, the only code change for activating subscription fulfillment is flipping on the Celery task. I re-reviewed the perk fulfillment code and accompanying unit tests this afternoon, but if you have any outstanding questions or concerns about subscription fulfillment, this PR would be a good time to voice them.


cassidysymons:
@wasade Do you have any outstanding questions or concerns before we flip the celery tasks for subscription fulfillment on?

PR #555, opened 2023-12-11
Added admin FFQ code lookup

colts661:


microsetta_private_api/repo/tests/test_perk_fulfillment_repo.py
```diff
@@ -1173,6 +1174,261 @@ def _count_fundrazr_ffq_codes(self, t):
...
+                "INSERT INTO campaign.interested_users " + \
+                "(campaign_id, first_name, last_name, email) VALUES " + \
+                ", ".join([
+                    f"('{self.test_campaign_id1}', 'First', 'Last', '{email}')" 
+                    for email in emails
+                ]) + \
+                "RETURNING interested_user_id"
+            )
+            iu_ids = [tup[0] for tup in cur.fetchall()]
+            t.commit()
```
wasade, line 1194: Can these state changes be rolled back either in `tearDown` or can the test be fully executed w/o commiting the transaction?

microsetta_private_api/repo/tests/test_perk_fulfillment_repo.py
```diff
@@ -1173,6 +1174,261 @@ def _count_fundrazr_ffq_codes(self, t):
...
+            cur.execute(
+                "INSERT INTO campaign.transaction VALUES " + \
+                ", ".join([
+                    f"('{tx_id}', '{iu_id}', 'fundrazr', '4Tqx5', " + \
+                    "'2023-01-01', 100, 100, 'usd', 'First', 'Last', " + \
+                    "'fake@bar.com', 'paypal', 'coolcool', TRUE)"
+                    for tx_id, iu_id in tx_ids.items()
+                ])
+            )
+            t.commit()
```
wasade, line 1225: same here

microsetta_private_api/repo/tests/test_perk_fulfillment_repo.py
```diff
@@ -1173,6 +1174,261 @@ def _count_fundrazr_ffq_codes(self, t):
...
+                    f"('{tx_id}', '3QeVd', 1, TRUE)" for tx_id in tx_ids
+                    if tx_id != 'MT3'
+                ]) + \
+                "RETURNING id"
+            )
+            ftp_ids = {
+                tup[0]: tx_id 
+                for tup, tx_id in zip(cur.fetchall(), tx_ids)
+            }
+            t.commit()
```
wasade, line 1260: same here

wasade: ...I'll stop noting :)

microsetta_private_api/repo/perk_fulfillment_repo.py
```diff
@@ -793,3 +793,57 @@ def check_perk_fulfillment_active(self):
             )
             row = cur.fetchone()
             return row['perk_fulfillment_active']
+    
+    def get_ffq_codes_by_email(self, email):
+        email = "%" + email + "%"
+        with self._transaction.dict_cursor() as cur:
+            # Note: Use left join to differentiate email not found possibility
+            cur.execute(
```
wasade, line 801: This is very clean

microsetta_private_api/repo/tests/test_perk_fulfillment_repo.py
colts661: I was hesitant to do so because:
1. These operations are naive and does not conform to normal workflow such as when an ftp/ffq code will be created but directly inserting information related to this query. Running other operations will probably trigger lots of logic problems;
2. Some tests were in-between the insertions for convenience, but I could reformat them to not be affected by one insertion;
3. I tried to contain all insertion and deletion in the same test function so that running that function won't affect other workflows

wasade: It's standard practice in this codebase, and others under the biocore organization, to not leave lingering state changes in databases during testing. Short term, it might simplify a test. Long term, it makes reasoning about what's going on difficult as the state of the entity being tested is changing. It can also make repeat runs of the same tests, without repopulating the test database in between, impossible

There are a few different ways it's managed in this codebase depending on the situation, see e.g., `repo/source.py`. I'm not sure what strategy will be best here. It might be to use exact identifiers on population, and have `tearDown` remove from the corresponding tables which is practical but leaves room for missing a few things and possibly some maintenance burden. Best is if this can be done where a `.rollback()` can just be issued, but that is not possible in all scenarios

colts661: Thanks for the information! Yes I agree that we should try to place all database state changes in `setUp` and `tearDown`, so is it okay if the inserted values have logic errors? For example, if a side effect of inserting a ftp involves populating another table and this never happens. We could make sure to only use these values for the purpose of this query.

wasade: Yes, exactly. If state is always reset, then it's fine to put garbage in to verify behavior. 

Test execution is out-of-order, so any state changes which occur can have side effects on other tests. This can lead to difficult to debug/understand scenarios when bad data are intentionally placed into the database.

microsetta_private_api/repo/perk_fulfillment_repo.py
cassidysymons: +1, impressive query

cassidysymons (approved):

cassidysymons:
Going to hold on merging this until the corresponding Interface PR is finalized. Great work @colts661 !

PR #558, opened 2024-01-29
added last_update column to ag_kit_barcodes

ayobi:
Added last_update column to ag_kit_barcodes table. This will enable a more descriptive sample status on the 'My Reports' tab. See [PR](https://github.com/biocore/microsetta-interface/pull/311) for microsetta-interface for FE changes.

### DIFF
--- a/microsetta_private_api/db/patches/0135.sql
+++ b/microsetta_private_api/db/patches/0135.sql
@@ -1,7 +1,9 @@
 -- Jan 26, 2024
--- Add sample metadata update timestamp
+-- Add sample information update timestamp
+-- For UI enhancement. 
+-- We are not nulling the field out, though it may become necessary if used outside of the UI.
 ALTER TABLE ag.ag_kit_barcodes
-    ADD COLUMN last_update timestamp with time zone;
+    ADD COLUMN latest_sample_information_update timestamp with time zone;
 
-COMMENT ON COLUMN ag.ag_kit_barcodes.last_update
-    IS 'Sample metadata update timestamp';
+COMMENT ON COLUMN ag.ag_kit_barcodes.latest_sample_information_update
+    IS 'Sample information update timestamp.';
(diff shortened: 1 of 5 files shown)


microsetta_private_api/model/sample.py
```diff
@@ -51,7 +52,7 @@ def from_db(cls, sample_id, date_collected, time_collected,
             datetime_collected = datetime.combine(date_collected,
                                                   time_collected)
         return cls(sample_id, datetime_collected, site, notes, barcode,
-                   latest_scan_timestamp, source_id,
+                   latest_scan_timestamp, last_update, source_id,
                    account_id, sample_projects, latest_scan_status)
 
     def to_api(self):
```
cassidysymons, line 58: Since this PR is changing what the API returns for a sample, can you please update the corresponding [component](https://github.com/biocore/microsetta-private-api/blob/master/microsetta_private_api/api/microsetta_private_api.yaml#L3069) and [schema](https://github.com/biocore/microsetta-private-api/blob/master/microsetta_private_api/api/microsetta_private_api.yaml#L3301) entries in the [YAML file](https://github.com/biocore/microsetta-private-api/blob/master/microsetta_private_api/api/microsetta_private_api.yaml)?

cassidysymons (changes requested):

microsetta_private_api/db/patches/0135.sql
```diff
@@ -0,0 +1,7 @@
+-- Jan 26, 2024
+-- Add sample metadata update timestamp
+ALTER TABLE ag.ag_kit_barcodes
+    ADD COLUMN last_update timestamp with time zone;
+
+COMMENT ON COLUMN ag.ag_kit_barcodes.last_update
+    IS 'Sample metadata update timestamp';
```
wasade, line 7: As a field name, `last_update` is ambiguous. Can the field reflect the type of update it corresponds too?

cassidysymons: Let's change this field name to `latest_sample_information_update` to clearly state that it's only tracking the info that belongs directly to a sample (e.g. collection date/time and type) and is ignoring other associations like metadata.

Also, we do not need to null the column when a sample is deleted, but it would be beneficial to add a comment to the sample deletion code noting this decision and suggesting it be revisited if we ever use the new column for anything other than its current purpose of UI enhancement.

PR #561, opened 2024-02-06
Added database constraint on AuthRocket

### LINKED ISSUE #559: Database constraint on auth_issuer and auth_sub columns
In the `ag.account` table, the `auth_issuer` and `auth_sub` columns should either both be null or both be non-null. However, there's not currently a constraint on the database to reflect this. I believe it would be something like `CONSTRAINT auth_state_coherent ((auth_issuer != null and auth_sub != null) or (auth_issuer is null and auth_sub is null))` but that syntax needs to be verified, and it would be good to review the code around the AuthRocket handoffs to ensure there aren't any secondary effects we'd regret.

colts661:
Fixed #559. The constraint error is handled in both `create_account()` and `update_account()`, and so the tests are built upon these two methods without committing in the tests.


colts661:
[The 422 code test](https://github.com/biocore/microsetta-private-api/blob/2a6644a8b59e0cbc18b80b421cf35d47db658dcb/microsetta_private_api/api/tests/test_api.py#L898) in `test_api` for `create_account()` directly addresses this issue but seems to have some extra legacy handling afterwards. Please let me know:
1. What's expected in this test case after the addition of db constraint;
2. Should I move all added tests to `api/tests/test_api.py` altogether.

Thank you!!

cassidysymons:
@colts661 You can delete lines 898 - 911 in `test_api.py`. Since the new database constraint makes the condition impossible, it's no longer necessary to test what the API does if it happens.

I'll review the rest of the PR in the next couple of hours. Thanks!

cassidysymons:
@colts661 When you have a moment, can you please rename `0135.sql` to `0136.sql` in your working branch? Another PR merge created a conflict. Thanks!

cassidysymons:
@colts661 Can you please update your working branch from `biocore/microsetta-private-api:master`? Patch `0135.sql` isn't in your working branch and is causing the build to fail when I test it on our staging server.

Issue #424, opened 2022-03-09
Include campaign.interested_users in account delete/scrub functionality


cassidysymons:
Once infrastructure is established to advance an interested user to an account - presumably a join table - that link will need to be used to delete or scrub the interested user record as appropriate when a user requests that we delete their account and all personally identifiable information.

PR #564, opened 2024-02-15
Scrub interested users

ayobi:
Once the account scrub is done we want to also scrub the campaign.interested_users entry as well per https://github.com/biocore/microsetta-private-api/issues/424. Currently, this uses the account email address and finds the entry in the interested_users table.


microsetta_private_api/repo/interested_user_repo.py
```diff
@@ -207,3 +207,41 @@ def verify_address(self, interested_user_id):
...
+            interested_user.latitude = 0
+            interested_user.longitude = 0
+            interested_user.confirm_consent = False
+            interested_user.ip_address = "scrubbed"
+            interested_user.address_checked = False
+            interested_user.address_valid = False
+            interested_user.residential_address = False
+
+            if self.update_interested_user(interested_user) == 1:
+                return True
```
cassidysymons, line 246: I believe that this will break the for loop that begins on line 225, meaning that only one record in `interested_users` will actually get scrubbed. Can you take a more defensive approach, such as creating an `errors` variable with a value of `False`, setting it to `True` if any records fail to update, then `return errors`?

microsetta_private_api/repo/interested_user_repo.py
```diff
@@ -207,3 +207,41 @@ def verify_address(self, interested_user_id):
                         (interested_user_id,)
                     )
                     return False
+
+    def get_interested_user_by_just_email(self, email):
+        email = "%" + email + "%"
+        with self._transaction.dict_cursor() as cur:
+            cur.execute(
+                "SELECT * FROM campaign.interested_users WHERE email LIKE %s",
```
cassidysymons, line 215: This should check email based on case-insensitive equivalency, not a LIKE match. If we're deleting john@hotmail.com from our database, this code as written would inappropriately scrub bigjohn@hotmail.com and littlejohn@hotmail.com from the `interested_users` table.

microsetta_private_api/admin/admin_impl.py
```diff
@@ -872,6 +872,8 @@ def delete_account(account_id, token_info):
 
         acct_repo.scrub(account_id)
 
+        interested_users_repo.scrub(acct.email)
```
cassidysymons, line 875: The `interested_users_repo.scrub()` function can return False if one or more of the records fails to scrub, but its return value is ignored here. Ideally, it would raise an exception so that we don't end up with inconsistent state in the database, where the account has been scrubbed but records in `interested_users` have not.

cassidysymons (changes requested):
Along with the specific comments in the most recent commit, can you please add unit tests that exercise the relevant functions for both successful and failed outcomes to ensure that we're handling each outcome appropriately?

Also, is there any way to cleanly isolate the most recent commit and remove the commits related to the deletion queue from this branch? If not, it's not a huge deal, but it muddies the water on comparing changes to have the same things showing up on multiple PRs.

microsetta_private_api/repo/interested_user_repo.py
ayobi: Bah, you're right, thanks. I'll fix that.

microsetta_private_api/admin/admin_impl.py
ayobi: @cassidysymons I see, at the moment I see a user can update their email address, but it doesn't look like it also updates their respective address on the interested_users table https://github.com/biocore/microsetta-interface/blob/621691c21223e1889b514c571ea7317372d94dc1/microsetta_interface/implementation.py#L1055C1-L1085C56 so there are probably users that have updated their email that wouldn't be on the table.

ayobi:
> Along with the specific comments in the most recent commit, can you please add unit tests that exercise the relevant functions for both successful and failed outcomes to ensure that we're handling each outcome appropriately?
> 
> Also, is there any way to cleanly isolate the most recent commit and remove the commits related to the deletion queue from this branch? If not, it's not a huge deal, but it muddies the water on comparing changes to have the same things showing up on multiple PRs.

@cassidysymons I've decided to close this PR and make a new one (with most of the improvements) to make it easier for migration 👍

PR #565, opened 2024-02-17
scrub interested users

ayobi:
Once the account scrub is done we want to also scrub the campaign.interested_users entry as well per https://github.com/biocore/microsetta-private-api/issues/424. Currently, this uses the account email address and finds the entry in the interested_users table.

### DIFF
--- a/microsetta_private_api/repo/interested_user_repo.py
+++ b/microsetta_private_api/repo/interested_user_repo.py
@@ -207,3 +207,45 @@ class InterestedUserRepo(BaseRepo):
                         (interested_user_id,)
                     )
                     return False
+
+    def get_interested_user_by_just_email(self, email):
+        # email = "%" + email + "%"
+        with self._transaction.dict_cursor() as cur:
+            cur.execute(
+                "SELECT * FROM campaign.interested_users "
+                "WHERE lower(email) = lower(%s)",
+                (email,)
+            )
+            rs = cur.fetchall()
+            return [__class__._row_to_interested_user(r) for r in rs]
+
+    def scrub(self, interested_user_email):
+        interested_users = self. \
+            get_interested_user_by_just_email(interested_user_email)
+        updated = False
+
+        for interested_user in interested_users:
+            interested_user.first_name = "scrubbed"
+            interested_user.last_name = "scrubbed"
+            interested_user.email = "scrubbed"
+            interested_user.phone = "scrubbed"
+            interested_user.address_1 = "scrubbed"
+            interested_user.address_2 = "scrubbed"
+            interested_user.address_3 = "scrubbed"
+            interested_user.city = "scrubbed"
+            interested_user.state = "scrubbed"
+            interested_user.postal_code = "scrubbed"
+            interested_user.country = "scrubbed"
+            interested_user.latitude = 0
+            interested_user.longitude = 0
+            interested_user.confirm_consent = False
+            interested_user.ip_address = "scrubbed"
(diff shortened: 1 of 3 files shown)


microsetta_private_api/repo/interested_user_repo.py
```diff
@@ -207,3 +207,45 @@ def verify_address(self, interested_user_id):
                         (interested_user_id,)
                     )
                     return False
+
+    def get_interested_user_by_just_email(self, email):
+        # email = "%" + email + "%"
```
cassidysymons, line 212: Can this comment be removed?

microsetta_private_api/repo/interested_user_repo.py
```diff
@@ -207,3 +207,45 @@ def verify_address(self, interested_user_id):
...
+            interested_user.email = "scrubbed"
+            interested_user.phone = "scrubbed"
+            interested_user.address_1 = "scrubbed"
+            interested_user.address_2 = "scrubbed"
+            interested_user.address_3 = "scrubbed"
+            interested_user.city = "scrubbed"
+            interested_user.state = "scrubbed"
+            interested_user.postal_code = "scrubbed"
+            interested_user.country = "scrubbed"
+            interested_user.latitude = 0
```
cassidysymons, line 239: Both latitude and longitude are nullable columns - since 0 is technically a valid value for both lat and long, it seems like setting them to null would be a bit more appropriate than 0 for a scrubbed record.

microsetta_private_api/repo/interested_user_repo.py
```diff
@@ -207,3 +207,45 @@ def verify_address(self, interested_user_id):
...
+            interested_user.ip_address = "scrubbed"
+            interested_user.address_checked = False
+            interested_user.address_valid = False
+            interested_user.residential_address = False
+
+            if self.update_interested_user(interested_user) == 1:
+                updated = True
+            else:
+                updated = False
+        return updated
```
cassidysymons, line 251: The scrub function returns a value but `admin_impl.delete_account()` doesn't do anything with it. If any of the update operations fail, I think we'd want to throw an Exception so we don't end up in a situation where part of the scrub/delete operation succeeded and part didn't.

cassidysymons (changes requested):

microsetta_private_api/repo/interested_user_repo.py
```diff
@@ -207,3 +207,42 @@ def verify_address(self, interested_user_id):
...
+            interested_user.confirm_consent = False
+            interested_user.ip_address = "scrubbed"
+            interested_user.address_checked = False
+            interested_user.address_valid = False
+            interested_user.residential_address = False
+
+            if self.update_interested_user(interested_user) == 1:
+                pass
+            else:
+                raise RepoException("Error scrubbing interested user")
```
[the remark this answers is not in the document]
ayobi: Ah yes, thank you

cassidysymons:
@ayobi Ran into one more issue while testing this on staging. The [update_interested_user](https://github.com/biocore/microsetta-private-api/blob/master/microsetta_private_api/repo/interested_user_repo.py#L54) function doesn't touch the `latitude`, `longitude`, or `confirm_consent` columns so those aren't being scrubbed. Can you add those columns to `update_interested_user` and I'll put it back up on staging for one more round of testing? Thanks.

ayobi:
Ah good catch, sorry about that, they're added now 👍

cassidysymons (approved):

PR #566, opened 2024-03-04
Fix sample summary project output

colts661:
Fixed biocore/microsetta-admin#97


cassidysymons:
@colts661 When we address code that isn't working as intended, we like to add tests that verify the issue has been resolved and can be verified through automated methods. In this scenario, I think two tests would suffice to reflect that the adjustments are working as intended:
1) When a user searches by project, assert that all samples returned include that project in the output. I'd recommend using a project ID in the database that has a relatively limited number of samples associated.
2) When a user searches for a single barcode, all of its project associations are returned. Please use a barcode from the database that has at least two samples associated.

Other than that, looks good. Thanks!

cassidysymons (approved):

PR #567, opened 2024-03-08
create barcodes for existing kits

ayobi:
Taking from task description, this PR covers part 1:

Add a barcode to an existing kit (new feature). In this scenario, there already exists a kit in the database with one or more barcodes. The user will log into microsetta-admin and add barcodes to one or more kits, and it needs to provide the following options:

- Add barcodes to a single kit. The user would enter a Kit ID into a text field, then have the option to either enter a barcode to add to the kit OR dynamically generate a new barcode to add to the kit (using the existing barcode generation function). If it's a generated barcode, the output should inform the user of the newly created barcode. Otherwise, the output should be a success/failure message.

- Add barcodes to several kits. The user would upload a CSV file with Kit IDs in the first column. They would then have the option to either add a dynamically generated barcode to each kit OR put a barcode in the second column of the CSV file. In either case, the tool should return a CSV file with the Kit IDs and the newly added barcodes.

### DIFF
--- a/microsetta_private_api/admin/admin_impl.py
+++ b/microsetta_private_api/admin/admin_impl.py
@@ -268,7 +268,8 @@ def create_kits(body, token_info):
     return jsonify(kits), 201
 
 
-def generate_barcodes(body):
+def generate_barcodes(body, token_info):
+    validate_admin_access(token_info)
 
     number_of_kits = body['number_of_kits']
     number_of_samples = body['number_of_samples']
@@ -281,7 +282,8 @@ def generate_barcodes(body):
     return barcode
 
 
-def insert_barcodes(body):
+def insert_barcodes(body, token_info):
+    validate_admin_access(token_info)
 
     barcode = body['barcodes']
     project_id = [body['project_id']]
--- a/microsetta_private_api/admin/admin_impl.py
+++ b/microsetta_private_api/admin/admin_impl.py
@@ -56,7 +56,7 @@ def search_kit_id(token_info, kit_id):
         admin_repo = AdminRepo(t)
         diag = admin_repo.retrieve_diagnostics_by_kit_id(kit_id)
         if diag is None:
-            return jsonify(code=404, message="Kit ID not found"), 404
+            return jsonify(code=404, message=f'Kit ID {kit_id} not found'), 404
         return jsonify(diag), 200
 
 
@@ -67,7 +67,7 @@ def kit_id_exists(token_info, kit_id):
         admin_repo = AdminRepo(t)
         diag = admin_repo.check_kit_id_exists(kit_id)
         if diag is None:
-            return jsonify(code=404, message="Kit ID not found"), 404
+            return jsonify(code=404, message=f'Kit ID {kit_id} not found'), 404
         return jsonify(diag), 200
 
 
@@ -304,10 +304,12 @@ def handle_barcodes(body, token_info):
     if isinstance(kit_ids, str):
         kit_ids = [kit_ids]
 
-    for kit_id in kit_ids:
-        response, code = kit_id_exists(token_info, kit_id)
-        if code == 404:
-            return jsonify(code=404, message='Kit ID not found'), 404
+    with Transaction() as t:
+        admin_repo = AdminRepo(t)
+        for kit_id in kit_ids:
+            diag = admin_repo.check_kit_id_exists(kit_id)
+            if diag is None:
+                return jsonify(f'Kit ID {kit_id} not found'), 404
 
     action = body.get('action')
 
@@ -349,6 +351,13 @@ def insert_barcodes(body, token_info):
     # Ensure kit_ids is a list, even if it's a single value
     if isinstance(kit_ids, str):
         kit_ids = [kit_ids]
+
+    with Transaction() as t:
+        admin_repo = AdminRepo(t)
+        for barcode in barcodes:
+            diag = admin_repo.check_barcode_exists(barcode)
+            if diag is not None:
+                return jsonify(f'Barcode {barcode} already exists'), 404
 
     # Check if the lengths match
     if len(kit_ids) != len(barcodes):
--- a/microsetta_private_api/admin/admin_impl.py
+++ b/microsetta_private_api/admin/admin_impl.py
@@ -275,8 +275,6 @@ def create_kits(body, token_info):
     kit_prefix = body.get('kit_id_prefix', None)
     project_ids = body['project_ids']
     user_barcodes = body.get('user_barcodes', [])
-    remaining_barcodes_to_generate = body.get(
-        'remaining_samples_to_generate', [])
 
     with Transaction() as t:
         admin_repo = AdminRepo(t)
@@ -286,7 +284,6 @@ def create_kits(body, token_info):
                                           number_of_samples,
                                           kit_prefix,
                                           user_barcodes,
-                                          remaining_barcodes_to_generate,
                                           project_ids)
         except KeyError:
             return jsonify(code=422, message="Unable to create kits"), 422
--- a/microsetta_private_api/admin/admin_impl.py
+++ b/microsetta_private_api/admin/admin_impl.py
@@ -286,7 +286,8 @@ def generate_barcodes(body, token_info):
     return barcode
 
 
-def insert_barcodes(body):
+def insert_barcodes(body, token_info):
+    validate_admin_access(token_info)
 
     barcode = body['barcodes']
     project_id = [body['project_id']]
--- a/microsetta_private_api/admin/admin_impl.py
+++ b/microsetta_private_api/admin/admin_impl.py
@@ -268,6 +268,31 @@ def create_kits(body, token_info):
     return jsonify(kits), 201
 
 
+def generate_barcodes(body):
+
+    number_of_kits = body['number_of_kits']
+    number_of_samples = body['number_of_samples']
+
+    with Transaction() as t:
+        admin_repo = AdminRepo(t)
+        barcode = admin_repo._generate_novel_barcodes_admin(
+            number_of_kits, number_of_samples)
+        t.commit()
+    return barcode
+
+
+def insert_barcodes(body):
+
+    barcode = body['barcodes']
+    project_id = [body['project_id']]
+
+    with Transaction() as t:
+        admin_repo = AdminRepo(t)
+        admin_repo._insert_barcodes_to_existing_kit(barcode, project_id)
+        t.commit()
+    return '', 204
+
+
 def get_account_events(account_id, token_info):
     validate_admin_access(token_info)
 
--- a/microsetta_private_api/admin/admin_impl.py
+++ b/microsetta_private_api/admin/admin_impl.py
@@ -253,13 +253,17 @@ def create_kits(body, token_info):
     number_of_samples = body['number_of_samples']
     kit_prefix = body.get('kit_id_prefix', None)
     project_ids = body['project_ids']
+    user_barcodes = body.get('user_barcodes', [])
 
     with Transaction() as t:
         admin_repo = AdminRepo(t)
 
         try:
-            kits = admin_repo.create_kits(number_of_kits, number_of_samples,
-                                          kit_prefix, project_ids)
+            kits = admin_repo.create_kits(number_of_kits,
+                                          number_of_samples,
+                                          kit_prefix,
+                                          user_barcodes,
+                                          project_ids)
         except KeyError:
             return jsonify(code=422, message="Unable to create kits"), 422
         else:
--- a/microsetta_private_api/admin/admin_impl.py
+++ b/microsetta_private_api/admin/admin_impl.py
@@ -288,7 +288,7 @@ def generate_barcodes(body, token_info):
 
 def insert_barcodes(body, token_info):
     validate_admin_access(token_info)
-    
+
     barcode = body['barcodes']
     project_id = [body['project_id']]
 
--- a/microsetta_private_api/admin/admin_impl.py
+++ b/microsetta_private_api/admin/admin_impl.py
@@ -60,17 +60,6 @@ def search_kit_id(token_info, kit_id):
         return jsonify(diag), 200
 
 
-def kit_id_exists(token_info, kit_id):
-    validate_admin_access(token_info)
-
-    with Transaction() as t:
-        admin_repo = AdminRepo(t)
-        diag = admin_repo.check_kit_id_exists(kit_id)
-        if diag is None:
-            return jsonify(code=404, message=f'Kit ID {kit_id} not found'), 404
-        return jsonify(diag), 200
-
-
 def search_email(token_info, email):
     validate_admin_access(token_info)
 
@@ -302,9 +291,9 @@ def handle_barcodes(body, token_info):
         kit_ids = [kit_ids]
 
     with Transaction() as t:
-        admin_repo = AdminRepo(t)
+        kit_repo = KitRepo(t)
         for kit_id in kit_ids:
-            diag = admin_repo.check_kit_id_exists(kit_id)
+            diag = kit_repo.get_kit_all_samples(kit_id)
             if diag is None:
                 return jsonify(f'Kit ID {kit_id} not found'), 404
 
@@ -330,7 +319,7 @@ def handle_barcodes(body, token_info):
 
             t.commit()
 
-        return jsonify(barcode[1])
+        return jsonify(barcode[1]), 201
 
     elif action == 'insert':
         return insert_barcodes(body, token_info)
@@ -352,14 +341,14 @@ def insert_barcodes(body, token_info):
     with Transaction() as t:
         admin_repo = AdminRepo(t)
         for barcode in barcodes:
-            diag = admin_repo.check_barcode_exists(barcode)
+            diag = admin_repo.retrieve_diagnostics_by_barcode(barcode)
             if diag is not None:
                 return jsonify(f'Barcode {barcode} already exists'), 404
 
     # Check if the lengths match
     if len(kit_ids) != len(barcodes):
-        raise ValueError("The number of kit IDs must "
-                         "match the number of barcodes")
+        return jsonify("The number of kit IDs must "
+                       "match the number of barcodes"), 400
 
     with Transaction() as t:
         admin_repo = AdminRepo(t)
--- a/microsetta_private_api/admin/admin_impl.py
+++ b/microsetta_private_api/admin/admin_impl.py
@@ -286,8 +286,7 @@ def generate_barcodes(body, token_info):
     return barcode
 
 
-def insert_barcodes(body, token_info):
-    validate_admin_access(token_info)
+def insert_barcodes(body):
 
     barcode = body['barcodes']
     project_id = [body['project_id']]
--- a/microsetta_private_api/admin/admin_impl.py
+++ b/microsetta_private_api/admin/admin_impl.py
@@ -272,29 +272,64 @@ def create_kits(body, token_info):
     return jsonify(kits), 201
 
 
-def generate_barcodes(body, token_info):
-    validate_admin_access(token_info)
+def handle_barcodes(body):
+    action = body.get('action')
 
-    number_of_kits = body['number_of_kits']
-    number_of_samples = body['number_of_samples']
+    if action == 'create':
+        return generate_barcodes(body)
+    elif action == 'insert':
+        return insert_barcodes(body)
+    else:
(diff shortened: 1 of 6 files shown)


ayobi:
Added part 2:

The user can now provide their own barcode(s) if needed when they're creating a kit. This can be done either by providing the barcode(s) in the field(s) OR they can upload a list of barcodes via a csv file.

microsetta_private_api/api/microsetta_private_api.yaml
```diff
@@ -2781,6 +2781,65 @@ paths:
...
+      responses:
+        '201':
+          description: Barcodes were successfully inserted
+          content:
+            application/json:
+              schema:
+                type: array
+        '500':
+          description: Duplicate barcodes found
+
```
cassidysymons, line 2842: Let's collapse these down to a single API path since they're both doing the same thing - adding barcode(s) to kit(s). Whether it's user-provided or system-generated barcodes can be a part of the request.

microsetta_private_api/repo/admin_repo.py
```diff
@@ -851,6 +859,80 @@ def _generate_novel_barcodes(self, number_of_kits, number_of_samples,
 
         return kit_name_and_barcode_tuples_list, new_barcodes
 
+    def _generate_novel_barcodes_admin(self,
```
cassidysymons, line 862: Is there a reason you added this function rather than using the `_generate_novel_barcodes` function that already exists?

cassidysymons (changes requested):
I suspect the UI changes in microsetta-admin combined with collapsing the two API paths into one will introduce significant changes, so let's knock those out, then I'll do another round of code review.

microsetta_private_api/repo/admin_repo.py
ayobi: `_generate_novel_barcodes` is more complex and tailored for generating barcodes tied to specific kit names, whereas `_generate_novel_barcodes_admin `is simpler and used for generating barcode(s) without associating them with kit names. I suppose I felt that separating the functions allows for more modularity, maintainability, and clarity in case things scale with this feature in the future.

cassidysymons: Is there an actual use case for generating barcodes and not associating them with a Kit ID? The corresponding PR in `microsetta-admin` ends up doing precisely that, first [requesting novel barcodes](https://github.com/biocore/microsetta-admin/pull/99/files#diff-b23ea711ab49d62652d71b16d5097513712c6df215296d1803f1f56982aace04R597), then [associating them with kits](https://github.com/biocore/microsetta-admin/pull/99/files#diff-b23ea711ab49d62652d71b16d5097513712c6df215296d1803f1f56982aace04R654).

Unless there's a reason we'd ever create barcodes and not associate them with kits (which I don't believe there is), this function is creating bloat and reducing maintainability.

wasade: The data model for a microsetta project shouldn't allow this given the FK between `ag_kit` and `ag_kit_barcodes`. 

The value of a kit ID outside of a microsetta project is uncertain. But, they're easy to generate and disregard (whereas generating after the fact would not be fun)

ayobi: Ah, I see. Thanks 👍

cassidysymons:
@ayobi Can you please take a look at the conflict on this PR? I think you'll need to merge biocore:master into your branch to pick up new changes, and then possibly re-adjust the conflicting function. Thanks.

microsetta_private_api/admin/admin_impl.py
```diff
@@ -278,6 +282,68 @@ def create_kits(body, token_info):
...
+        if body['generate_barcode_single'] == 'on':
+            number_of_kits = 1
+            number_of_samples = 1
+    else:
+        number_of_kits = len(body['kit_ids'])
+        number_of_samples = 1
+
+    with Transaction() as t:
+        admin_repo = AdminRepo(t)
+        barcode = admin_repo._generate_novel_barcodes(
```
cassidysymons, line 307: Per the comments that Daniel and I left on the PR earlier, generating barcodes that are not associated with a kit_name shouldn't be allowed. I'm leaving this comment once, but there are changes throughout the PR that need to be adjusted (here, tests, the `_generate_novel_barcodes` function in `admin_repo.py`).

microsetta_private_api/repo/admin_repo.py
```diff
@@ -884,15 +892,74 @@ def _generate_novel_barcodes(self, number_of_kits, number_of_samples,
...
+            else:
+                return new_barcodes
 
-            kit_name_and_barcode_tuples_list = []
-            barcode_offset = range(0, total_barcodes, number_of_samples)
-            for offset, name in zip(barcode_offset, kit_names):
-                for i in range(number_of_samples):
-                    kit_name_and_barcode_tuples_list.append(
-                        (name, new_barcodes[offset + i]))
+    def _insert_barcodes_to_existing_kit(self,
```
cassidysymons, line 907: Given that there's no upstream check that the project_ids for all of the kits being updated are identical, it's unsafe to make that assumption here. The best course of action is to check the project IDs on a per-kit basis within this function.

cassidysymons (changes requested):

microsetta_private_api/repo/admin_repo.py
cassidysymons: @ayobi Was this comment addressed? Or am I missing a reason that it's not relevant?

microsetta_private_api/admin/admin_impl.py
```diff
@@ -278,6 +282,88 @@ def create_kits(body, token_info):
...
+        number_of_samples = 1
+
+    with Transaction() as t:
+        admin_repo = AdminRepo(t)
+
+        kit_names = admin_repo._generate_novel_kit_names(
+            number_of_kits, kit_prefix=None)
+
+        barcode = admin_repo._generate_novel_barcodes(
+            number_of_kits, number_of_samples, kit_names)
```
cassidysymons, line 319: Doesn't this essentially duplicate the existing code to create kits?

cassidysymons (changes requested):
Please let me know if you have any questions about isolated issues and/or this task in broad strokes. Each round of review has been a combination of steps forward and backwards and I'm concerned that it's not moving towards a polished and accurate addition to the codebase.

microsetta_private_api/repo/admin_repo.py
ayobi: If I understood correctly I set the check here https://github.com/biocore/microsetta-private-api/blob/224e5fe7b64c773fbfab8f55ab2b425f6d5aeb2c/microsetta_private_api/repo/admin_repo.py#L926-L927

microsetta_private_api/admin/admin_impl.py
ayobi: Ah shoot, my apologies. Tunnel vision.

microsetta_private_api/repo/admin_repo.py
cassidysymons: It's not working as intended from two different standpoints:

1) I have a test kit that has multiple barcodes, each of which is associated with two projects. I added a single barcode to this test kit. The newly created barcode is only associated with one of the projects, when it should be associated with both.
2) I have two test kits, call them KitA and KitB. KitA has one barcode that is associated with Project A. KitB has multiple barcodes, all of which are associated with Projects B and C. I used the Add Barcodes to Multiple Kits feature, uploaded a spreadsheet with KitA and KitB, and received a spreadsheet with newly added barcodes for both kits. However, neither of those barcodes exist in the database.

There's no practical reason to prevent admins from adding barcodes to kits from multiple projects in one transaction. Since the code to associate new barcodes with the right projects needs to be refactored, I'd suggest dropping the validation and just doing all of the project ID logic (read existing association/create new association) on a per-kit basis.

microsetta_private_api/admin/admin_impl.py
ayobi: On second look @cassidysymons, it does not. The `create_kits` function in the `admin_impl ` uses `_generate_novel_barcodes` in the `admin_repo`. https://github.com/biocore/microsetta-private-api/blob/224e5fe7b64c773fbfab8f55ab2b425f6d5aeb2c/microsetta_private_api/repo/admin_repo.py#L808-L818

This function, `generate_barcodes`, is used by `handle_barcodes` in the `admin_impl` which has its isolated use of adding barcodes to existing kits.
https://github.com/biocore/microsetta-private-api/blob/224e5fe7b64c773fbfab8f55ab2b425f6d5aeb2c/microsetta_private_api/admin/admin_impl.py#L285-L322

cassidysymons: Both functions generate novel kit names, then generate novel barcodes. The new code in `generate_barcodes` is replicating what happens in `create_kits`. Is that not duplicating code?

ayobi: Ah I see what you mean. Sorry, I thought you meant if it gets called twice resulting in duplicated kit names and barcodes.

ayobi: Would you rather me modify the `create_kits ` function accordingly to use in the `handle_barcodes` to avoid the duplication?

microsetta_private_api/repo/admin_repo.py
```diff
@@ -801,15 +802,71 @@ def create_kits(self, number_of_kits, number_of_samples, kit_prefix,
             Number of samples that each kit will contain
         kit_prefix : str or None
             A prefix to put on to the kit IDs, this is optional.
+        user_barcodes : list of str
```
cassidysymons, line 805: Given the fact that the position of barcodes in provided CSVs matters, this should be something with more structure than a list.

cassidysymons (changes requested):

microsetta_private_api/repo/admin_repo.py
```diff
@@ -801,18 +801,93 @@ def create_kits(self, number_of_kits, number_of_samples, kit_prefix,
...
+
+            all_barcodes_per_slot.append(slot_barcodes)
+
+        kit_name_and_barcode_tuples_list = []
+        for kit_index, kit_name in enumerate(kit_names):
+            for slot_index in range(number_of_samples):
+                barcode = all_barcodes_per_slot[slot_index][kit_index]
+                kit_name_and_barcode_tuples_list.append(
+                    (kit_name, barcode)
+                )
```
cassidysymons, line 876: This feels fairly convoluted and possibly driven more by the earlier implementation than the understanding of sample slots. Given the intent of sample slots, couldn't this be simplified to something like (spitballing concepts, not actual code):
```
for i in range(number_of_samples):
    if i in range(len(user_barcodes)):
        slot[i] = user_barcodes[i]
    else:
        slot[i] = _generate_novel_barcodes()
````

There's code needed to set up the tuples at each slot, but it feels like there's a lot of room to make this code more readable and maintainable.

File: microsetta_private_api/repo/admin_repo.py [hunk 8]
```diff
@@ -892,7 +967,67 @@ def _generate_novel_barcodes(self, number_of_kits, number_of_samples,
...
+                try:
+                    cur.executemany("INSERT INTO ag_kit_barcodes "
+                                    "(ag_kit_id, barcode) "
+                                    "SELECT ag_kit_id, %s "
+                                    "FROM ag_kit "
+                                    "WHERE supplied_kit_id = %s",
+                                    [(barcode, kit_id)
+                                     for kit_id, barcode
+                                     in kit_barcodes_insert])
+                except Exception as e:
```
cassidysymons, line 1029: Did you encounter any exceptions here while developing this? It seems concerning that there's an expectation of a potential error here.

microsetta_private_api/repo/admin_repo.py
```diff
@@ -924,9 +1059,17 @@ def _create_kits(self, kit_names, new_barcodes,
...
+                    all(isinstance(item, list) for item in new_barcodes):
+                for barcodes in new_barcodes:
+                    for barcode in barcodes:
+                        for prj_id in project_ids:
+                            barcode_projects.append((barcode, prj_id))
+            else:
+                for barcode in new_barcodes:
+                    for prj_id in project_ids:
+                        barcode_projects.append((barcode, prj_id))
+            print("barcode projects", barcode_projects)
```
cassidysymons, line 1072:
```suggestion
```

microsetta_private_api/repo/admin_repo.py
```diff
@@ -1023,19 +1166,48 @@ def create_kit(self, kit_name, box_id, address_dict,
...
         kit_details = [{KIT_BOX_ID_KEY: box_id,
                        KIT_ADDRESS_KEY: address,
                        KIT_OUTBOUND_KEY: outbound_fedex_code,
                        KIT_INBOUND_KEY: inbound_fedex_code}]
         kit_name_and_barcode_tuples_list = \
             [(kit_name, x) for x in barcodes_list]
+        kit_names = [kit_name]
+
+        print("create kit barcode list", barcodes_list)
+        print("create kit tuple", kit_name_and_barcode_tuples_list)
```
cassidysymons, line 1179:
```suggestion
```

cassidysymons (changes requested):
Please let me know if you have any questions about what I've flagged here.

microsetta_private_api/admin/admin_impl.py
```diff
@@ -56,7 +56,18 @@ def search_kit_id(token_info, kit_id):
...
+
+def kit_id_exists(token_info, kit_id):
+    validate_admin_access(token_info)
+
+    with Transaction() as t:
+        admin_repo = AdminRepo(t)
+        diag = admin_repo.check_kit_id_exists(kit_id)
+        if diag is None:
+            return jsonify(code=404, message=f'Kit ID {kit_id} not found'), 404
         return jsonify(diag), 200
```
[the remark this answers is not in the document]
ayobi: Ah yes, deleting.

microsetta_private_api/repo/admin_repo.py
```diff
@@ -801,18 +801,93 @@ def create_kits(self, number_of_kits, number_of_samples, kit_prefix,
             Number of samples that each kit will contain
         kit_prefix : str or None
             A prefix to put on to the kit IDs, this is optional.
+        user_barcodes : list of str
```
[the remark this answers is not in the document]
ayobi: Ah yes, should say list of lists of strings

microsetta_private_api/api/microsetta_private_api.yaml
```diff
@@ -2805,6 +2805,43 @@ paths:
...
+                  type: array
+                  items:
+                    type: string
+                  description: Barcodes to insert (required if action is 'insert')
+                kit_id:
+                  type: string
+                  description: Kit ID to associate the barcodes with (required if action is 'insert')
+              required:
+                - action
+      responses:
```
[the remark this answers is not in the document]
ayobi: Ah yes, I'll update

microsetta_private_api/admin/admin_impl.py
```diff
@@ -278,6 +293,93 @@ def create_kits(body, token_info):
...
+    with Transaction() as t:
+        admin_repo = AdminRepo(t)
+        for barcode in barcodes:
+            diag = admin_repo.check_barcode_exists(barcode)
+            if diag is not None:
+                return jsonify(f'Barcode {barcode} already exists'), 404
+
+    # Check if the lengths match
+    if len(kit_ids) != len(barcodes):
+        raise ValueError("The number of kit IDs must "
```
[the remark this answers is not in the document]
ayobi: Ah, yes. Thanks👍

microsetta_private_api/admin/admin_impl.py
```diff
@@ -278,6 +293,93 @@ def create_kits(body, token_info):
...
+    barcodes = body['barcodes']
+
+    # Ensure kit_ids is a list, even if it's a single value
+    if isinstance(kit_ids, str):
+        kit_ids = [kit_ids]
+
+    with Transaction() as t:
+        admin_repo = AdminRepo(t)
+        for barcode in barcodes:
+            diag = admin_repo.check_barcode_exists(barcode)
```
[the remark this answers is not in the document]
ayobi: It was taking a while when I would upload a list of barcodes so I decided to create the function. Testing it again though, it's now checking quite quickly. I did reinstall my local db recently so maybe that was it. I'll switch to using `retrieve_diagnostics_by_barcode` again. 👍

microsetta_private_api/admin/admin_impl.py
```diff
@@ -278,6 +293,93 @@ def create_kits(body, token_info):
...
+
+        with Transaction() as t:
+            admin_repo = AdminRepo(t)
+
+            barcode = admin_repo._generate_novel_barcodes(
+                number_of_kits, number_of_samples, kit_names=kit_ids)
+
+            t.commit()
+
+        return jsonify(barcode[1])
```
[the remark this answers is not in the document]
ayobi: Ah, I should add `201` there.

microsetta_private_api/admin/admin_impl.py
```diff
@@ -278,6 +293,93 @@ def create_kits(body, token_info):
...
+
+    if action == 'create':
+        if 'generate_barcode_single' in body:
+            if body['generate_barcode_single'] == 'on':
+                number_of_kits = 1
+                number_of_samples = 1
+        elif 'generate_barcodes' in body:
+            number_of_kits = (body['num_kits'])
+            number_of_samples = (body['num_samples'])
+        else:
```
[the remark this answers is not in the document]
ayobi: `generate_barcodes` is for creating kits, as you can see, it has `num_kits` and `num_samples` in the body
`generate_barcodes_multiple` is for adding barcodes to existing kits, it has just `kit_ids` in the body so I just check the length as one kit id = one barcode.
I could make it clearer in the naming of the payload to say something like `generate_barcodes_for_kits`, `generate_barcode_single_for_existing_kits`, and `generate_barcodes_multiple_for_existing_kits`?

microsetta_private_api/admin/admin_impl.py
```diff
@@ -278,6 +293,93 @@ def create_kits(body, token_info):
...
+        for kit_id in kit_ids:
+            diag = admin_repo.check_kit_id_exists(kit_id)
+            if diag is None:
+                return jsonify(f'Kit ID {kit_id} not found'), 404
+
+    action = body.get('action')
+
+    if action == 'create':
+        if 'generate_barcode_single' in body:
+            if body['generate_barcode_single'] == 'on':
```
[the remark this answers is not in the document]
ayobi: Yeah, I don't think so, I'll just put  `if generate_barcode_single` then.

microsetta_private_api/admin/admin_impl.py
```diff
@@ -278,6 +293,93 @@ def create_kits(body, token_info):
...
+
+    kit_ids = body['kit_ids']
+
+    if isinstance(kit_ids, str):
+        kit_ids = [kit_ids]
+
+    with Transaction() as t:
+        admin_repo = AdminRepo(t)
+        for kit_id in kit_ids:
+            diag = admin_repo.check_kit_id_exists(kit_id)
```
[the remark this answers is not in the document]
ayobi: Ah, I didn't see that function. No issue using that instead. 👍

microsetta_private_api/repo/admin_repo.py
ayobi: No, I think that was lingering from when I was testing something. Removing.

microsetta_private_api/admin/sample_summary.py
```diff
@@ -83,6 +83,7 @@ def per_sample(project, barcodes, strip_sampleid):
                 ffq_complete, ffq_taken, _ = vs_repo.get_ffq_status_by_sample(
                     sample.id
                 )
+                print("ffq complete", ffq_complete, "ffq taken", ffq_taken)
```
cassidysymons, line 86:
```suggestion
```

cassidysymons (changes requested):
Along with the noted comment, there are three issues identified through user testing:
1) When I tried to Create a Kit using barcodes that already existed, I received an error of "Error message not found in the results."
2) When I click Add Barcode to Kit in the menu, click the Add Barcode to Single Kit button, enter a valid Kit ID, and choose Generate barcode, the output displays a barcode in red, but that barcode does not exist in the database.
3) When I click Add Barcode to Kit in the menu, click the Add Barcode to Multiple Kits button, upload a list of valid Kit IDs, and choose Generate barcodes, the output displays a single barcode in red, which does not exist in the database.

PR #570, opened 2024-04-17
survey association lock change

ayobi:
Moved the point at which sample-survey associations are locked from the frontend microsetta-interface to immediately before the sample’s metadata is pushed up to Qiita on the backend.

### DIFF
--- a/microsetta_private_api/repo/qiita_repo.py
+++ b/microsetta_private_api/repo/qiita_repo.py
@@ -9,25 +9,24 @@ from microsetta_private_api.repo.survey_answers_repo import SurveyAnswersRepo
 class QiitaRepo(BaseRepo):
     def lock_completed_surveys_to_barcodes(self, barcodes):
         # lock survey-sample association
-        with self._transaction as t:
-            admin_repo = AdminRepo(t)
-            sar_repo = SurveyAnswersRepo(t)
+        admin_repo = AdminRepo(self._transaction)
+        sar_repo = SurveyAnswersRepo(self._transaction)
 
-            for sample_barcode in barcodes:
-                ids = admin_repo._get_ids_relevant_to_barcode(sample_barcode)
+        for sample_barcode in barcodes:
+            ids = admin_repo._get_ids_relevant_to_barcode(sample_barcode)
 
-                if ids is not None:
-                    account_id = ids.get('account_id')
-                    source_id = ids.get('source_id')
-                    sample_id = ids.get('sample_id')
+            if ids is not None:
+                account_id = ids.get('account_id')
+                source_id = ids.get('source_id')
+                sample_id = ids.get('sample_id')
 
-                    survey_ids = sar_repo.list_answered_surveys(
-                        account_id, source_id)
+                survey_ids = sar_repo.list_answered_surveys(
+                    account_id, source_id)
 
-                    if survey_ids is not None:
-                        for survey_id in survey_ids:
-                            sar_repo.associate_answered_survey_with_sample(
-                                account_id, source_id, sample_id, survey_id)
+                if survey_ids is not None:
+                    for survey_id in survey_ids:
+                        sar_repo.associate_answered_survey_with_sample(
+                            account_id, source_id, sample_id, survey_id)
 
     def push_metadata_to_qiita(self, barcodes=None):
         """Attempt to format and push metadata for the set of barcodes
--- a/microsetta_private_api/repo/qiita_repo.py
+++ b/microsetta_private_api/repo/qiita_repo.py
@@ -61,8 +61,6 @@ class QiitaRepo(BaseRepo):
         list
             Any error detail when constructing metadata
         """
-        # lock survey-sample association
-        self.lock_sample_to_survey(barcodes)
 
         if barcodes is None:
             with self._transaction.cursor() as cur:
@@ -93,6 +91,9 @@ class QiitaRepo(BaseRepo):
                 barcodes = {r[0] for r in cur.fetchall()}
         else:
             barcodes = set(barcodes)
+
+        # lock survey-sample association
+        self.lock_sample_to_survey(barcodes)
 
         # determine what samples are already known in qiita
         samples_in_qiita = set(qclient.get('/api/v1/study/10317/samples'))
--- a/microsetta_private_api/repo/qiita_repo.py
+++ b/microsetta_private_api/repo/qiita_repo.py
@@ -51,7 +51,7 @@ class QiitaRepo(BaseRepo):
                     source_id = ids.get('source_id')
                     sample_id = ids.get('sample_id')
 
-                    survey_ids = sar_repo.list_answered_surveys_by_sample(
+                    survey_ids = sar_repo.list_answered_surveys(
                         account_id, source_id, sample_id)
 
                     for survey_id in survey_ids:
--- a/microsetta_private_api/repo/qiita_repo.py
+++ b/microsetta_private_api/repo/qiita_repo.py
@@ -1,7 +1,10 @@
+from microsetta_private_api.repo.admin_repo import AdminRepo
 from microsetta_private_api.repo.base_repo import BaseRepo
 from microsetta_private_api.qiita import qclient
 from microsetta_private_api.repo.metadata_repo import retrieve_metadata
 from microsetta_private_api.repo.metadata_repo._constants import MISSING_VALUE
+from microsetta_private_api.repo.survey_answers_repo import SurveyAnswersRepo
+from microsetta_private_api.repo.transaction import Transaction
 
 
 class QiitaRepo(BaseRepo):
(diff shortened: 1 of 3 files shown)


microsetta_private_api/repo/qiita_repo.py
```diff
@@ -1,10 +1,40 @@
...
 from microsetta_private_api.repo.base_repo import BaseRepo
 from microsetta_private_api.qiita import qclient
 from microsetta_private_api.repo.metadata_repo import retrieve_metadata
 from microsetta_private_api.repo.metadata_repo._constants import MISSING_VALUE
+from microsetta_private_api.repo.survey_answers_repo import SurveyAnswersRepo
+from microsetta_private_api.repo.transaction import Transaction
 
 
 class QiitaRepo(BaseRepo):
+    def lock_sample_to_survey(self, barcodes):
```
cassidysymons, line 12: In the interest of keeping function names as descriptive and obvious as possible, I'd recommend changing this to `lock_completed_surveys_to_barcodes`. The current name makes it sound like a 1:1 relationship.

microsetta_private_api/repo/qiita_repo.py
```diff
@@ -1,10 +1,40 @@
...
+
+                    survey_ids = sar_repo.list_answered_surveys(
+                        account_id, source_id)
+
+                    if survey_ids is not None:
+                        for survey_id in survey_ids:
+                            sar_repo.associate_answered_survey_with_sample(
+                                account_id, source_id, sample_id, survey_id)
+                    else:
+                        raise RepoException("Survey IDs not found for barcode")
```
cassidysymons, line 34:
```suggestion
```
This shouldn't be an exception. It's possible for users to claim a sample without completing any surveys, and while it's not ideal, it shouldn't be treated as an error.

microsetta_private_api/repo/qiita_repo.py
```diff
@@ -65,6 +96,9 @@ def push_metadata_to_qiita(self, barcodes=None):
         else:
             barcodes = set(barcodes)
 
+        # lock survey-sample association
+        self.lock_sample_to_survey(barcodes)
```
cassidysymons, line 100: Looking at this code more closely, we should move this to happen after `to_push = list(barcodes - samples_in_qiita)[:1000]` as we only want to update the locks for samples actually being pushed up. Leaving it here would mean we're changing the associations for samples that are already pushed to Qiita.

cassidysymons (changes requested):

microsetta_private_api/repo/qiita_repo.py
```diff
@@ -84,6 +112,9 @@ def push_metadata_to_qiita(self, barcodes=None):
         # calls to this function if and as needed.
         to_push = list(barcodes - samples_in_qiita)[:1000]
 
+        # lock survey-sample association
+        self.lock_completed_surveys_to_barcodes(barcodes)
```
cassidysymons, line 116:
```suggestion
        self.lock_completed_surveys_to_barcodes(to_push)
```

Need to limit it to only the barcodes actually being pushed to Qiita.

microsetta_private_api/repo/tests/test_qiita.py
```diff
@@ -69,6 +70,65 @@ def test_push_metadata_to_qiita(self, test_retrieve_metadata,
                                       "associated with any surveys "
                                       "matching this template id")}])
 
+    def test_lock_sample_to_survey(self):
```
cassidysymons, line 73:
```suggestion
    def test_lock_completed_surveys_to_barcodes(self):
```

microsetta_private_api/repo/tests/test_qiita.py
```diff
@@ -69,6 +70,65 @@ def test_push_metadata_to_qiita(self, test_retrieve_metadata,
...
+                '148': 'Unspecified',
+                '492': 'Unspecified',
+                '493': 'Unspecified',
+                '502': 'Male'
+            }
+            sar.submit_answered_survey(
+                account_id,
+                source_id,
+                'en_US', 10, survey_10)
+            t.commit()
```
cassidysymons, line 118: Is there a reason you're committing the transaction here? We try to avoid committing in unit tests, and when it's strictly necessary, we need to remove the record(s) from the database in the `tearDown()` function.

cassidysymons (changes requested):

microsetta_private_api/repo/tests/test_qiita.py
ayobi: Hi @cassidysymons, since I need to submit a survey for the barcode in order to proceed with locking the barcode to the survey I am committing so that I can ensure after that there was indeed an addition to the database table. Should we maybe move out this test to someone else since it's not really testing qiita?

microsetta_private_api/repo/qiita_repo.py
```diff
@@ -1,10 +1,37 @@
...
 from microsetta_private_api.repo.metadata_repo import retrieve_metadata
 from microsetta_private_api.repo.metadata_repo._constants import MISSING_VALUE
+from microsetta_private_api.repo.survey_answers_repo import SurveyAnswersRepo
+from microsetta_private_api.repo.transaction import Transaction
 
 
 class QiitaRepo(BaseRepo):
+    def lock_completed_surveys_to_barcodes(self, barcodes):
+        # lock survey-sample association
+        with Transaction() as t:
```
cassidysymons, line 13:
```suggestion
        with self._transaction as t:
```

microsetta_private_api/repo/qiita_repo.py
```diff
@@ -1,10 +1,37 @@
...
+
+                    survey_ids = sar_repo.list_answered_surveys(
+                        account_id, source_id)
+
+                    if survey_ids is not None:
+                        for survey_id in survey_ids:
+                            sar_repo.associate_answered_survey_with_sample(
+                                account_id, source_id, sample_id, survey_id)
+
+            t.commit()
```
cassidysymons, line 33:
```suggestion
```

microsetta_private_api/repo/qiita_repo.py
```diff
@@ -1,10 +1,37 @@
+from microsetta_private_api.repo.admin_repo import AdminRepo
 from microsetta_private_api.repo.base_repo import BaseRepo
 from microsetta_private_api.qiita import qclient
 from microsetta_private_api.repo.metadata_repo import retrieve_metadata
 from microsetta_private_api.repo.metadata_repo._constants import MISSING_VALUE
+from microsetta_private_api.repo.survey_answers_repo import SurveyAnswersRepo
+from microsetta_private_api.repo.transaction import Transaction
```
cassidysymons, line 7:
```suggestion
```

microsetta_private_api/repo/tests/test_qiita.py
cassidysymons: Sorry, I wasn't considering that you were opening a new transaction in the lock function. It's considerably cleaner to keep everything in one transaction since there's no need to worry about a race condition occurring. With everything now using a single transaction, you can run the entire test using one transaction as well.

cassidysymons (changes requested):

microsetta_private_api/repo/tests/test_qiita.py
ayobi: With those changes suggested I get `psycopg2.pool.PoolError: trying to put unkeyed connection`

cassidysymons: Can you push the code up as it stands now so I can see it in context?

ayobi: Pushed it up 👍

microsetta_private_api/repo/tests/test_qiita.py
```diff
@@ -69,6 +70,64 @@ def test_push_metadata_to_qiita(self, test_retrieve_metadata,
...
+
+            # now lock the barcode to the survey that was recently submitted
+            qiita_repo = QiitaRepo(t)
+            qiita_repo.lock_completed_surveys_to_barcodes(test_barcodes)
+
+            with t.dict_cursor() as cur:
+                cur.execute("SELECT * FROM source_barcodes_surveys "
+                            "WHERE barcode = '000069747'")
+                rows_after = cur.fetchall()
+            t.commit()
```
cassidysymons, line 127:
```suggestion
```

cassidysymons (changes requested):

microsetta_private_api/repo/tests/test_qiita.py
cassidysymons: Any luck debugging the issue this morning?

ayobi: No luck as of yet

microsetta_private_api/repo/qiita_repo.py
```diff
@@ -1,10 +1,34 @@
...
 from microsetta_private_api.repo.metadata_repo._constants import MISSING_VALUE
+from microsetta_private_api.repo.survey_answers_repo import SurveyAnswersRepo
 
 
 class QiitaRepo(BaseRepo):
+    def lock_completed_surveys_to_barcodes(self, barcodes):
+        # lock survey-sample association
+        with self._transaction as t:
+            admin_repo = AdminRepo(t)
+            sar_repo = SurveyAnswersRepo(t)
```
cassidysymons, line 14:
```suggestion
        admin_repo = AdminRepo(self._transaction)
        sar_repo = SurveyAnswersRepo(self._transaction)
```

Fixed the indenting on the rest of the function accordingly and it should work.

ayobi: That did it, thank you

microsetta_private_api/tasks.py
```diff
@@ -86,3 +86,5 @@ def update_qiita_metadata():
                        {"what": "qiita metadata push errors",
                         "content": json.dumps(error, indent=2)},
                        EN_US)
+        else:
+            t.commit()
```
cassidysymons, line 90:
```suggestion
        t.commit()
```

PR #574, opened 2024-05-13
vocab for sample issues

ayobi:
Adding a fixed vocabulary of potential sample issues to the Scan Barcode tool.


microsetta_private_api/db/patches/0139.sql
```diff
@@ -0,0 +1,28 @@
+-- May 13, 2024
+-- Create two tables for observation categories and observation errors list
+-- Add observations column to barcode_scans table
+CREATE TABLE observation_categories (
```
cassidysymons, line 4:
```suggestion
CREATE TABLE sample_observation_categories (
```

Adding "sample_" as a prefix to these table names will help make their purpose abundantly clear to anyone jumping in later.

microsetta_private_api/db/patches/0139.sql
```diff
@@ -0,0 +1,28 @@
...
+    categories VARCHAR(255) PRIMARY KEY
```
cassidysymons, line 5:
```suggestion
    category VARCHAR(255) PRIMARY KEY
```

Prefer to keep column names in the same singular/plural form as the data they're storing.

microsetta_private_api/db/patches/0139.sql
```diff
@@ -0,0 +1,28 @@
...
+);
+
+INSERT INTO observation_categories (categories) VALUES ('Sample'), ('Swab'), ('Tube');
```
cassidysymons, line 8:
```suggestion
INSERT INTO sample_observation_categories (category) VALUES ('Sample'), ('Swab'), ('Tube');
```

microsetta_private_api/db/patches/0139.sql
```diff
@@ -0,0 +1,28 @@
...
+
+CREATE TABLE observation_errors (
```
cassidysymons, line 10:
```suggestion
CREATE TABLE sample_observations (
```

Some of these conditions are not errors in a strict sense, so it would be more appropriate to name the table "sample_observations."

microsetta_private_api/db/patches/0139.sql
```diff
@@ -0,0 +1,28 @@
...
+    category VARCHAR(255),
+    error VARCHAR(255),
```
cassidysymons, line 12:
```suggestion
    category VARCHAR(255) NOT NULL,
    observation VARCHAR(255) NOT NULL,
```

Along with the column name change and enforcing `NOT NULL` constraints, want a couple of keys on this table:
1) A primary key, either an int or UUID (e.g. `observation_id UUID PRIMARY KEY DEFAULT uuid_generate_v4()`)
2) A unique key on the combination of category+observation to ensure we don't end up with duplicate values down the line.

microsetta_private_api/db/patches/0139.sql
```diff
@@ -0,0 +1,28 @@
...
+    project_id INT,
```
cassidysymons, line 13: A couple of things here:
1) The API currently isn't respecting the association between an observation and a project ID. I pulled up the code on our staging server and I'm getting the observations created below for barcodes associated with any project, not just 118.
2) Let's move this association to a separate table (e.g. `sample_observation_project_associations`). A given observation can (and likely will) be used for multiple projects, so we don't want to duplicate all of the data.

microsetta_private_api/db/patches/0139.sql
```diff
@@ -0,0 +1,28 @@
...
+('Tube', 'Screw cap is loose', 118),
+('Tube', 'Insufficient ethanol', 118),
+('Tube', 'No ethanol', 118),
+('Swab', 'No swab in tube', 118),
+('Swab', 'Multiple swabs in tube', 118),
+('Swab', 'Incorrect swab type', 118),
+('Sample', 'No visible sample', 118),
+('Sample', 'Excess sample on swab', 118);
+
+ALTER TABLE barcodes.barcode_scans ADD COLUMN observations VARCHAR(255);
```
cassidysymons, line 28: I think it makes more sense to have a separate table to store the association between a scan and the observations that were recorded as part of that scan. One of the forward-looking use cases we have for this data is analyzing the frequency with which they occur, and querying a table for that specifically will be more efficient than clunky `LIKE` statements on a string column in the `barcode_scans` table.

microsetta_private_api/admin/tests/test_admin_repo.py
```diff
@@ -323,15 +323,18 @@ def make_tz_datetime(y, m, d):
             "barcode": test_barcode,
             "scan_timestamp": make_tz_datetime(2017, 7, 16),
             "sample_status": 'no-registered-account',
-            "technician_notes": "huh?"
+            "technician_notes": "huh?",
+            "observations": None
```
cassidysymons, line 327: It looks like there aren't any new unit tests that exercise the observations functionality being added. Off the top of my head, there are a few scenarios that would be valuable to test:
1) When no observations are recorded for a scan, confirm that the database does not reflect any observations.
2) When one or more observations are recorded for a scan, confirm that the database reflects all of the observations.
3) When a scan attempts to record observations that are not associated with the barcode's project(s), confirm that it raises a RepoException.

microsetta_private_api/admin/tests/test_admin_repo.py
```diff
@@ -347,6 +350,7 @@ def make_tz_datetime(y, m, d):
                 self.assertGreater(len(diag['projects_info']), 0)
                 self.assertEqual(len(diag['scans_info']), 2)
                 # order matters in the returned vals, so test that
+                print(diag['scans_info'][0], first_scan)
```
cassidysymons, line 353:
```suggestion
```

Let's make sure to pull out debug info when PRs are ready to review.

cassidysymons (changes requested):

microsetta_private_api/db/patches/0140.sql
```diff
@@ -0,0 +1,56 @@
...
+    observation VARCHAR(255) NOT NULL,
+    FOREIGN KEY (category) REFERENCES sample_observation_categories(category),
+    UNIQUE (category, observation)
+);
+
+-- Create table to store associations between observations and projects
+CREATE TABLE sample_observation_project_associations (
+    id UUID PRIMARY KEY DEFAULT uuid_generate_v4(),
+    observation_id UUID NOT NULL,
+    project_id INT NOT NULL,
```
cassidysymons, line 24: `project_id` should also be a foreign key referencing `barcodes.project.project_id`

microsetta_private_api/db/patches/0140.sql
```diff
@@ -0,0 +1,56 @@
+-- May 13, 2024
+-- Create table to store observation categories
+CREATE TABLE sample_observation_categories (
```
cassidysymons, line 3:
```suggestion
CREATE TABLE barcodes.sample_observation_categories (
```

Apologies for missing this the first time around, but let's make sure to specify a schema when creating tables (this applies to this and all following queries, don't want to needlessly replicate the comment). Ideally, we want all new queries (whether in DB patches or repos) to specify the schema as well. Some legacy code doesn't, and queries work without it, but specificity helps prevent issues down the road.

microsetta_private_api/repo/admin_repo.py
```diff
@@ -1102,6 +1133,32 @@ def scan_barcode(self, sample_barcode, scan_info):
                 scan_args
             )
 
+            if scan_info['observations']:
+                for observation in scan_info['observations']:
+                    cur.execute(
+                        "SELECT observation_id FROM sample_observations "
+                        "WHERE observation = %s",
+                        (observation,)
+                    )
```
cassidysymons, line 1142: Given that there's not a unique constraint on `observation` (and we don't want there to be), this isn't a safe approach. The UI in `microsetta-admin` should directly pass the `observation_id` into the API.

microsetta_private_api/repo/admin_repo.py
```diff
@@ -424,14 +446,23 @@ def _rows_to_dicts_list(rows):
                     and barcode_info is None:
                 return None
 
+            cur.execute("""
```
cassidysymons, line 449: I don't think it's appropriate to return every possible observation in `retrieve_diagnostics_by_barcode` function. There are a handful of points in the codebase that call `retrieve_diagnostics_by_barcode` and most of them are only looking for info about the barcode, so it's creating bloat.

Instead, let's create a new API path to serve this purpose. The input would be a list of one or more project IDs and the output would be the possible observations that can be rendered for those project IDs.

microsetta_private_api/repo/admin_repo.py
```diff
@@ -1091,7 +1122,7 @@ def scan_barcode(self, sample_barcode, scan_info):
                 sample_barcode,
                 datetime.datetime.now(),
                 scan_info['sample_status'],
-                scan_info['technician_notes']
+                scan_info['technician_notes'],
```
cassidysymons, line 1125: Is this code change necessary or is it a residual effect of moving the observations for a given scan into a separate table?

microsetta_private_api/repo/admin_repo.py
```diff
@@ -1102,6 +1133,32 @@ def scan_barcode(self, sample_barcode, scan_info):
...
+                    result = cur.fetchone()
+                    if result is None:
+                        raise RepoException(
+                            f"No observation_id found for "
+                            f"observation: {observation}"
+                        )
+
+                    observation_id = result[0]
+
+                    cur.execute(
```
cassidysymons, line 1153: Before inserting the observations into `sample_barcode_scan_observations`, it would be valuable to verify that they're associated with the barcode(s) project(s). The UI _shouldn't_ allow invalid observations to be used, but validating would be more defensive.

cassidysymons (changes requested):

microsetta_private_api/repo/admin_repo.py
ayobi: Yeah, residual, removing

microsetta_private_api/db/patches/0140.sql
```diff
@@ -0,0 +1,57 @@
...
+);
+
+-- Create table to store associations between observations and projects
+CREATE TABLE barcodes.sample_observation_project_associations (
+    id UUID PRIMARY KEY DEFAULT uuid_generate_v4(),
+    observation_id UUID NOT NULL,
+    project_id INT NOT NULL,
+    FOREIGN KEY (observation_id) REFERENCES barcodes.sample_observations(observation_id),
+    FOREIGN KEY (project_id) REFERENCES barcodes.project(project_id),
+    UNIQUE (observation_id)
```
cassidysymons, line 27:
```suggestion
    UNIQUE (observation_id, project_id)
```

cassidysymons: We need to be able to associate observations with more than one project.

microsetta_private_api/repo/admin_repo.py
```diff
@@ -443,6 +466,23 @@ def _rows_to_dicts_list(rows):
 
             return diagnostic
 
+    def retrieve_observations_by_project(self, sample_barcode):
+        def _rows_to_dicts_list(rows):
+            return [dict(x) for x in rows]
```
cassidysymons, line 471: Since multiple functions in admin_repo are now using this helper function, let's move it from being an inner function with multiple instances to a single top-level function.

microsetta_private_api/repo/admin_repo.py
```diff
@@ -443,6 +466,23 @@ def _rows_to_dicts_list(rows):
...
+                        SELECT so.*, sopa.project_id
+                FROM barcodes.sample_observations so
+                JOIN barcodes.sample_observation_project_associations sopa
+                ON so.observation_id = sopa.observation_id
+                JOIN barcodes.project_barcode pb
+                ON sopa.project_id = pb.project_id
+                WHERE pb.barcode = %s
+            """, (sample_barcode,))
+            observations = _rows_to_dicts_list(cur.fetchall())
+            return observations
```
cassidysymons, line 484: This code does not work as intended when a sample is associated with multiple projects and two or more of the projects are associated with the same observation. It shows multiple copies of the same observation on the `microsetta-admin` UI.

microsetta_private_api/admin/admin_impl.py
```diff
@@ -153,6 +153,25 @@ def scan_barcode(token_info, sample_barcode, body):
...
+    if not sample_barcode:
+        response = jsonify({"error": "Barcode is required"})
+        response.status_code = 400
+        return response
+
+    with Transaction() as t:
+        admin_repo = AdminRepo(t)
+        observation = \
+            admin_repo.retrieve_observations_by_project(sample_barcode)
+        t.commit()
```
cassidysymons, line 168: Given that this API path is read-only, is there a reason to have a `commit` here?

microsetta_private_api/admin/admin_impl.py
```diff
@@ -153,6 +153,25 @@ def scan_barcode(token_info, sample_barcode, body):
...
+
+    response = jsonify({"observation": observation})
+    response.status_code = 201
```
cassidysymons, line 171: Is there a reason you're returning `201` rather than `200` here?

cassidysymons (changes requested):

microsetta_private_api/admin/admin_impl.py
ayobi: You're right, I can't remember what I was doing when I was trying to get it to work

microsetta_private_api/admin/admin_impl.py
ayobi: My mistake, changed.

microsetta_private_api/admin/admin_impl.py
```diff
@@ -153,6 +153,25 @@ def scan_barcode(token_info, sample_barcode, body):
     return response
 
 
+def get_observations(token_info, sample_barcode):
+    validate_admin_access(token_info)
+
+    if not sample_barcode:
+        response = jsonify({"error": "Barcode is required"})
+        response.status_code = 400
+        return response
```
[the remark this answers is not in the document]
ayobi: You're right, it isn't necessary, thank you for the catch.

microsetta_private_api/repo/admin_repo.py
ayobi: Should be fixed now

cassidysymons:
@ayobi Can you please merge biocore:master into ayobi:vocab_samples_issues? There's a database patch missing that causes the patching process to fail. Thanks.

ayobi:
@cassidysymons merged 🎉

microsetta_private_api/admin/tests/test_admin_repo.py
```diff
@@ -323,15 +324,21 @@ def make_tz_datetime(y, m, d):
             "barcode": test_barcode,
             "scan_timestamp": make_tz_datetime(2017, 7, 16),
             "sample_status": 'no-registered-account',
-            "technician_notes": "huh?"
+            "technician_notes": "huh?",
+            "observation_id": None,
+            "observations": None,
+            "category": None
```
cassidysymons, line 330:
```suggestion
            "observations": None
```

Tests should reflect exactly what microsetta-admin is posting to microsetta-private-api.

microsetta_private_api/admin/tests/test_admin_repo.py
```diff
@@ -323,15 +324,21 @@ def make_tz_datetime(y, m, d):
...
         second_scan = {
             "barcode_scan_id": second_scan_id,
             "barcode": test_barcode,
             "scan_timestamp": make_tz_datetime(2020, 12, 4),
             "sample_status": 'sample-is-valid',
-            "technician_notes": None
+            "technician_notes": None,
+            "observation_id": None,
+            "observations": None,
+            "category": None
```
cassidysymons, line 341:
```suggestion
            "observations": None
```

Tests should reflect exactly what microsetta-admin is posting to microsetta-private-api.

cassidysymons (changes requested):
Two minor adjustments, then this should be good to merge. Thanks!

PR #575, opened 2024-06-04
Copy Barcodes

cassidysymons:
The wet lab recently had to move two samples into tubes with new barcodes. This pull request updates the state of the database to copy all information associated with the two samples into their new barcodes.


wasade (commented):
Isn't `ag_kit_barcode_id` used as an FK with other tables?

cassidysymons:
> Isn't `ag_kit_barcode_id` used as an FK with other tables?

Good catch, ag.vioscreen_registry uses it as an FK. Will adjust, re-deploy to staging, and ping you when it's ready for re-review.

wasade (approved):

PR #581, opened 2024-08-31
multi ffqs fix

ayobi:


microsetta_private_api/admin/sample_summary.py
```diff
@@ -80,9 +80,12 @@ def per_sample(project, barcodes, strip_sampleid):
...
 
-                ffq_complete, ffq_taken, _ = vs_repo.get_ffq_status_by_sample(
-                    sample.id
-                )
+                if source:
+                    ffq_complete, ffq_taken, _ = \
+                        vs_repo.get_ffq_status_by_source(source.id)
+                else:
+                    ffq_complete = False
+                    ffq_taken = False
```
cassidysymons, line 88: Given the change from sample to source, I think it makes sense to move this code block into the if statement on Line 51 (`if source is not None and source_type == Source.SOURCE_TYPE_HUMAN`). Additionally, it makes sense to harmonize what they're doing. For example, `get_ffq_status_by_source` could return `True, True` but `vio_id` could be None.

I'd suggest first looking for a match including `sample_id` since a good chunk of historical data will include that. Failing an exact match with `sample_id` you can fall back to a match on source.

microsetta_private_api/repo/vioscreen_repo.py
```diff
@@ -55,6 +55,10 @@ def upsert_session(self, session):
                                {doupdateset}
                            """,
                         tuple([getattr(session, attr) for attr in self.COLS]))
+            cur.execute("SELECT * FROM ag.vioscreen_sessions "
+                        "WHERE sessionId = %s", (session.sessionId,))
+            rows = cur.fetchall()
+            print("Upserted rows:", rows)
```
cassidysymons, line 61: Was this exploratory/debug code that was intended to be removed prior to committing?

cassidysymons (changes requested):

microsetta_private_api/repo/vioscreen_repo.py
ayobi: Ah yes, meant to remove that

microsetta_private_api/admin/sample_summary.py
ayobi: So you're saying it'd be better to change `if source is not None and source_type == Source.SOURCE_TYPE_HUMAN` to first check for a `sample_id` match then fall back on source? Something like this on line 51:
``` python 
if sample is not None:
    vio_id = template_repo.get_vioscreen_id_if_exists(account.id,
                                                      source.id,
                                                      sample.id)
    if not vio_id:
        vio_id = template_repo.get_vioscreen_id_if_exists(account.id,
                                                          source.id,
                                                          None)
else:
    if source is not None and source_type == Source.SOURCE_TYPE_HUMAN:
        vio_id = template_repo.get_vioscreen_id_if_exists(account.id,
                                                          source.id,
                                                          None)
    if source:
        ffq_complete, ffq_taken, _ = vs_repo.get_ffq_status_by_source(source.id)
    else:
        ffq_complete = False
        ffq_taken = False
```

microsetta_private_api/repo/vioscreen_repo.py
```diff
@@ -190,17 +190,17 @@ def get_ffq_status_by_sample(self, sample_uuid):
...
-                is_taken = status in ('Started', 'Review', 'Finished')
-                return (is_complete, is_taken, status)
             else:
-                raise ValueError("A sample should not have multiple FFQs")
+                results = []
+                for status in res:
+                    is_complete = status[0] == 'Finished'
+                    is_taken = status[0] in ('Started', 'Review', 'Finished')
+                    results.append((is_complete, is_taken, status[0]))
+                return results
```
cassidysymons, line 203: Doesn't changing the return value from a tuple to a list of tuples will break the code that uses this function? The unit tests work fine because they test the function in isolation, but is this still working when you try to run a sample summary report through microsetta-admin?

microsetta_private_api/admin/sample_summary.py
```diff
@@ -49,10 +49,24 @@ def per_sample(project, barcodes, strip_sampleid):
...
                                                                   sample.id)
+                # fall back on matching with source id
+                if not vio_id:
+                    vio_id = \
+                        template_repo.get_vioscreen_id_if_exists(account.id,
+                                                                 source.id,
+                                                                 None)
+                if source:
+                    ffq_complete, ffq_taken, _ = \
+                        vs_repo.get_ffq_status_by_source(source.id)
```
cassidysymons, line 63: Two comments here:
1) These lines are within a block of code that already checked `if source is not None` - isn't checking `if source` here redundant?
2) If we successfully obtain a `vio_id` we should check the taken/complete status of that particular `vio_id` rather than anything associated with the source.

cassidysymons (changes requested):

microsetta_private_api/repo/vioscreen_repo.py
ayobi: That's true, it does break. So, if the user has taken multiple FFQs for a particular sample, we'd probably need to change the columns for ffq-taken and ffq-complete to indicate which ffq they've taken and if that one was taken/complete?

cassidysymons: I think that's overkill for the sake of this particular function. If a sample (or source) is associated with more than one FFQ, let's return the one (and it's taken/complete statuses) that's temporally nearest to the sample collection date.

ayobi: Ok, in my latest commit I get the closest sample collection data in `sample_date` and `sample_time` in `ag.ag_kit_barcode` to compare in `startdate` in `ag.vioscreen_sessions`.

microsetta_private_api/admin/sample_summary.py
ayobi: Gotcha. In my latest commit I've changed it in favor of looking up FFQs status via `vio_id`. In the query though, since `vio_id` is not in `ag.ag_kit_barcode` I first use the `vio_id` to obtain the `source_id` from `ag.vioscreen_registry`. Not sure if that's the best approach.

PR #583, opened 2024-09-19
external survey - skin scoring app

ayobi:


microsetta_private_api/api/_survey.py
```diff
@@ -181,6 +182,37 @@ def _remote_survey_url_spain_ffq(transaction, account_id, source_id):
...
+                                            source_id,
+                                            language_tag):
+    st_repo = SurveyTemplateRepo(transaction)
+
+    skin_scoring_app_ffq_id, study = \
+        st_repo.get_skin_scoring_app_ffq_id_if_exists(account_id,
+                                                      source_id)
+
+    if skin_scoring_app_ffq_id is None:
+        study = 'SBI'
```
cassidysymons, line 196: `study` was a URL parameter specific to the Polyphenols FFQ, it should be removed for the skin-scoring app.

microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -755,6 +764,115 @@ def delete_spain_ffq(self, account_id, source_id):
...
+        characters = string.ascii_lowercase + string.digits
+        skin_scoring_app_ffq_id = ''.join(random.choices(characters, k=8))
+
+        with self._transaction.cursor() as cur:
+            cur.execute("""INSERT INTO ag.skin_scoring_app_ffq_registry
+                        (skin_scoring_app_ffq_id, account_id,
+                         source_id, language_tag, study)
+                        VALUES (%s, %s, %s, %s, %s)""",
+                        (skin_scoring_app_ffq_id, account_id,
+                         source_id, language_tag, study))
```
cassidysymons, line 799: Granting the fact that it's unlikely for this to occur, the code needs to gracefully handle the possibility that a duplicate ID will be generated by attempting to generate IDs until a unique value is created.

microsetta_private_api/repo/sample_repo.py
```diff
@@ -404,6 +407,17 @@ def _get_supplied_kit_id_by_sample(self, sample_barcode):
...
 
+    def _get_project_id_by_sample(self, sample_barcode):
+        with self._transaction.cursor() as cur:
+            cur.execute(
+                "SELECT project_id "
+                "FROM barcodes.project_barcode "
+                "WHERE barcode = %s",
+                (sample_barcode, )
+            )
+            row = cur.fetchone()
```
cassidysymons, line 418: A barcode can be associated with multiple projects, so `fetchall()` would be more appropriate here. Downstream, that also needs to be accounted for.

microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -755,6 +766,127 @@ def delete_spain_ffq(self, account_id, source_id):
...
+            The user's language tag
+
+        Returns
+        -------
+        str
+            The newly created Skin Scoring App ID
+        """
+        characters = string.ascii_lowercase + string.digits
+
+        for attempt in range(5):
```
cassidysymons, line 791: I think it would be more appropriate for this to be a while loop than a numerically constrained for loop. A collision is unlikely, and five consecutive collisions is even more unlikely, but exhausting all 36^8 possible strings is impossible on any reasonable timeline.

microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -755,6 +766,127 @@ def delete_spain_ffq(self, account_id, source_id):
...
+                                 source_id,
+                                    SurveyTemplateRepo.SKIN_SCORING_APP_ID))
+
+                    return skin_scoring_app_id
+            except psycopg2.IntegrityError as e:
+                self._transaction.rollback()
+
+                if e.pgcode == '23505':
+                    print(f"Duplicate ID '{skin_scoring_app_id}' "
+                          f"detected. Retrying... (Attempt {attempt + 1}/{5})")
```
cassidysymons, line 821:
```suggestion
```

microsetta_private_api/repo/survey_template_repo.py
```diff
@@ -755,6 +766,127 @@ def delete_spain_ffq(self, account_id, source_id):
...
+                cur.execute("""DELETE FROM ag.ag_login_surveys
+                               WHERE ag_login_id=%s
+                                   AND source_id=%s
+                                   AND survey_id=%s""",
+                            (account_id, source_id, existing))
+            cur.execute("""DELETE FROM ag.skin_scoring_app_registry
+                           WHERE account_id=%s
+                               AND source_id=%s""",
+                        (account_id, source_id))
+
```
cassidysymons, line 889: As far as I can tell, nothing in the code base actually uses the function `delete_polyphenol_ffq` anymore - is that correct? If so, it doesn't make sense to create this function, which also will not be called.

cassidysymons (changes requested):
Along with the specific comments, can you please add unit tests that exercise the changes to any modified repos? Thanks.

PR #590, opened 2024-12-16
Add Barcodes to Existing Kits

cassidysymons:
This PR addresses two related needs:
1) It allows an admin to add barcodes to existing kits, either by providing barcodes or allowing the system to generate novel barcodes. The function to add barcodes is only setup to add one barcode per kit, and they can only be admin-provided _or_ system-generated each time the function is used. We have no current or imminent use cases that would require adding a mix of barcode types to existing kits and chose not to undertake the labor necessary to support more complex operations.
2) It enhances the kit creation process by allowing the admin to create kits with a mix of admin-provided barcodes (e.g. matrix tubes with pre-determined barcodes) and system-generated barcodes. This mix of barcode types operates under the notion of "sample slots," such that, for a set of kits with x number of samples, each of those samples may be either admin-provided or system-generated for the entire set of samples. We have no current or imminent use cases that would justify more complex mixing and matching of barcode types.

The corresponding microsetta-admin PR is https://github.com/biocore/microsetta-admin/pull/110


microsetta_private_api/admin/admin_impl.py
```diff
@@ -267,13 +267,42 @@ def create_kits(body, token_info):
...
     kit_prefix = body.get('kit_id_prefix', None)
     project_ids = body['project_ids']
+    barcodes = body['barcodes']
 
     with Transaction() as t:
         admin_repo = AdminRepo(t)
 
+        # First, do some basic sanitation
+        errors = []
+        for barcode_list in barcodes:
```
wasade, line 277: What's the expected `len` of `barcodes`, is it `== len(number_of_kits)`? That's what I would expect, where the interpretation would be "use these samples for kit 0, use these samples for kit 1, etc", but it looks like the _elements_ of `barcodes` are `== len(number_of_kits)`? Or should that be `len(number_of_samples)`?

microsetta_private_api/admin/admin_impl.py
```diff
@@ -267,13 +267,42 @@ def create_kits(body, token_info):
...
+                if len(barcode_list) != number_of_kits:
+                    errors.append(
+                        f'Incorrect number of barcodes given for Sample {x}'
+                    )
+
+                # And verify that the barcodes don't already exist
+                for barcode in barcode_list:
+                    diag = admin_repo.retrieve_diagnostics_by_barcode(barcode)
+                    if diag is not None:
+                        errors.append(f'Barcode {barcode} already exists')
```
wasade, line 291: I believe the test for existence, followed by creation, needs to occur within a lock, right?

microsetta_private_api/admin/admin_impl.py
```diff
@@ -267,13 +267,42 @@ def create_kits(body, token_info):
...
+                # If barcodes were provided for a given slot, ensure the
+                # quantity matches the number of kits
+                if len(barcode_list) != number_of_kits:
+                    errors.append(
+                        f'Incorrect number of barcodes given for Sample {x}'
+                    )
+
+                # And verify that the barcodes don't already exist
+                for barcode in barcode_list:
+                    diag = admin_repo.retrieve_diagnostics_by_barcode(barcode)
```
wasade, line 289: This is a heavy weight request given how the result is used. I wonder whether `admin_repo.exists_{barcode,kit,project}` would be valuable additions? If that is deferred, the cost of this request can be reduce by passing `grab_kit=False`

microsetta_private_api/admin/admin_impl.py
```diff
@@ -282,6 +311,52 @@ def create_kits(body, token_info):
...
+
+    errors = []
+
+    with Transaction() as t:
+        kit_repo = KitRepo(t)
+        admin_repo = AdminRepo(t)
+
+        # First, make sure all of the Kit IDs are valid
+        for kit_id in kit_ids:
+            diag = kit_repo.get_kit_all_samples(kit_id)
```
wasade, line 329: Same comment here re: testing for existence

microsetta_private_api/admin/admin_impl.py
```diff
@@ -282,6 +311,52 @@ def create_kits(body, token_info):
...
+        if generate_barcodes is False:
+            # Next, check that the Barcodes are unique if we're not generating
+            # novel ones
+            for barcode in barcodes:
+                diag = admin_repo.retrieve_diagnostics_by_barcode(barcode)
+                if diag is not None:
+                    errors.append(f'Barcode {barcode} already exists')
+
+            # And verify that the number of barcodes and kit IDs are equal
+            if len(barcodes) != len(kit_ids):
```
wasade, line 342: This test should be performed before the checks against the database under the fail fast principle

microsetta_private_api/admin/admin_impl.py
```diff
@@ -282,6 +311,52 @@ def create_kits(body, token_info):
...
+                errors.append("Unequal number of Kit IDs and Barcodes")
+
```
wasade, line 344: Can a test for `generate_barcodes is True and len(barcodes) > 0` be included? That would be erroneous, right?

microsetta_private_api/admin/admin_impl.py
```diff
@@ -282,6 +311,52 @@ def create_kits(body, token_info):
...
+        # We found one or more issues with the supplied Kit IDs or Barcodes.
+        # Return the error message to microsetta-admin.
+        if len(errors) > 0:
+            error_str = "; ".join(errors)
+            return jsonify(
+                code=422,
+                message=error_str
+            ), 422
+
+        diag = admin_repo.add_barcodes_to_kits(kit_ids, barcodes)
```
wasade, line 354: I _think_ this should be in a try/except given that `create_kits` is above

microsetta_private_api/admin/tests/test_admin_repo.py
```diff
@@ -669,6 +671,63 @@ def test_create_kits_success_is_microsetta(self):
...
+        # Sample Slot 3 - generated
+        sample_3_barcodes = []
+
+        with Transaction() as t:
+            admin_repo = AdminRepo(t)
+            tmi = admin_repo.create_kits(
+                1,
+                3,
+                'foo',
+                [sample_1_barcodes, sample_2_barcodes, sample_3_barcodes],
```
wasade, line 697: I'm puzzled by this, and not sure this works as intended. What's provided are not the barcode IDs, but UUIDs? I was expecting the interface to consume barcodes as we print on the tubes?

microsetta_private_api/admin/tests/test_admin_repo.py
```diff
@@ -669,6 +671,63 @@ def test_create_kits_success_is_microsetta(self):
...
+                cur.execute("SELECT supplied_kit_id "
+                            "FROM ag.ag_kit "
+                            "WHERE supplied_kit_id IN %s",
+                            (tuple(tmi_kits),))
+                observed = cur.fetchall()
+                self.assertEqual(len(observed), 1)
+
+            kit = tmi['created'][0]
+            # Assert that the provided barcodes are present
+            self.assertEqual(kit['sample_barcodes'][0], sample_1_barcodes[0])
```
wasade, line 726: These aren't barcodes though

wasade (commented):
I'm following up by email

microsetta_private_api/admin/admin_impl.py
cassidysymons: Given that this is an admin-facing tool with infrequent usage, I've opted to return a comprehensive set of errors so they can all be remedied at once, rather than simply returning the first error the API encounters. As such, the sequence of checks isn't important.

microsetta_private_api/admin/admin_impl.py
cassidysymons: I'm not following why this would be in a try/except block. It's a completely separate function/API path from creating kits, and all of the reasonable checks have been performed by the time we attempt to add barcodes to kits.

microsetta_private_api/admin/admin_impl.py
cassidysymons: The length of barcodes is number of samples, and the length of each element of barcodes is the number of kits.

I've added a comment to the code to clarify the intended structure.

cassidysymons:
Lint is failing as GitHub is changing (has changed) what "ubuntu-latest" means for workflow and python 3.7 isn't available in the new version (source: https://github.com/actions/runner-images/issues/10636). I'll revisit updating our workflow files to use Ubuntu 22.04 on Monday. Not a long-term fix as support for that will expire in two years, I'll revisit the longer term implications in a couple of months.

marie202378 (approved):

cassidysymons:
Workflow issue is resolved

microsetta_private_api/repo/admin_repo.py
```diff
@@ -924,18 +924,46 @@ def create_kits(self, number_of_kits, number_of_samples, kit_prefix,
...
+        # Iterate through sample slots and use provided barcodes
+        for barcode_list in barcodes:
+            if len(barcode_list) > 0:
+                # Admin provided barcodes for this sample slot, use them
+                kit_barcode_tuples = list(zip(kit_names, barcode_list))
+                kit_name_and_barcode_tuples_list += kit_barcode_tuples
+                new_barcodes += barcode_list
+
+        # See if we need to generate barcodes
+        slots_to_generate = sum(1 for i in barcodes if len(i) == 0)
```
wasade, line 948: I think this and the next few lines would follow as an `else` off the if statement on line 941?

wasade: I think this means that we either generate barcodes, or have received from admin-provided barcodes, for ALL kits for the given `barcodes` element.

microsetta_private_api/repo/admin_repo.py
```diff
@@ -1015,7 +1043,71 @@ def _generate_novel_barcodes(self, number_of_kits, number_of_samples,
                     kit_name_and_barcode_tuples_list.append(
                         (name, new_barcodes[offset + i]))
 
-        return kit_name_and_barcode_tuples_list, new_barcodes
+            return kit_name_and_barcode_tuples_list, new_barcodes
```
wasade, line 1046: Was the change of scope consequential? I wouldn't expect it to be? 

```suggestion
        return kit_name_and_barcode_tuples_list, new_barcodes
```

microsetta_private_api/admin/admin_impl.py
```diff
@@ -282,6 +321,60 @@ def create_kits(body, token_info):
     return jsonify(kits), 201
 
 
+def add_barcodes_to_kits(body, token_info):
+    validate_admin_access(token_info)
+
+    kit_ids = body['kit_ids']
+    barcodes = body['barcodes']
```
wasade, line 328: Is this expected to be the same structure as described above?

wasade: I don't think so, as line 355 makes sure `len(barcodes) == len(kit_ids)` rather than `number_of_samples`

microsetta_private_api/admin/admin_impl.py
```diff
@@ -282,6 +321,60 @@ def create_kits(body, token_info):
...
+        # First, make sure all of the Kit IDs are valid
+        for kit_id in kit_ids:
+            diag = admin_repo.check_exists_kit(kit_id)
+            if diag is False:
+                errors.append(f'Kit ID {kit_id} not found')
+
+        if generate_barcodes is False:
+            # Next, check that the Barcodes are unique if we're not generating
+            # novel ones
+            for barcode in barcodes:
```
wasade, line 349: I _think_ much of the validation work here is similar or the same as what's done in `create_kits`, should it be decomposed?

microsetta_private_api/admin/admin_impl.py
```diff
@@ -282,6 +321,60 @@ def create_kits(body, token_info):
...
+                if diag is True:
+                    errors.append(f'Barcode {barcode} already exists')
+
+            # And verify that the number of barcodes and kit IDs are equal
+            if len(barcodes) != len(kit_ids):
+                errors.append("Unequal number of Kit IDs and Barcodes")
+        else:
+            if len(barcodes) > 0:
+                errors.append(
+                    "Barcodes may not be both generated and provided"
```
wasade, line 360: The structure description though seems to suggest this mode is supported?

```
    # This allows the creation of a set of kits with a mixture of
    # system-generated and admin-provided barcodes.
```

microsetta_private_api/admin/admin_impl.py
wasade: The methods have criteria which would trigger a raise. It would be more useful to provide a error message to the admin than to trigger a `500`

microsetta_private_api/repo/admin_repo.py
```diff
@@ -1015,7 +1043,71 @@ def _generate_novel_barcodes(self, number_of_kits, number_of_samples,
...
+            kits_barcodes_tuples = list(zip(kit_ids, barcodes))
+
+        with self._transaction.cursor() as cur:
+            for kit_id, barcode in kits_barcodes_tuples:
+                kit_d = self.retrieve_diagnostics_by_kit_id(kit_id)
+                ag_kit_id = kit_d['kit_id']
+
+                projects = kit_d['sample_diagnostic_info'][0]['projects_info']
+                project_ids = [x['project_id'] for x in projects]
+                is_tmi = self._are_any_projects_tmi(project_ids)
```
wasade, line 1081: It should be impossible for `is_tmi` to be `False` as the `retrieve_diagnostics_by_kit_id` call queries the `ag.ag_kit` table

microsetta_private_api/repo/admin_repo.py
```diff
@@ -1015,7 +1043,71 @@ def _generate_novel_barcodes(self, number_of_kits, number_of_samples,
...
+
+                barcode_projects = []
+                for project_id in project_ids:
+                    barcode_projects.append((barcode, project_id))
+
+                # Add barcodes to barcode table
+                cur.execute(
+                    "INSERT INTO barcodes.barcode(kit_id, barcode, status) "
+                    "VALUES (%s, %s, 'unassigned')",
```
wasade, line 1090: It looks like this is the second use of this string for this purpose in two different functions. Could this be pulled out and expressed as a literal?

microsetta_private_api/repo/admin_repo.py
```diff
@@ -1015,7 +1043,71 @@ def _generate_novel_barcodes(self, number_of_kits, number_of_samples,
...
+        kit_ids : list of str
+            The kit IDs for which we're adding barcodes
+        barcodes : list of str
+            The list of admin-supplied barcodes; if empty generates novel
+            barcodes
+
+        Returns
+        -------
+        list of tuples
+            The pairings of newly added kit IDs/barcodes
```
wasade, line 1062: These barcodes are not associated with box IDs. It looks like it may be necessary to insert the kit UUID (see https://github.com/biocore/microsetta-private-api/blob/210e5c3d15d9a5b3a5c47202cff4fc679e7d5ff5/microsetta_private_api/repo/admin_repo.py#L1039)

wasade: Will samples added to an existing kit be shipped via daklapack?

microsetta_private_api/repo/admin_repo.py
```diff
@@ -1047,9 +1139,16 @@ def _create_kits(self, kit_names, new_barcodes,
         with self._transaction.cursor() as cur:
             # create barcode project associations
             barcode_projects = []
-            for barcode in new_barcodes:
-                for prj_id in project_ids:
-                    barcode_projects.append((barcode, prj_id))
+            if isinstance(new_barcodes, list) and \
+                    all(isinstance(item, list) for item in new_barcodes):
```
wasade, line 1143: The docstring should be revised

microsetta_private_api/repo/admin_repo.py
```diff
@@ -1047,9 +1139,16 @@ def _create_kits(self, kit_names, new_barcodes,
...
+                for barcodes in new_barcodes:
+                    for barcode in barcodes:
+                        for prj_id in project_ids:
+                            barcode_projects.append((barcode, prj_id))
+            else:
+                for barcode in new_barcodes:
+                    for prj_id in project_ids:
+                        barcode_projects.append((barcode, prj_id))
```
wasade, line 1151:
```suggestion
                flat_barcodes = [bc for barcodes in new_barcodes for bc in barcodes]       
                new_barcodes = flat_barcodes
                
            for barcode in new_barcodes:
                for prj_id in project_ids:
                    barcode_projects.append((barcode, prj_id))
```

microsetta_private_api/repo/admin_repo.py
```diff
@@ -1146,14 +1245,14 @@ def create_kit(self, kit_name, box_id, address_dict,
...
 
-        kit_names = [kit_name]
         address = None if address_dict is None else json.dumps(address_dict)
         kit_details = [{KIT_BOX_ID_KEY: box_id,
                        KIT_ADDRESS_KEY: address,
                        KIT_OUTBOUND_KEY: outbound_fedex_code,
                        KIT_INBOUND_KEY: inbound_fedex_code}]
         kit_name_and_barcode_tuples_list = \
             [(kit_name, x) for x in barcodes_list]
+        kit_names = [kit_name]
```
wasade, line 1255: Was this relocation consequential?

microsetta_private_api/repo/admin_repo.py
```diff
@@ -924,18 +924,46 @@ def create_kits(self, number_of_kits, number_of_samples, kit_prefix,
...
-                                 kit_name_and_barcode_tuples_list,
-                                 number_of_samples, project_ids)
+        new_barcodes = []
+        kit_name_and_barcode_tuples_list = []
+
+        # Iterate through sample slots and use provided barcodes
+        for barcode_list in barcodes:
+            if len(barcode_list) > 0:
+                # Admin provided barcodes for this sample slot, use them
+                kit_barcode_tuples = list(zip(kit_names, barcode_list))
```
wasade, line 943: ...but `create_kits` cannot assume it is only ever called from the API endpoint where the validation is presently performed, therefore we do not have assurance that `len(kit_names) == len(barcode_list)` here. Can that check be added?

wasade (commented):
I'm not sure if this is ready to re-review or not, but I cannot get to it later this week and next week carries a lot of uncertainty. Added some additional comments, but was unable to review the unit tests. The input structure on creation of kits feels surprising still even with documentation, but I may just be looking at this from a different angle -- I'm not sure if @AmandaBirmingham has time for a second opinion

microsetta_private_api/admin/admin_impl.py
```diff
@@ -267,13 +267,52 @@ def create_kits(body, token_info):
     number_of_samples = body['number_of_samples']
     kit_prefix = body.get('kit_id_prefix', None)
     project_ids = body['project_ids']
+    # NB: The barcodes variable is a list of lists, structured with the notion
+    # of representing barcodes' constituent lists each representing one sample
+    # slot in the set of kits being created.
+    # len(barcodes) == number_of_samples && len(barcodes[x]) == number_of_kits
+    # This allows the creation of a set of kits with a mixture of
+    # system-generated and admin-provided barcodes.
```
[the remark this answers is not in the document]
cassidysymons: I've restructured the payload to be explicit, rather than implicit, which I believe addresses these concerns.

microsetta_private_api/repo/admin_repo.py
cassidysymons: No, we have no current or foreseeable use cases where Daklapack would ship the additional samples.

cassidysymons:
@wasade This PR is ready for re-review.

microsetta_private_api/admin/admin_impl.py
```diff
@@ -267,21 +267,137 @@ def create_kits(body, token_info):
...
+                    diag = admin_repo.check_exists_barcode(barcode)
+                    if diag is True:
+                        errors.append(f'Barcode {barcode} already exists')
+            else:
+                # Barcodes list should be empty in this sample slot's dict
+                if len(sample_slot['barcodes']) != 0:
+                    errors.append(
+                        f'Barcodes erroneously provided for Sample {x}'
+                    )
+            x += 1
```
wasade, line 310:
```suggestion
        for sidx, sample_slot in enumerate(barcodes_container, 1):
            if sample_slot['barcodes_provided']:
                # If barcodes were provided for a given slot, ensure the
                # quantity matches the number of kits
                if len(sample_slot['barcodes']) != number_of_kits:
                    errors.append(
                        f'Incorrect number of barcodes provided for Sample {sidx}'
                    )

                # And verify that the barcodes don't already exist
                for barcode in sample_slot['barcodes']:
                    diag = admin_repo.check_exists_barcode(barcode)
                    if diag:
                        errors.append(f'Barcode {barcode} already exists')
            else:
                # Barcodes list should be empty in this sample slot's dict
                if len(sample_slot['barcodes']) != 0:
                    errors.append(
                        f'Barcodes erroneously provided for Sample {sidx}'
                    )
```

microsetta_private_api/admin/admin_impl.py
```diff
@@ -267,21 +267,137 @@ def create_kits(body, token_info):
...
+
+        if len(errors) > 0:
```
wasade, line 312:
```suggestion
        if errors:
```

microsetta_private_api/admin/admin_impl.py
```diff
@@ -267,21 +267,137 @@ def create_kits(body, token_info):
...
+                number_of_kits,
+                number_of_samples,
+                kit_prefix,
+                barcodes_container,
+                project_ids
+            )
+        except KeyError as ke:
+            return jsonify(
+                code=422,
+                message="Unable to create kits: " + str(ke)
```
wasade, line 327: This will only contain the message within the exception (e.g., `KeyError("the stuff here")`) but it will not contain the traceback itself. Just to verify, is that the desired intent?

microsetta_private_api/admin/admin_impl.py
```diff
@@ -267,21 +267,137 @@ def create_kits(body, token_info):
...
+        admin_repo = AdminRepo(t)
+
+        # Lock the barcode table for the duration of checking the existience
+        # of barcodes, then (hopefully) adding them to kits
+        t.lock_table("barcode")
+
+        # First, make sure all of the Kit IDs are valid
+        for kit_id in kit_ids:
+            diag = admin_repo.check_exists_kit(kit_id)
+            if diag is False:
```
wasade, line 359:
```suggestion
            if not diag:
```

microsetta_private_api/admin/admin_impl.py
```diff
@@ -267,21 +267,137 @@ def create_kits(body, token_info):
...
+                errors.append(f'Kit ID {kit_id} not found')
+
+        if generate_barcodes is False:
```
wasade, line 362:
```suggestion
        if not generate_barcodes:
```

microsetta_private_api/admin/admin_impl.py
```diff
@@ -267,21 +267,137 @@ def create_kits(body, token_info):
...
+            # Next, check that the Barcodes are unique if we're not generating
+            # novel ones
+            for barcode in barcodes:
+                diag = admin_repo.check_exists_barcode(barcode)
+                if diag is True:
```
wasade, line 367:
```suggestion
                if diag:
```

microsetta_private_api/admin/admin_impl.py
```diff
@@ -267,21 +267,137 @@ def create_kits(body, token_info):
...
+                errors.append("Unequal number of Kit IDs and Barcodes")
+        else:
+            if len(barcodes) > 0:
+                errors.append(
+                    "Barcodes may not be both generated and provided"
+                )
+
+        # We found one or more issues with the supplied Kit IDs or Barcodes.
+        # Return the error message to microsetta-admin.
+        if len(errors) > 0:
```
wasade, line 381:
```suggestion
        if errors:
```

microsetta_private_api/repo/admin_repo.py
```diff
@@ -924,18 +926,61 @@ def create_kits(self, number_of_kits, number_of_samples, kit_prefix,
...
-        return self._create_kits(kit_names, new_barcodes,
-                                 kit_name_and_barcode_tuples_list,
-                                 number_of_samples, project_ids)
+        new_barcodes = []
+        kit_name_and_barcode_tuples_list = []
+
+        slots_to_generate = 0
+        # Iterate through sample slots and use provided barcodes
+        for sample_slot in barcodes_container:
+            if sample_slot['barcodes_provided'] is True:
```
wasade, line 948:
```suggestion
            if sample_slot['barcodes_provided']:
```

microsetta_private_api/repo/admin_repo.py
```diff
@@ -1619,3 +1727,37 @@ def update_perk_fulfillment_state(self, new_state):
...
+    def check_exists_barcode(self, barcode):
+        with self._transaction.cursor() as cur:
+            cur.execute(
+                "SELECT COUNT(barcode) "
+                "FROM barcodes.barcode "
+                "WHERE barcode = %s",
+                (barcode, )
+            )
+            res = cur.fetchone()
+            return True if res[0] > 0 else False
```
wasade, line 1740:
```suggestion
            return res[0] > 0
```

microsetta_private_api/repo/admin_repo.py
```diff
@@ -1619,3 +1727,37 @@ def update_perk_fulfillment_state(self, new_state):
...
+    def check_exists_kit(self, kit_id):
+        with self._transaction.cursor() as cur:
+            cur.execute(
+                "SELECT COUNT(kit_id) "
+                "FROM barcodes.kit "
+                "WHERE kit_id = %s",
+                (kit_id, )
+            )
+            res = cur.fetchone()
+            return True if res[0] > 0 else False
```
wasade, line 1751:
```suggestion
            return res[0] > 0
```

wasade (commented):
Thanks, @cassidysymons. The suggestions here are largely superficial but help tighten the code.

Looking at this all again, I'm left uncertain about _why_ this PR is needed. This PR introduces a fundamental change to the roughly 13 year history of the notion of a kit. Specifically, kits previously were immutable. While I cannot identify an issue on static review, a change like this creates a high risk of unintended consequences later. In the past, we've simply resent kits to people rather than modify existing kits. 

If the intention is to allow the creation of kits separate from Daklapak, then there is no need to modify existing kits. If that is the goal, then we cannot rely on administrators to define IDs correctly, and should instead have the system construct IDs for the administrator as was done prior to offloading kit construction onto Daklapak. It's not an if but when an administrator will get the ID construction wrong, and it can be quite costly to address the fallout from that.

cassidysymons:
@wasade I just sent you an email about the reasoning behind the PR. I'll go through the technical feedback shortly (as well as the feedback on the corresponding microsetta-admin PR).

PR #596, opened 2025-01-10
Skin Surveys

cassidysymons:
This PR adds two new surveys:

- Skin
- Skin Health Diagnosis

And all of the constituent questions, responses, and information for those surveys. In addition, it adds a few new questions to other surveys and moves a couple of questions between surveys.

Corresponding microsetta-interface PR is https://github.com/biocore/microsetta-interface/pull/329


microsetta_private_api/api/tests/test_integration.py
```diff
@@ -702,9 +704,9 @@ def test_bobo_takes_all_local_surveys(self):
                                  SurveyTemplateRepo.ALLERGIES_ID,
                                  SurveyTemplateRepo.DIET_ID,
                                  SurveyTemplateRepo.DETAILED_DIET_ID,
-                                 SurveyTemplateRepo.MIGRAINE_ID,
-                                 SurveyTemplateRepo.SURFERS_ID,
-                                 SurveyTemplateRepo.COVID19_ID):
```
AmandaBirmingham, line 707: Curious: why were these surveys deleted?

AmandaBirmingham (approved):
Just one question, for my own edification

microsetta_private_api/api/tests/test_integration.py
cassidysymons: The surveys were removed from the actual user flow during the relaunch process, but not from this integration test. The fact that the test wasn't failing, even though we retired the surveys, is something that I'll need to investigate and I've created a [GitHub issue](https://github.com/biocore/microsetta-private-api/issues/598) reflecting that. Good catch.

PR #597, opened 2025-01-13
Cheek Samples

cassidysymons:
-- Description to be provided --

### DIFF
--- a/microsetta_private_api/db/patches/0144.sql
+++ b/microsetta_private_api/db/patches/0144.sql
@@ -1,0 +1,39 @@
+-- Beginning with cheek samples, we're collecting metadata that are explicitly
+-- linked to sample collection (unlike surveys, which are implicitly linked
+-- to samples via sources), but not globally collected, and therefore don't
+-- belong in the ag.ag_kit_barcodes table. A new table will store these
+-- fields and could eventually be extended to a much more robust framework.
+
+-- First, we're going to set up a table to enforce validation. Specifically,
+-- we want to ensure that we don't receive any unexpected field names, and
+-- the values we receive are conformant to either an expected format or a set
+-- of pre-defined values.
+CREATE TYPE BARCODE_METADATA_VALIDATION_TYPE AS ENUM ('date_or_time', 'set_value');
+CREATE TABLE ag.ag_kit_barcodes_metadata_validation (
+    field_name VARCHAR NOT NULL PRIMARY KEY,
+    validation_type BARCODE_METADATA_VALIDATION_TYPE NOT NULL,
+    validation_value TEXT NOT NULL
+);
+
+-- Sets of pre-defined values will use | as delimiter, as that is unlikely to
+-- appear in any foreseeable string we'll be using.
+INSERT INTO ag.ag_kit_barcodes_metadata_validation (field_name, validation_type, validation_value) VALUES
+    ('sample_site_last_washed_date', 'date_or_time', '%m/%d/%Y'), -- MM/DD/YYYY format, e.g. '01/15/2024'
+    ('sample_site_last_washed_time', 'date_or_time', '%I:%M %p'), -- H:MM AM/PM format, e.g. '3:05 PM'
+    ('sample_site_last_washed_product', 'set_value', 'Soap (includes bar and liquid soap)|Foaming face wash|Face cleanser|Just water|Other (e.g. shampoo, body wash, all-in-one or all-over wash)|Not sure');
+
+-- Then, create the table to store the data
+CREATE TABLE ag.ag_kit_barcodes_metadata (
+    ag_kit_barcode_id UUID NOT NULL,
+    field_name VARCHAR NOT NULL,
+    field_value VARCHAR NOT NULL,
+
+    -- Foreign key relationship on ag_kit_barcode_id
+    CONSTRAINT fk_ag_kit_barcode_id FOREIGN KEY (ag_kit_barcode_id) REFERENCES ag.ag_kit_barcodes (ag_kit_barcode_id),
+
+    -- Foreign key relationship on ag_kit_barcodes_metadata_validation
+    CONSTRAINT fk_ag_kit_barcodes_metadata_field_name FOREIGN KEY (field_name) REFERENCES ag.ag_kit_barcodes_metadata_validation(field_name),
+
+    -- Unique key on the combination of ag_kit_barcode_id and field_name
+    CONSTRAINT uk_ag_kit_barcode_id_field_name UNIQUE (ag_kit_barcode_id, field_name)
+);
--- a/microsetta_private_api/model/sample.py
+++ b/microsetta_private_api/model/sample.py
@@ -26,10 +26,14 @@ class Sample(ModelBase):
         self.account_id = account_id
 
         self.accession_urls = []
+        self.barcode_meta = {}
         self.kit_id = kit_id
 
     def set_accession_urls(self, accession_urls):
         self.accession_urls = accession_urls
+
+    def set_barcode_meta(self, barcode_meta):
+        self.barcode_meta = barcode_meta
 
     @property
     def edit_locked(self):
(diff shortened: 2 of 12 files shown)


microsetta_private_api/repo/metadata_repo/_constants.py
```diff
@@ -356,7 +356,18 @@
...
+        'sample_type': 'Cheek',
+        'scientific_name': 'human skin metagenome',
+        'taxon_id': '539655',
+        'host_body_habitat': 'UBERON:skin',
+        'empo_3': 'Animal surface',
+        'env_material': 'sebum',
+        'env_package': 'human-skin',
+        'description': 'Microsetta Initiative Cheek sample',
+        'host_body_site': 'UBERON:skin of head'}
 }
```
AmandaBirmingham, line 371: In October, I had to check the constant metadata values for human cheek with Gail Ackermann, who is the final arbiter of these values in the Knight lab; she indicated that current best practice (it is constantly changing!) is that `host_body_site` should be `UBERON: cheek`, `sample_type` should be `skin of cheek`, and `env_material` should be `sebum material`.  Where does CMI get its guidance from?  I'm not surprised there are differences, but I'd be happier if we could align them ...

Also, for some point in the future, it would be nice to get these constants updated to v2 of EMPO, which specifies 4 levels instead of 3 (and it isn't just a bolt-on of one more level ... the whole level structure changes :-/ )

AmandaBirmingham (commented):
One question/future request re constant metadata values for cheek samples.

microsetta_private_api/repo/metadata_repo/_constants.py
cassidysymons: The values here are based on what was used for skin samples when AGP was actively collecting/processing them. We'll update these (cheek and other types) to reflect current recommendations.

microsetta_private_api/repo/sample_repo.py
```diff
@@ -321,6 +330,106 @@ def update_info(self, account_id, source_id, sample_info,
                             sample_info.id
                         ))
 
+        # NB: We run this even if the barcode_meta dict is empty, as that
+        # means the answers associated with the last sample update should be
+        # purged. This applies both to the normal sample update process, as
+        # well as dissociate_sample().
```
AmandaBirmingham, line 336: This seems a little weird.  If this metadata is *truly* barcode-level (that is, sample level) and NOT source level; and if it is NOT survey data, then it doesn't make sense that it is wiped when a sample is dissociated from its source (and thus also its surveys)--you'd think that the sample-level metadata would remain, since the sample record itself continues to exist (just not attached to a source or survey).  

The fact that this metadata *should* be wiped on the dissociation of a sample from its source & survey(s) suggests that our modeling of this metadata as as "barcode-level" is flawed.  

At this point, I am not going to ask that we revisit this (I don't think anyone is in love with the current table solution, so I suspect we will be circling back to it sometime soon anyway) but I do think it is a code smell that we will eventually have to pay attention to.

microsetta_private_api/repo/sample_repo.py
```diff
@@ -451,3 +560,62 @@ def scrub(self, account_id, source_id, sample_id):
...
+            return False
+
+        ret_dict = {}
+        bc_valid = True
+        for fn, fv in barcode_meta.items():
+            if fn == "sample_site_last_washed_date":
+                try:
+                    ret_val = datetime.datetime.strptime(
+                        fv,
+                        "%m/%d/%Y"
```
AmandaBirmingham, line 601: This hardcoded value is used twice in this file (here and in `_get_barcode_meta`), which to me means it should be stored in a string variable (well, constant) so that it can be found/changed/fixed if necessary.  

Same goes for the time format (which, I must note, is slightly *different* here than it is in `_get_barcode_meta`: "%I:%M %p" vs "%-I:%M %p" !)

microsetta_private_api/db/patches/0144.sql
```diff
@@ -0,0 +1,23 @@
...
+
+-- Then, create the table to store the data
+-- Note: the date and time are stored separately because we're not enforcing
+-- either as a required field. As such, using a timestamp type would not be
+-- appropriate since it forces us into a both or neither paradigm.
+CREATE TABLE ag.ag_kit_barcodes_cheek (
+    ag_kit_barcode_id UUID NOT NULL PRIMARY KEY,
+    sample_site_last_washed_date DATE,
+    sample_site_last_washed_time TIME,
+    sample_site_last_washed_product SAMPLE_SITE_LAST_WASHED_PRODUCT_TYPE,
```
AmandaBirmingham, line 19: Is this allowed to be null?  Under what circumstances would that be the case?

microsetta_private_api/model/sample.py
```diff
@@ -74,18 +78,22 @@ def to_api(self):
...
+            "barcode_meta": self.barcode_meta
         }
 
 
 # A SampleInfo represents the set of end user editable fields whose lifetime
 # matches that of the association between a sample and a source
 class SampleInfo:
-    def __init__(self, sample_id, datetime_collected, site, notes):
+    def __init__(self, sample_id, datetime_collected, site, notes,
+                 barcode_meta=None):
```
AmandaBirmingham, line 90: Curious: why does barcode_meta here default to `None` whereas in `Sample` it defaults to an empty dictionary?

AmandaBirmingham (changes requested):
Nothing major

microsetta_private_api/repo/sample_repo.py
cassidysymons: The reason I treated it this way is because the dissociate_sample() function removes sample-related info such as collection date and sample site. From that perspective, I think that removing the values for these cheek fields is the logical move.

microsetta_private_api/repo/sample_repo.py
cassidysymons: It's not entirely logical on the surface, but `strptime()` and `strftime()` treat `%I` and `%-I` slightly differently. I created constants for all three formats (one date and two time) and included a comment explaining the reasoning for the date format differences.

cassidysymons:
@AmandaBirmingham I think this newest commit addresses the feedback that required changes, and I've added comments to notes that either don't seem to require change or were a strictly informational question. Please let me know if I missed anything or you have any further questions or concerns. Thanks!

microsetta_private_api/model/sample.py
```diff
@@ -87,7 +87,7 @@ def to_api(self):
 # matches that of the association between a sample and a source
 class SampleInfo:
     def __init__(self, sample_id, datetime_collected, site, notes,
-                 barcode_meta=None):
+                 barcode_meta={}):
```
wasade, line 90: Can this be reverted? There is an important unexpected edgecase with default objects:

```
>>> def foo(a, b, example={}):
...   example[a] = b
...   print(example)
... 
>>> foo(1, 2)
{1: 2}
>>> foo(3, 4)
{1: 2, 3: 4}
>>> foo(1, 10)
{1: 10, 3: 4}
```

In almost all cases, the intended behavior and recommended approach is:

```
>>> def foo(a, b, example=None):
...   if example is None:
...     example = {}
...   example[a] = b
...   print(example)
... 
>>> foo(1, 2)
{1: 2}
>>> foo(3, 4)
{3: 4}
>>> foo(1, 10)
{1: 10}
```

These _should_ be caught by linters (see e.g. https://stackoverflow.com/a/73765790), which if not flagged here suggests we may need to revisit the linter configuration or linter used.

cassidysymons: Done - https://github.com/biocore/microsetta-private-api/pull/597/commits/6a31b02cc7d4b9500c6a8800a1aab1de99c3de4e

wasade: Thanks!

microsetta_private_api/db/patches/0144.sql
AmandaBirmingham: Ok ... now I'm wondering what the meaningful difference is between (a) a completed cheek sample with no record in the `ag_kit_barcodes_cheek` table and (b) a completed cheek sample with a record in the `ag_kit_barcodes_cheek` table for which all fields except the foreign key back to the `ag_kit_barcodes` table are NULL.  Do these two cases have two distinct meanings?  What is signified by a completely empty (except for FK) record in the cheek table?

microsetta_private_api/repo/sample_repo.py
AmandaBirmingham: Thanks!

microsetta_private_api/db/patches/0144.sql
cassidysymons: From a metadata perspective, they do have a distinct meaning. NULL values would be reported as "not provided" while a cheek sample with no values for the questions (including NULL) would be reported as "not collected."

AmandaBirmingham: Huh, that's fascinating--I don't think I really understand what these differences mean in this context, but good to know!

AmandaBirmingham (approved):

PR #601, opened 2025-02-15
Fix query of FFQ registration code to check not null

lmerchant:
Fix query of registration FFQ code where previous SQL was "ffq_registration_code = <passed FFQ code>" AND "registration_code_used IS NULL". The new SQL adds "IS NOT NULL".


cassidysymons:
This change is counter to the intended behavior, as it would allow FFQ codes to be re-used. Can you expand on why you believe it was causing the button on the front end to not work properly?

lmerchant:
OK, I see what I did. I entered a FFQ code once and the code worked but then I used the same code a second time and it was refused, which was correct, but I looked at the database and saw code_used was filled in and assumed that was needed for a True result. I should have created some tests to make sure my code worked. I'll delete this pull request.

